PR #1, opened 2019-09-18
Update image installed packages and code to python3

aweimeow:
Due to some weird reason, the container built by 2nd step will have the `ANSI` as default encoding, so I need to install `locales` and generate `UTF-8` encoding for mininet script's emoji (☠️ & ⚡️). If we didn't change the encoding to UTF-8, it may cause python encoding exception.

EDIT: The built container image has tested with `stratum/tutorial:basic`


ccascone:
Hi @aweimeow,  under which circumstances do you see python encoding exceptions? stratum.py has already the UTF-8 encoding [special comment](https://www.python.org/dev/peps/pep-0263/#proposed-solution) at the beginning of the file: https://github.com/stratum/stratum/blob/master/tools/mininet/stratum.py#L1

Bumping to python 3 might be abreaking change for anyone using this Docker image to run custom Mininet scripts. I'd rather find a fix that does not involve upgrading the Python version, or simply remove the emojis from the script...

Yi-Tseng:
I think we can ask developers to move to py2 to py3, and officially drop python2 form this image on 2020/01/01.
The code change looks fine, it should works on both py2 and py3.

aweimeow:
Hello @ccascone, although the `stratum.py` has the encoding header, if the terminal encoding didn't compatible with emoji character, it'll cause exception finally, here is a proof of concept.

```
aweimeow@MacBook-Pro:[~]$ cat test.py
# -*- coding: utf-8 -*-

print("☠️")
aweimeow@MacBook-Pro:[~]$ echo "$LANG $LC_CTYPE"
zh_TW.UTF-8 zh_TW.UTF-8
aweimeow@MacBook-Pro:[~]$ python test.py
☠️
aweimeow@MacBook-Pro:[~]$ LANG=ANSI LC_CTYPE=ANSI python test.py
Traceback (most recent call last):
  File "test.py", line 3, in <module>
    print("\u2620\ufe0f")
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
```

I have no idea about which package or what command makes the new container's environment doesn't have UTF-8's encoding, so that is the reason for installing `locales` package to generate the corresponding language locale file.

<br>

About the upgrading to Python 3, I think it's necessary. There is a lot of [famous Python 3rd library announced](https://python3statement.org) they won't support Python 2 after Python 2's EoL, despite `stratum.py` don't have many dependencies, but if the 3rd libraries have something wrong, we also can't get support on the failure.

From now to the Python 2 EoL, there have 3 months to go, if the upgrade causes the problem, the team still has time to fallback / fix the issue :)

tools/mininet/Dockerfile
```diff
@@ -23,8 +23,8 @@
 FROM stratumproject/build:build as builder
 
 ENV BUILD_DEPS \
-    python-setuptools \
```
bocon13, line 26: maybe install both python2 and python3 packages for backwards compatibility

pudelkoM:
Closed due to inactivity. Feel free to reopen if you want to continue to work on this.

Yi-Tseng:
Let's reopen this pull request since we think it is better to move to python3

.circleci/config.yml
```diff
@@ -70,7 +70,7 @@ jobs:
   # Build targets and run unit tests.
   unit_tests:
     docker:
-      - image: stratumproject/build:build
+      - image: stratumproject/build:build-py3
```
Yi-Tseng, line 73: change back after this PR merged?

bocon13 (dismissed):

Yi-Tseng:
TODO:
delete the `update_config.py` since we  already have pipeline builder

bocon13 (dismissed):
lgtm

pudelkoM (approved):

bocon13 (approved):

aweimeow:
🎉🎉🎉

bocon13:
Finally PR #1 is merged!! Thanks @aweimeow!

PR #3, opened 2019-10-01
Adding interface mac_config

mhwang408:
@Yi-Tseng Add a new field mac_address into config_params of
message SingletonPort to enable configuration of mac_address.

### DIFF
--- a/stratum/hal/lib/common/common.proto
+++ b/stratum/hal/lib/common/common.proto
@@ -279,6 +279,7 @@ message PortConfigParams {
   ModulationType modulation_type = 5;
   // FEC operational mode
   FecMode fec_mode = 6;
+  uint64 mac_address =7;
 }
 
 // Chassis uniquely identifies a switch with a single management interface,
--- a/stratum/hal/lib/common/common.proto
+++ b/stratum/hal/lib/common/common.proto
@@ -280,7 +280,7 @@ message PortConfigParams {
   // FEC operational mode
   FecMode fec_mode = 6;
   // The configured mac address for this port.
-  uint64 mac_address =7;
+  uint64 mac_address = 7;
 }
 
 // Chassis uniquely identifies a switch with a single management interface,
--- a/stratum/hal/lib/common/common.proto
+++ b/stratum/hal/lib/common/common.proto
@@ -279,6 +279,7 @@ message PortConfigParams {
   ModulationType modulation_type = 5;
   // FEC operational mode
   FecMode fec_mode = 6;
+  // The configured mac address for this port.
   uint64 mac_address =7;
 }
 
--- a/stratum/hal/lib/common/utils.cc
+++ b/stratum/hal/lib/common/utils.cc
@@ -333,6 +333,14 @@ std::string MacAddressToYangString(
                          mac_address & 0xFF);
 }
 
+::google::protobuf::uint64 YangStringToMacAddress(
+    const std::string& yang_string) {
+  std::string tmp_str = yang_string;
+  // Remove colons
+  tmp_str.erase(std::remove(tmp_str.begin(), tmp_str.end(), ':'), tmp_str.end());
+  return strtoul(tmp_str.c_str(), NULL, 16);
+}
+
 bool IsPortAutonegEnabled(const TriState& state) {
     return state == TriState::TRI_STATE_TRUE;
 }
--- a/stratum/hal/lib/common/utils.cc
+++ b/stratum/hal/lib/common/utils.cc
@@ -17,6 +17,7 @@
 #include "stratum/hal/lib/common/utils.h"
 
 #include <sstream>  // IWYU pragma: keep
+#include <regex>
 
 #include "stratum/lib/constants.h"
 #include "stratum/lib/macros.h"
@@ -339,6 +340,11 @@ std::string MacAddressToYangString(
   // Remove colons
   tmp_str.erase(std::remove(tmp_str.begin(), tmp_str.end(), ':'), tmp_str.end());
   return strtoull(tmp_str.c_str(), NULL, 16);
+}
+
+bool IsMacAddressValid(const std::string& mac_address) {
+  const std::regex mac_address_regex(kMacAddressRegex);
+  return regex_match(mac_address, mac_address_regex);
 }
 
 bool IsPortAutonegEnabled(const TriState& state) {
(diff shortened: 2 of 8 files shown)


stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -2455,9 +2455,6 @@ void YangParseTreePaths::AddSubtreeInterfaceFromSingleton(
 
   node = tree->AddNode(GetPath("interfaces")(
       "interface", name)("ethernet")("config")("mac-address")());
-  // TODO(tmadejski) Replace the mock value of config MAC address
-  // (0x112233445566) with a value read from the configuration pushed by the
-  // controller once such field is added.
   SetUpInterfacesInterfaceEthernetConfigMacAddress(0x112233445566, node);
```
stolsma: Shouldn't the fixed mac address be changed too? Or do I miss something? (If thats the case just ignore this remark... 😎)

mhwang408: I copy these lines to another place and forget to delete them.
Thank you.

stratum/hal/lib/common/yang_parse_tree_test.cc
```diff
@@ -1002,6 +1002,50 @@ TEST_F(YangParseTreeTest,
...
 }
 
+// Check if the 'config/mac-address' OnUpdate action works correctly.
+TEST_F(YangParseTreeTest,
+       InterfacesInterfaceEthernetConfigMacAddressOnUpdateSuccess) {
+  auto path = GetPath("interfaces")(
+      "interface", "interface-1")("ethernet")("config")("mac-address")();
+  static constexpr char kMacAddressYangString[] = "11:22:33:44:55:66";
+  static constexpr uint64 kMacAddressUint64 = 0x112233445566ull;
+  static constexpr char kMacAddressSomethingString[] = "00;00;00;00;00;00";
```
stolsma: You used ; instead of : as separator, probably wrong?

mhwang408: Yes, you're right.
But in this case, it's on purpose to test if a user inputs an unformed mac address, he/she will get a "wrong value" response.

stratum/hal/lib/common/common.proto
```diff
@@ -279,6 +279,8 @@ message PortConfigParams {
   ModulationType modulation_type = 5;
   // FEC operational mode
   FecMode fec_mode = 6;
+  // The configured mac address for this port.
+  uint64 mac_address =7;
```
Yi-Tseng: add a space between `=` and `7`

stratum/hal/lib/common/utils.cc
```diff
@@ -333,6 +333,14 @@ std::string MacAddressToYangString(
                          mac_address & 0xFF);
 }
 
+::google::protobuf::uint64 YangStringToMacAddress(
+    const std::string& yang_string) {
+  std::string tmp_str = yang_string;
+  // Remove colons
+  tmp_str.erase(std::remove(tmp_str.begin(), tmp_str.end(), ':'), tmp_str.end());
+  return strtoul(tmp_str.c_str(), NULL, 16);
```
Yi-Tseng: I'd suggest using `strtoull` since `unsigned long` can be 32-bit in some cases.
https://en.cppreference.com/w/cpp/language/types

mhwang408: Got it, thanks.

stratum/hal/lib/common/common.proto
mhwang408: Done!

stratum/hal/lib/common/yang_parse_tree_test.cc
pudelkoM: How about giving the malformed MAC string its own test (`InterfacesInterfaceEthernetConfigMacAddressOnUpdateFailure`)? There are more error cases (too long, too short, null bytes, ...) to be checked, which would clutter the success test case.

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -1047,7 +1047,7 @@ void SetUpInterfacesInterfaceEthernetConfigMacAddress(uint64 node_id,
     std::string mac_address_string = typed_val->string_val();
     ::google::protobuf::uint64 mac_address =
         YangStringToMacAddress(mac_address_string);
-    if (mac_address == 0) {
+    if (mac_address_string.compare("00:00:00:00:00:00") !=0 && mac_address == 0) {
```
pudelkoM: Use `==` for string comparison where the ordering does not matter.

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -1030,9 +1033,67 @@ void SetUpInterfacesInterfaceEthernetConfigMacAddress(uint64 mac_address,
...
+    // Set the value.
+    auto status = SetValue(node_id, port_id, tree,
+                           &SetRequest::Request::Port::mutable_mac_address,
+                           &MacAddress::set_mac_address, mac_address);
+    if (status != ::util::OkStatus()) {
+      return status;
+    }
+
+    // Update the chassis config
+    ChassisConfig* new_config = config->writable();
```
pudelkoM: Minor nitpick: Do you see an easy way to hit this code in a test?
Would improve test coverage and make the coverall check happy.

stratum/hal/lib/common/yang_parse_tree_paths.cc
mhwang408: Cool...Good point.
I've added the test for InterfacesInterfaceEthernetConfigMacAddressOnUpdateFailure as your suggestion, thanks.

stratum/hal/lib/common/admin_service_test.cc
```diff
@@ -166,7 +166,7 @@ TEST_P(AdminServiceTest, CancelReboot) {
   ::gnoi::system::RebootResponse resp;
   ASSERT_OK(admin_service_->Setup(false));
 
-  req.set_delay(5000000);  // 5ms
+  req.set_delay(10000000);  // 5ms
```
Yi-Tseng: This belongs to another patch, please remove this commit.
Also, need to update the comment

mhwang408: I've removed it and refined the test cases.

stratum/hal/lib/common/yang_parse_tree_paths.cc
mhwang408: The issue: Fix ChassisConfig in YangParseTreeTest reported by Yi Tseng results in this low coverage problem.

pudelkoM: Ok then. 80% should work too.

PR #15, opened 2019-10-10
Fix empty ChassisConfig problem in YangParseTreeTest

mhwang408:
@Yi-Tseng Empty ChassisConfig results in low coverage because
some codes in on_set_functor will never be hit

### DIFF
--- a/stratum/hal/lib/common/yang_parse_tree_test.cc
+++ b/stratum/hal/lib/common/yang_parse_tree_test.cc
@@ -109,15 +109,17 @@ class YangParseTreeTest : public ::testing::Test {
   }
 
   // A proxy for YangParseTree::AddSubtreeInterface().
-  SingletonPort& AddSubtreeInterface(const std::string& name) {
+  void AddSubtreeInterface(const std::string& name,
+                           SingletonPort* singleton_p) {
     absl::WriterMutexLock l(&parse_tree_.root_access_lock_);
-
+    SingletonPort singleton;
+    SingletonPort* singleton_ptr = (singleton_p == nullptr)? &singleton \
+                                                           : singleton_p;
     // Add one singleton port.
-    static SingletonPort singleton;
-    singleton.set_name(name);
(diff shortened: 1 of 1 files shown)


Yi-Tseng:
Please rebase to new master or merger the master to resolve the conflict

stratum/hal/lib/common/yang_parse_tree_test.cc
```diff
@@ -109,15 +109,17 @@ class YangParseTreeTest : public ::testing::Test {
   }
 
   // A proxy for YangParseTree::AddSubtreeInterface().
-  void AddSubtreeInterface(const std::string& name) {
+  void AddSubtreeInterface(const std::string& name,
+                           SingletonPort* singleton_p) {
```
Yi-Tseng: You can use `SingletimePort* singleton_p = nullptr` for default value, so you don't need to modify line 176, 575, 707, ... etc,

stratum/hal/lib/common/yang_parse_tree_test.cc
```diff
@@ -109,15 +109,17 @@ class YangParseTreeTest : public ::testing::Test {
...
     absl::WriterMutexLock l(&parse_tree_.root_access_lock_);
-
-    // Add one singleton port.
     SingletonPort singleton;
-    singleton.set_name(name);
-    singleton.set_node(kInterface1NodeId);
-    singleton.set_id(kInterface1PortId);
-    singleton.set_speed_bps(kTwentyFiveGigBps);
+    SingletonPort* singleton_ptr = (singleton_p == nullptr)? &singleton \
```
Yi-Tseng: Add space before `?`
Also please break the line from `=`
```
SingletonPort* singleton_ptr =
        (singleton_p == nullptr) ? &singleton : singleton_p;
```

mhwang408: I encounter a problem....So I delete the branch and redo works on a new branch. 
I've sent a new pull request.

PR #20, opened 2019-10-19
Implement handling of action annotations

pudelkoM:
This PR implements the missing annotations for actions from the `p4c_annotation_map_files` compiler flag. We already had and used annotations for tables and and match fields, with this we can also enhance table actions.

It also sets the default log level for `p4c-fpm` backend messages down to `WARNING` from `ERROR`, which yields important information when using the compiler:

### Example

In our `main.p4` example P4 program, we have the following action to the set next hop in IPv4 routing:

```p4
    action set_nexthop(PortNum port, EthernetAddress smac, EthernetAddress dmac,
                       bit<12> dst_vlan) {
        standard_metadata.egress_spec = port;
        local_metadata.dst_vlan = dst_vlan;
        hdr.ethernet.src_addr = smac;
        hdr.ethernet.dst_addr = dmac;
        hdr.ipv4_base.ttl = hdr.ipv4_base.ttl - 1;
    }
```

The `dst_vlan` parameter is only there because BCM operates on VLANs internally and we have to get the default VLAN id into the action somehow. `main.p4` does not use VLAN tags in incomming or outgoing packets, the VLAN id here is purely an implementation detail of stratum_bcm.

By using action annotations we can hide these details from the P4 program, when not needed:

In the table map file, we add the following addenda:

```
action_addenda_map {
  key: "ingress.l3_fwd.set_nexthop"
  value {
    addenda_names: "default-vlan-action-fallback"
  }
}

# Defines an internal action parameter on the default VLAN ID
action_addenda {
  name: "default-vlan-action-fallback"
  assignments_addenda {
    assigned_value {
      constant_param: 1
    }
    destination_field_name: "local_metadata.dst_vlan"
  }
}
```

This will be incorporated into the bcm pipeline config, so that stratum_bcm knows to add a vlan ID of 1 to all `set_nexthop` actions: 
```
table_map {
  key: "ingress.l3_fwd.set_nexthop"
  value {
    action_descriptor {
      type: P4_ACTION_TYPE_FUNCTION
      [...]
      assignments {
        assigned_value {
          constant_param: 1
        }
        destination_field_name: "local_metadata.dst_vlan"
      }
    }
  }
}
```

The action then is simplified to:

```p4
    action set_nexthop(PortNum port, EthernetAddress smac, EthernetAddress dmac) {
        standard_metadata.egress_spec = port;
        hdr.ethernet.src_addr = smac;
        hdr.ethernet.dst_addr = dmac;
        hdr.ipv4_base.ttl = hdr.ipv4_base.ttl - 1;
    }
```

### DIFF
--- a/stratum/p4c_backends/fpm/annotation_mapper.cc
+++ b/stratum/p4c_backends/fpm/annotation_mapper.cc
@@ -103,6 +103,10 @@ bool AnnotationMapper::ProcessAnnotations(
           success = false;
         break;
       case hal::P4TableMapValue::kActionDescriptor:
+        if (!HandleActionAnnotations(
+            table_map_iter.first, p4_info_manager,
+            table_map_iter.second.mutable_action_descriptor()))
+          success = false;
       case hal::P4TableMapValue::kHeaderDescriptor:
         break;
       case hal::P4TableMapValue::kInternalAction:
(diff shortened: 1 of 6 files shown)


pudelkoM:
@teverman I can't add you as a reviewer for some reason, but if you have time, take a look at these compiler additions I made.

teverman:
Max, I'm having trouble logging in to github, so I'll post these comments
out of band.

annotation_mapper.cc:

I think you want a break at line 109 instead of falling through.

The warning at line 261 seems like it could be a little less ominous since
it warns every time it executes.  OTOH, filtering it so it only warns when
there are assignments or primitives in the addenda means you might as well
implement the TODO.  Actually, it looks like you removed the warning in the
second commit.

Should line 265 be has_primitive_ops_addenda?

teverman:
I think you want Maximilian Pudelko, not Massimiliano Poletto. :-D

pudelkoM:
> I think you want a break at line 109 instead of falling through.

Absolutely.



> Should line 265 be has_primitive_ops_addenda?

It can't be. `action_addendum->has_assignments_addenda()` is possible because it's a embedded message of type `P4ActionInstructions`. `primitive_ops_addenda` is only an enum field (`P4ActionOp`), which defaults to value 0 when not set. I added an explicit NE check to make it more clear.


As for the warning I added earlier. I was confused by this comment in annotation_map.proto:
```
//  assignments_addenda - if present, replaces or extends the assignments field
//      in the P4ActionDescriptor.
//  primitive_ops_addenda - if present, extends or replaces the primitive_ops
//      field in the P4ActionDescriptor.
```
My first interpretation was, that these fields replace a *single* assignment or primitive op, but together with the comment for `assignment_primitive_replace` it became clear, that they replace the complete descriptor field. I think we should wait until there is a compelling use case that requires a more sophisticated behavior.

Thanks for your review!

stratum/p4c_backends/fpm/p4c_switch_main.cc
```diff
@@ -45,7 +45,10 @@ using stratum::p4c_backends::SwitchP4cBackend;
 using stratum::p4c_backends::TableMapGenerator;
 using stratum::p4c_backends::TargetInfo;
 
+DECLARE_int32(stderrthreshold);
+
 int main(int argc, char** argv) {
+  FLAGS_stderrthreshold = 1;
```
pudelkoM: Remove this. Should be part of separate PR if needed.

stratum/p4c_backends/fpm/annotation_mapper.cc
```diff
@@ -216,5 +221,53 @@ bool AnnotationMapper::MapTableAnnotation(
   return true;
 }
 
+bool AnnotationMapper::HandleActionAnnotations(const std::string& action_name,
+                            const hal::P4InfoManager& p4_info_manager,
+                            hal::P4ActionDescriptor* action_descriptor) {
+  // The name is always the first annotation lookup.
+  bool name_ok = MapActionAnnotation(action_name, action_descriptor);
```
Yi-Tseng: any reason we need to create `name_ok` variable?

pudelkoM: No particular reason, `HandleFieldAnnotations` looks the same: https://github.com/stratum/stratum/pull/20/files/1372bd31c16718c27818e6763951a5536d6ab4f3#diff-791e4f5f37a5882c64543f76fac21897R144-R149

PR #22, opened 2019-10-23
WIP: initial support checkin for direction

renner-osi:


stratum/hal/lib/common/yang_parse_tree.h
```diff
@@ -308,6 +310,7 @@ class TreeNode {
       const EventHandlerRecordPtr& record) const;
 
   bool IsAKey() { return is_name_a_key_; }
+  bool supports_data_type(::gnmi::GetRequest_DataType&) const;
```
Yi-Tseng: Looks good to me to have this function to check if the node supports specific data type.
However we need to handle some cases carefully, for example

A tree supports following paths:
```
/interfaces/interface[name=1]/state/A
/interfaces/interface[name=1]/state/B
/interfaces/interface[name=1]/config/A
/interfaces/interface[name=1]/config/B
```

If you try to send gNMI `GetRequest` with `STATE` type on path `/interfaces/interface/[name=1]`, you will receive values from these paths
```
/interfaces/interface[name=1]/state/A
/interfaces/interface[name=1]/state/B
```

Need to find a way to check the type when Stratum calls the on poll handler.

Yi-Tseng:
@renner-osi Any updates?

bocon13:
Doing some cleanup of open PRs. Feel free to reopen this one if you continue working on it.

Thanks!

PR #33, opened 2019-11-08
Cleanup workarounds

### LINKED ISSUE #28: Revert (now) unnecessary workarounds
Some changes done in, e.g. a0cbf3517ad86cb9c88f124d4c03b3f606d77eb3 (sorry @craigsdell), are unnecessary now that we fixed the underlying issues.

Example: https://github.com/stratum/stratum/commit/a0cbf3517ad86cb9c88f124d4c03b3f606d77eb3#diff-544b79186d8682de1dfea855225dc05bR46-R49
Fixed by moving this snippet to `logging.h`:

```c++
// ostream overload for std::nullptr_t for C++11
// see: https://stackoverflow.com/a/46256849
#if __cplusplus == 201103L
#include <cstddef>
#include <iostream>
namespace std {
::std::ostream& operator<<(::std::ostream& s, ::std::nullptr_t) {
    return s << static_cast<void *>(nullptr);
}
}
#endif
```

pudelkoM:
Fixes #28

### DIFF
--- a/stratum/glue/logging.cc
+++ b/stratum/glue/logging.cc
@@ -71,3 +71,15 @@ void InitStratumLogging() {
 }
 
 }  // namespace stratum
+
+// ostream overload for std::nulptr_t for C++11
+// see: https://stackoverflow.com/a/46256849
+#if __cplusplus == 201103L
+#include <cstddef>
+#include <iostream>
+namespace std {
+::std::ostream& operator<<(::std::ostream& s, ::std::nullptr_t) {
+    return s << static_cast<void *>(nullptr);
+}
+}
+#endif  // __cplusplus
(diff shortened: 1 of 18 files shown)


stratum/glue/logging.cc
```diff
@@ -71,3 +71,15 @@ void InitStratumLogging() {
...
+// see: https://stackoverflow.com/a/46256849
+#if __cplusplus == 201103L
+#include <cstddef>
+#include <iostream>
+namespace std {
+::std::ostream& operator<<(::std::ostream& s, ::std::nullptr_t) {
+    return s << static_cast<void *>(nullptr);
+}
+}
+#endif  // __cplusplus
```
Yi-Tseng: do we need this if we already defined in `logging.h` file?

pudelkoM: Look closely, logging.h contains the function declaration, in logging.cc is the function definition.
It has to be separated or else we would get ODR violations in every file that includes logging.h

Yi-Tseng: I see, thanks

PR #39, opened 2019-11-21
Support new Barefoot SDE and remove old one

Yi-Tseng:


stratum/hal/bin/barefoot/README.md
```diff
@@ -83,14 +83,19 @@ export BSP_PATH=`pwd`/bf-reference-bsp-<SDE_VERSION>
...
 ./p4studio_build.py -up profiles/stratum_profile.yaml --bsp-path $BSP_PATH
 ```
 
 You may also still install the BSP manually. If you are not using the reference
 BSP, you will need to install the BSP yourself (under `$BF_SDE_INSTALL`) based
 on your vendor's instructions.
 
+### Supported SDE versions
+
+ - 8.9.2
```
ccascone: This is inconsistent with the statement above: "These instructions are valid for SDE versions 8.8.0 through 9.0.0."

bazel/deps.bzl
```diff
@@ -98,10 +98,10 @@ def stratum_deps():
             commit = "ca0291420b5b47fa2596a00877d1713aab61dc7a",
         )
 
-    if "com_github_p4lang_PI_bf" not in native.existing_rules():
-        # ----- PI for Barefoot targets -----
+    if "com_github_p4lang_PI_bf_8_9" not in native.existing_rules():
+        # ----- PI for Barefoot SDE 8.9.x targets -----
         remote_workspace(
-            name = "com_github_p4lang_PI_bf",
+            name = "com_github_p4lang_PI_bf_8_9",
```
habeebmohammed: Is PI not required after 8.9? Why name it as PI_bf_8_9?
Can I run stratum on the older version ex: 8.9.1 as well if required?

Yi-Tseng: SDE 9.0.0 and bmv2 uses default PI commit

Yi-Tseng: I didn't add 8.9.1 here. Any reason we need to support 8.9.1?

Yi-Tseng: Actually both 8.9.1 and 8.9.2 uses same PI library

PR #43, opened 2019-11-26
This is the PhalDB Adapter commit that adds the following functions:

craigsdell:
- An Adapater layer as per original design spec, used by upper layers
  to make calls down into the Phal DB for platform management. It supports
  `Get`, `Subscribe` and `Set`.
- Interfaces the onlpphal GetFrontPanelPortInfo function into the adapter
  layer instead of calling the datasources directly.  They can then take
  advantage of whatever caching policy is set for those nodes.
- The adapter layer also exposes a new PhalDB gRPC service that will allow
  `Get`, `Subscribe` and `Set` calls to be made into the DB for debug
  purposes.
- Finally a debug command called `phal_cli` has been created that uses
  the PhalDB gRPC service to make calls to the database on a live running
  Stratum server.  This can be used to test devices and especially the
  dynamic configurators.
- This commit also includes a bug fix for the PhalDB that has been
  in the code for a while now but has only now surfaced as we started to
  do dynamic adding/removing of devices along with Get and Subscription
  requests.  The problem was in the attribute_group.cc code where the
  registered_queries_ map was using a RegisteredQuery object as the
  value and running the deconstructor when the map entry was still in
  use.  This has now been changed over to use a unique_ptr which seems
  to have better memory management and has fixed the problem. A unit test
  has been added to the sfp_configurator_test to test for this problem
  called `DBConcurrencyTest`.


stratum/hal/lib/phal/attribute_group.cc
```diff
@@ -965,10 +977,11 @@ ::util::Status AttributeGroupInternal::TraverseQuery(
         attribute_function) {
   auto reader_lock = AcquireReadable();
   absl::ReaderMutexLock lock(&registered_query_lock_);
-  auto query_info = gtl::FindOrNull(registered_queries_, query);
```
pudelkoM: What's the reason for not using `gtl::FindOrNull` here and further down?

If there is a compiler error it should be fixed (by me probably), instead of using workarounds.

stratum/hal/lib/phal/onlp/switch_configurator.cc
```diff
@@ -443,7 +438,7 @@ ::util::Status OnlpSwitchConfigurator::AddPort(
 
     // All other port types
     default:
-        LOG(INFO) << "card[" << card_id << "]/port[" << port_id
+        LOG(INFO) << "card[" << card_id << "]/port[" << port_id
```
pudelkoM: trailing whitespace

stratum/hal/lib/phal/onlp/switch_configurator.cc
```diff
@@ -421,19 +416,19 @@ ::util::Status OnlpSwitchConfigurator::AddPort(
...
             // Create a new data source
             ASSIGN_OR_RETURN(auto datasource,
-                OnlpSfpDataSource::Make(port_id, onlp_interface_, cache));
+                OnlpSfpDataSource::Make(port, onlp_interface_, cache));
 
             // Create an SFP Configurator
             ASSIGN_OR_RETURN(auto configurator,
-                OnlpSfpConfigurator::Make(port_id, datasource, sfp,
-                                          onlp_interface_));
+                OnlpSfpConfigurator::Make(card_id, port_id, slot, port,
```
pudelkoM: trailing whitespace

craigsdell: fixed, thanks

stratum/hal/lib/phal/onlp/switch_configurator.cc
craigsdell: fixed, thanks

stratum/hal/lib/phal/attribute_group.cc
pudelkoM: `gtl::FindOrNull` does not support `unique_ptr` at the moment. This workaround is thus fine.

bazel/deps.bzl
```diff
@@ -232,6 +232,13 @@ def stratum_deps():
             commit = "a3b25bf1a854ca7245d5786fda4821df77c57827",
         )
 
+    if "rules_cc" not in native.existing_rules():
+        http_archive(
+            name = "rules_cc",
+	    strip_prefix = "rules_cc-master",
+	    urls = ["https://github.com/bazelbuild/rules_cc/archive/master.zip"],
```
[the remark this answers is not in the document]
craigsdell: ok thanks, I've updated to latest commit cfe68f6bc79dea602f2f6a767797f94a5904997f

stratum/hal/lib/common/hal.cc
```diff
@@ -253,6 +257,8 @@ ::util::Status Hal::Run() {
 
 void Hal::HandleSignal(int value) {
   LOG(INFO) << "Received signal: " << strsignal(value);
+  // Need to do Teardown first to clear up any services first
+  Teardown();
```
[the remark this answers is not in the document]
craigsdell: Yes I added this Teardown in the signal handler as interrupts were hanging and not cleaning up (i.e. the Run Wait() was waiting forever.  However, if process dies without going thru signal handler then we'll still need to call Teardown so was leaving both in?

stratum/hal/lib/phal/onlp/sfp_configurator_test.cc
```diff
@@ -251,6 +279,119 @@ TEST_F(OnlpSfpConfiguratorTest, HandlEventNotPresent) {
...
+    // Now wait for subscribe to go off again (should explode)
+    LOG(INFO) << "Issue Read 2";
+    EXPECT_OK(reader->Read(&sub_resp, absl::Seconds(10)));
+
+    // Inspect msg
+    LOG(INFO) << "sub_resp 2" << std::endl
+              << sub_resp.DebugString();
+
+    // Add the sfp
+    LOG(INFO) << "Issue HandleEvent(HW_STATE_PRESENT";
```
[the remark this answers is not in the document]
craigsdell: ok done, thanks

stratum/hal/lib/phal/phal.proto
```diff
@@ -58,7 +58,7 @@ message PhalCardConfig {
     repeated string tx_disable_paths = 7;
   }
 
-  int32 id = 1;                 // (optional) 1-base index identifier
+  int32 slot = 1;                 // (optional) 1-base index identifier
```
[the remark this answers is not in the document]
craigsdell: done

stratum/hal/lib/common/hal.cc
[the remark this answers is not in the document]
craigsdell: Looking at the code some more, it should be fine to get rid of the Teardown call in the Run() function as the Signal Handler should be called in all circumstances to shutdown the server.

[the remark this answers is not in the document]
Yi-Tseng: The `Wait()` should return after we called `Shutdown()`(https://github.com/stratum/stratum/pull/43/files#diff-4a43a81e5c12f8b203e802dcf2e15857R268)

I just realized that we still need to put `Teardown()` call to `Run` function because we need to get status from `Teardown()` and show error messages in main function, for example:
https://github.com/stratum/stratum/blob/master/stratum/hal/bin/barefoot/main.cc#L189
https://github.com/stratum/stratum/blob/master/stratum/hal/bin/bcm/standalone/main.cc#L135
https://github.com/stratum/stratum/blob/master/stratum/hal/bin/bmv2/main.cc#L184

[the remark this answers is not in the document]
craigsdell: I'm happy to pull this change out from my PR if you want to just leave it the way it was, but I found that the Wait was hanging when there is an outstanding Subscribe request so put the Teardown in the Signal Handler and left it in the Run function as well just to be sure.

stratum/hal/lib/phal/phal_cli.cc
```diff
@@ -0,0 +1,401 @@
...
+    }
+    std::cout << "Executed query in " << execute_duration << " us."
+              << std::endl;
+    return ::util::OkStatus();
+  }
+
+  enum cmd_type {Get, Subscribe, Set};
+
+  // Runs the main CLI loop.
+  ::util::Status  RunCli() {
```
Yi-Tseng: remove one space between `Status` and `RunCli`

stratum/hal/lib/phal/phal_cli.cc
```diff
@@ -0,0 +1,401 @@
...
+    absl::Time execute_time = absl::Now();
+
+    // RPC failed
+    if (!status.ok()) {
+      return MAKE_ERROR(ERR_INTERNAL)
+        << "gRPC Get call failed: "
+        << status.error_message();
+    }
+
+    int execute_duration =
```
pudelkoM: `int64` as per abseil docs

stratum/hal/lib/phal/phal_cli.cc
```diff
@@ -0,0 +1,401 @@
...
```
pudelkoM: `int64`

stratum/hal/lib/phal/phal_cli.cc
```diff
@@ -0,0 +1,401 @@
...
+            std::cout << "Invalid value." << std::endl;
+            continue;
+        }
+        // Grab type and then set it in the request
+        std::string type;
+        std::cout << "Type: ";
+        std::getline(std::cin, type);
+
+        if (!strcmp(type.c_str(), "int32")) {
+            ::google::protobuf::int32 val;
```
pudelkoM: What's the reason for using protobuf's integer definitions?

stratum/hal/lib/phal/sfp_adapter.h
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2019
```
pudelkoM: Missing company

stratum/hal/lib/common/hal.cc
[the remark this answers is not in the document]
craigsdell: ok not a bad idea Max.  I think that the reason I put it in here is that for the most part interrupting the stratum binary cleans up properly unless you have an outstanding Subscribe request (which issues a read on a channel).  The comments around the Shutdown call using now() to interrupt active Reads doesn't seem to work for Reads on a channel from memory.

stratum/hal/lib/phal/phal.proto
```diff
@@ -76,7 +76,7 @@ message PhalFanTrayConfig {
     // Path to set the fan speed.
     string fan_control_path = 4;
   }
-  int32 id = 1;                 // (optional) 1-base index identifier
+  int32 slot = 1;                 // (optional) 1-base index identifier
```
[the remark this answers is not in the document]
craigsdell: done

stratum/hal/lib/phal/db.proto
```diff
@@ -191,3 +191,94 @@ message TemperatureData {
...
+    bool indexed = 3;
+    bool all = 4;
+    bool terminal_group = 5;
+  }
+  repeated PathEntry entries = 1;
+}
+
+message GetRequest {
+  oneof query {
+    string str = 1;
```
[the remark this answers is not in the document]
craigsdell: I've allowed for one of either a "string" with a request formatted using the same syntax as the phal_cli command (originally from the onlpphal_cli) or you can send a PathQuery object.

stratum/hal/lib/phal/phal_cli.cc
craigsdell: done

stratum/hal/lib/common/phaldb_service.cc
```diff
@@ -0,0 +1,454 @@
...
+      }
+      from.add_details()->PackFrom(error);
+    }
+  } else {
+    from.set_code(::google::rpc::OK);
+  }
+
+  return ::grpc::Status(ToGrpcCode(from.code()), from.message(),
+                        from.SerializeAsString());
+}
```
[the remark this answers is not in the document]
craigsdell: agree, initially I pulled it from the p4_service code with the expectation that I might have to tune it for phaldb_service, but in the end, I did not need to.  Happy to pull it out.

stratum/hal/lib/common/phaldb_service.cc
```diff
@@ -0,0 +1,454 @@
...
+
+// Parse PB Query string to Phal DB Path
+::util::StatusOr<::stratum::hal::phal::Path> ParseQuery(
```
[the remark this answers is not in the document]
craigsdell: good idea, thanks

stratum/hal/lib/common/phaldb_service.cc
```diff
@@ -0,0 +1,454 @@
...
+    const std::string& query) {
+
+    ::stratum::hal::phal::Path path;
+
+    std::vector<std::string> query_fields = absl::StrSplit(query, "/");
+    bool use_terminal_group = false;
+    if (query_fields[query_fields.size() - 1] == "") {
```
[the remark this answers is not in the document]
craigsdell: yeap 8). thanks.

stratum/hal/lib/common/phaldb_service.cc
```diff
@@ -0,0 +1,454 @@
...
+      use_terminal_group = true;  // Query ends with a '/'
+      query_fields = {query_fields.begin(),
```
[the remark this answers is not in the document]
craigsdell: yeap again. much simpler.

stratum/hal/lib/phal/phal_cli.cc
craigsdell: ok

stratum/hal/lib/phal/phal_cli.cc
craigsdell: ok

stratum/hal/lib/phal/sfp_adapter.h
craigsdell: Hmm looks like a merge problem.  thanks.  Should we also have the "Copyright 2018-present Open Networking Foundation" header?

stratum/hal/lib/phal/phal_cli.cc
craigsdell: should make no difference, happy to change to int32.

stratum/hal/lib/phal/phal_cli.cc
```diff
@@ -0,0 +1,401 @@
...
+        if (val_str == "") {
+            std::cout << "Invalid value." << std::endl;
+            continue;
+        }
+        // Grab type and then set it in the request
+        std::string type;
+        std::cout << "Type: ";
+        std::getline(std::cin, type);
+
+        if (!strcmp(type.c_str(), "int32")) {
```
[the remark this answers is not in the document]
craigsdell: ok, also changed to use the std::string compare instead

stratum/hal/lib/common/hal.cc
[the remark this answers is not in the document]
craigsdell: do you want to try and pull these two commits or do you want me to make another to pull it out?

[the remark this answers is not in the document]
pudelkoM: Whatever is easier for you. In the end, we squash all commits anyway.
Just make sure to keep the changes somewhere to submit them in another PR.

stratum/hal/lib/common/phaldb_service.cc
```diff
@@ -0,0 +1,454 @@
...
+      case ::stratum::hal::phal::Update::kPath: {
+        attr_res = ToPhalDBPath(update.path());
+        break;
+      }
+      default:
+        attr_res = MAKE_ERROR(ERR_INVALID_PARAM) << "Invalid update query";
+    }
+
+    if (!attr_res.ok()) {
+        LOG(ERROR) << "Set update " << std::to_string(i) << " failed: "
```
pudelkoM: 1. `int` converts automatically when stream-redirected

2. Instead of logging the position of the update, log the update itself: `update.ShortDebugString`

stratum/hal/lib/common/phaldb_service.cc
```diff
@@ -0,0 +1,454 @@
...
+}
+
+::grpc::Status PhalDBService::Subscribe(::grpc::ServerContext* context,
+    const ::stratum::hal::phal::SubscribeRequest* req,
+    ::grpc::ServerWriter<::stratum::hal::phal::SubscribeResponse>* stream) {
+
+  RETURN_IF_NOT_AUTHORIZED(auth_policy_checker_, PhalDBService,
+                           Subscribe, context);
+
+  ::util::Status status;
```
pudelkoM: Declare on first use, i.e. Line 358

stratum/hal/lib/phal/onlp/onlpphal.cc
```diff
@@ -131,20 +132,27 @@ ::util::Status OnlpPhal::HandleTransceiverEvent(const TransceiverEvent& event) {
...
+  RETURN_IF_ERROR(it->second->HandleEvent(event.state));
 
   // Write the event to each writer
   return WriteTransceiverEvent(event);
 }
 
+::util::StatusOr<AttributeDatabaseInterface*> OnlpPhal::GetPhalDB() {
+  if (database_ == nullptr) {
+    RETURN_ERROR() << "database_ is nullptr";
+  }
```
pudelkoM: This could be simplified to `CHECK_RETURN_IF_FALSE(database_ != nullptr) << "no database";`

stratum/hal/lib/phal/sfp_adapter.h
pudelkoM: Well, depends on who wrote this code. If it's yours then go with DELL EMC, else `Copyright 2019 Open Networking Foundation` should do it.

pudelkoM:
I decided to ignore any formatting issues and clear them as part of #48.

stratum/hal/lib/common/phaldb_service.cc
craigsdell: actually needs to be declared outside the if statement that follows.

stratum/hal/lib/phal/onlp/onlpphal.cc
craigsdell: ok thanks

stratum/hal/lib/common/phaldb_service.cc
pudelkoM: Are you sure? I see `result` being used, but not status.

stratum/hal/lib/common/hal.cc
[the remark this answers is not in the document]
craigsdell: I've created another commit to remove this change.

stratum/hal/lib/common/phaldb_service.cc
craigsdell: good suggestion, done

stratum/hal/lib/common/phaldb_service.cc
craigsdell: line 422, 
// Convert to grpc status and return
  return ToGrpcStatus(status, {});

status is set inside the "if statement" and then ToGrpcStatus converts it to a gRPC status and returns it outside the "if statement"?  Guess I could move the definition down to line 330 (i.e. first use).

pudelkoM: I missed that. Yes, right before the loop would be good.

stratum/hal/lib/common/phaldb_service.cc
```diff
@@ -0,0 +1,454 @@
...
+        //       the stream and channel for changes but for now this
+        //       will do.
+        do {
+            ::stratum::hal::phal::PhalDB phaldb_resp;
+            int code = reader->Read(&phaldb_resp,
+                                    absl::InfiniteDuration()).error_code();
+
+            // Exit if the channel is closed
+            if (code == ERR_CANCELLED) {
+                status = MAKE_ERROR(ERR_INTERNAL)
```
pudelkoM: Don't overwrite the previous status. `APPEND_STATUS_IF_ERROR` collects all of them.

Example: https://github.com/stratum/stratum/blob/a3e56b36d8a312e19ab929d0b39814382f38dd26/stratum/hal/lib/bcm/bcm_packetio_manager.cc#L259

stratum/hal/lib/common/phaldb_service.cc
```diff
@@ -0,0 +1,454 @@
...
+
+  // Save status
+  status = result.status();
+
+  // If conversion worked
+  if (result.ok()) {
+    auto path = result.ConsumeValueOrDie();
+
+    // Create writer and reader channels
+    std::shared_ptr<Channel<::stratum::hal::phal::PhalDB>> channel =
```
pudelkoM: No full path required. Same next line

craigsdell: ok, there's a few more I'll find and change too

stratum/hal/lib/common/phaldb_service.cc
```diff
@@ -0,0 +1,425 @@
...
+    {
+        // Lock subscriber channels
+        absl::MutexLock l(&subscriber_thread_lock_);
+
+        // Save channel to subscriber channel map
+        subscriber_channels_[pthread_self()] = channel;
+    }
+
+    auto writer =
+            ChannelWriter<phal::PhalDB>::Create(channel);
```
Yi-Tseng: unused variable?

stratum/hal/lib/common/phaldb_service.h
```diff
@@ -0,0 +1,118 @@
...
+
+ private:
+  // Determines the mode of operation:
+  // - OPERATION_MODE_STANDALONE: when Hercules stack runs independently and
+  // therefore needs to do all the SDK initialization itself.
+  // - OPERATION_MODE_COUPLED: when Hercules stack runs as part of Sandcastle
+  // stack, coupled with the rest of stack processes.
+  // - OPERATION_MODE_SIM: when Hercules stack runs in simulation mode.
+  // Note that this variable is set upon initialization and is never changed
+  // afterwards.
```
Yi-Tseng: Replace `Hercules` to Stratum

stratum/hal/lib/common/phaldb_service_test.cc
```diff
@@ -0,0 +1,861 @@
...
+  std::unique_ptr<PhalDBService> phaldb_service_;
+  std::unique_ptr<::grpc::Server> server_;
+  std::unique_ptr<::stratum::hal::phal::PhalDBSvc::Stub> stub_;
+  std::unique_ptr<PhalMock> phal_mock_;
+  std::unique_ptr<::stratum::hal::phal::AttributeDatabaseMock> database_mock_;
+};
+
+TEST_P(PhalDBServiceTest, SetupWarm) {
+  auto status = phaldb_service_->Setup(true);
+  EXPECT_TRUE(status.ok());
```
Yi-Tseng: Maybe `EXPECT_OK(phaldb_service_->Setup(true));` ?

stratum/hal/lib/common/phaldb_service_test.cc
```diff
@@ -0,0 +1,861 @@
...
+
+  req.set_str("cards[0]/ports[0]/transceiver/");
+
+  EXPECT_CALL(*auth_policy_checker_mock_, Authorize("PhalDBService", "Get", _))
+      .WillOnce(Return(::util::OkStatus()));
+
+  // Invoke the RPC and validate the results.
+  // Call and validate results.
+  auto status = stub_->Get(&context, req, &resp);
+  ASSERT_TRUE(status.ok());
```
Yi-Tseng: We can use `EXPECT_OK` for these kinds of checks, same as below

stratum/hal/lib/common/phaldb_service.cc
craigsdell: look on line 356, it's passed down to the Subscribe request and then used to pass messages back to this process via the channel.

stratum/hal/lib/common/phaldb_service.h
craigsdell: ok done

stratum/hal/lib/common/phaldb_service_test.cc
craigsdell: had some conflict with the  stratum/glue/status/status_test_util.h, have now removed that and EXPECT_OK now works

stratum/hal/lib/common/phaldb_service_test.cc
craigsdell: done

PR #48, opened 2019-11-26
Add C++ style check

### LINKED ISSUE #44: Enable cpplint and clang_format checks in CI
At least on newly committed files.

Yi-Tseng:
We use cpplint tool to check every .h and .cc file.

Headers and c++ code should match all rules from cpplint (which based on Google C++ style guide) except:

 - Maximum columns will be 110 instead of 80
 - Some implicit single-argument constructures
 - Unapproved C++11 headers such as thread, chrono, ....

This patch also includes some code style fixes

Closes #44


pudelkoM:
> Maximum columns will be 110 instead of 80
I don't agree here. This would require changing our entire code base from the Google standard of 80. I'd rather review the sdk_wrapper once.

> Some implicit single-argument constructures
I don't understand this point. Are we ignoring this specific issue or do we just allow use of `// NOLINT` comments

.circleci/config.yml
```diff
@@ -167,7 +167,10 @@ jobs:
       - checkout
       - run:
           name: Run cpplint
-          command: cpplint --linelength=110 --recursive stratum
+          command: cpplint --recursive stratum
+      - run:
+          name: Run clang-format
+          command: clang-format --style=file --output-replacements-xml stratum | ! grep "<replacement "
```
pudelkoM: clang-format is not recursive

PR #51, opened 2019-12-02
Added port map for stordis-bf2556 switch.

kamalkrbh:
Added port map for stordis-bf2556 (25G) switch.


stratum/hal/bin/barefoot/platforms/x86-64-stordis-bf2556x-1t-r0.json
```diff
@@ -0,0 +1,1141 @@
...
+          "tx_eq_pre": 2
+        }, 
+        "tx_lane": 3, 
+        "tx_pn_swap": 0
+      }, 
+      "mac_block": 10, 
+      "media_type": "optical"
+    }, 
+    {
+      "connector": 21,
```
Yi-Tseng: Looks like there are only 21 connectors(ports) defined in this port map file. However, there are 56 ports on the switch.
Is it possible to add more port/connector support to this file?

stolsma: I have the same comment! The file doesnt look right, missing lots of ports. By trail and error we created our own version and it looks totally different but we dont know if ours is correct but it works though...

Yi-Tseng: I actually got a similar json port map file (21 connectors) when I tried to generate the port map file from BSP header file.
@kamalkrbh can you help me confirm that we have correct BSP header file?

kamalkrbh: Yes the file is correct. Actually our 25G switch is different from our 100G one.
This 25G switch have (20+1)x100G tofino ports.

On the front panel:
ports 1-16 support 1G,2.5G,10G,25G = 4 tofino ports breakout
ports 17-48 support 10G,25G = 8 tofino ports in breakout
ports 48-56 support 10G,25G,40G,100G = 8 tofino ports breakout or non-breakout

Total 4+8+8=20 + 1(cpu port) = 21 Tofino ports

stolsma: I find it a little bit strange that no physical port mapping is found in this port mapping file. A little bit confused also because doesnt that mean that configuring ports through gNMI is strange? With our file you just use the correct port names,  how does it work with this file?

stolsma: Maybe I need to rephrase my question to: can anybody explain what the port mapping file is doing? And why it is there?

kamalkrbh: @stolsma For port mapping I will reach out HW team once again, 
In regards to "Why it is there" would like to know isn't it the right place to keep the port mapping files required to run in bsp-less mode ? BTW I referred location of x86-64-stordis-bf6064x-t-r0.json for our 100G switch which is already pushed in this location.

stolsma: @kamalkrbh its not about the location of the file, with 'why its there' my question was if someone could explain why it is needed if it doesnt do 'real port mapping'...

Yi-Tseng: @stolsma Usually the port mapping (from physical to logical) is done by BSP.
We need to provide this mapping if we want to use BSP-less mode.

@kamalkrbh I see, I tried to use 1G SFP-to-Cat6 adapter. I connect to port 1 and port 2. But looks like in Tofino you maps these two port to `21/0` and `22/0`. Is it correct?

kamalkrbh: @Yi-Tseng Yes, Correct in-fact only 21/0.

Yi-Tseng: @kamalkrbh I still trying to use 1G SFP-to-CAT6 adapter. With and without BSP support.
However the 1G port never changes to `UP` state. I am using ONL with ONLPv2 you provided and port map file you provided here.
Should I set up anything(e.g., enable special breakout device?) before I start the software?

kamalkrbh: @Yi-Tseng Ports with 1G support (1-16) are not directly connected to Tofino, instead via a GearBox, you need driver for GearBox, do you have something called "AppDemo" in SW packages provided by us ?

Yi-Tseng: @kamalkrbh I think we don't have the `AppDemo` software. Can you send it to us by mail? Thanks

kamalkrbh: Port mapping JSON is updated.

PR #55, opened 2019-12-06
Enable loopback option to BCM target

Yi-Tseng:
Adds loop back option to both chassis config and BCM chassis map


Yi-Tseng:
conflict with #157

PR #60, opened 2019-12-10
Ignore SerDes config errors

pudelkoM:
It prevents startup and we don't support it in the SDK at the moment anyway.


Yi-Tseng:
Rebased on top of master (reduce CPU for CI)

PR #63, opened 2019-12-12
ONLP: Fix SFP info strings

pudelkoM:
ONLPv2 reports strings with trailing spaces, this commit adds getter functions that strip these.

### DIFF
--- a/stratum/hal/lib/phal/onlp/onlp_wrapper.cc
+++ b/stratum/hal/lib/phal/onlp/onlp_wrapper.cc
@@ -334,7 +334,7 @@ void SfpInfo::GetModuleCaps(SfpModuleCaps* caps) const {
 
 ::util::StatusOr<const SffInfo*> SfpInfo::GetSffInfo() const {
   CHECK_RETURN_IF_FALSE(sfp_info_.sff.sfp_type != SFF_SFP_TYPE_INVALID)
-      << "Cannot get SFF info: Invalid SFP type: " << sfp_info_.sff.sfp_type;
+      << "Cannot get SFF info: Invalid SFP type.";
   return &sfp_info_.sff;
 }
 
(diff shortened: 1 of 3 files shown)


stratum/hal/lib/phal/onlp/onlp_wrapper.cc
```diff
@@ -313,7 +334,7 @@ void SfpInfo::GetModuleCaps(SfpModuleCaps* caps) const {
 
 ::util::StatusOr<const SffInfo*> SfpInfo::GetSffInfo() const {
   CHECK_RETURN_IF_FALSE(sfp_info_.sff.sfp_type != SFF_SFP_TYPE_INVALID)
-      << "Cannot get SFF info: Invalid SFP type.";
+      << "Cannot get SFF info: Invalid SFP type: " << sfp_info_.sff.sfp_type;
```
pudelkoM: This would could be removed since the check is already against `SFF_SFP_TYPE_INVALID`

Yi-Tseng: Should we also modify this check?
https://github.com/stratum/stratum/blob/master/stratum/hal/lib/phal/onlp/sfp_datasource.cc#L82-L92

pudelkoM: I have modified ` SfpInfo::GetSffInfo()` directly:

```diff
 ::util::StatusOr<const SffInfo*> SfpInfo::GetSffInfo() const {
-  CHECK_RETURN_IF_FALSE(sfp_info_.sff.sfp_type != SFF_SFP_TYPE_INVALID)
-      << "Cannot get SFF info: Invalid SFP type: " << sfp_info_.sff.sfp_type;
+  // CHECK_RETURN_IF_FALSE(sfp_info_.sff.sfp_type != SFF_SFP_TYPE_INVALID)
+  //     << "Cannot get SFF info: Invalid SFP type: " << sfp_info_.sff.sfp_type;
+  // FIXME: Some broken DACs don't report a correct type.
+  if (sfp_info_.sff.sfp_type != SFF_SFP_TYPE_INVALID) {
+    LOG(ERROR)
+      << "Cannot get SFF info: Invalid SFP type.";
+  }
   return &sfp_info_.sff;
 }
```

PR #67, opened 2019-12-16
Bump version of PI, bmv2, and P4Runtime

ccascone:
Includes fixes to multicast group handling in PI and support for P4 logging
primitives in bmv2.

To make the PI build pass, the version of P4Runtime has been bumped to
1.1.0-rc.1, which should not break backward-compatibility at the P4Runtime
message level.

Changes have been tested end-to-end with stratum_bmv2 on Mininet with ONOS and
Trellis, using the instructions under tools/mininet/examples/trellis. The ONOS
version used by the example was bumped to 2.2.0 which is an LTS release and should be
preferred to 2.1.0.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -15,8 +15,8 @@
 #
 
 ARG BAZEL_VERSION=0.23.2
-ARG PI_COMMIT=0bcaeda2269a4f2f0539cf8eac49868e389a8c18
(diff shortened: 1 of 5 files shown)


ccascone:
@Yi-Tseng done. I put a reference to `Dockerfile.build` instead of duplicating the same commit IDs, which is harder to maintain over time.

Yi-Tseng:
@ccascone Make sense, thanks

PR #72, opened 2019-12-17
Initial TAI support

maksym-tropets-plvision:


Yi-Tseng:
> Unless the gNMI cli is somehow needed by the TAI code, it has to be a separate PR.
> 
> On that note: I agree that Stratum should have a gNMI cli to query values, but unless there is a strong argument for using Golang, it will be in C++.

I think we can also move the current python-based gNMI CLI to another repo.

pudelkoM:
> > Unless the gNMI cli is somehow needed by the TAI code, it has to be a separate PR.
> > On that note: I agree that Stratum should have a gNMI cli to query values, but unless there is a strong argument for using Golang, it will be in C++.
> 
> I think we can also move the current python-based gNMI CLI to another repo.

Didn't even know we already had one. That more reason to not add another one. Either extent the python cli or replace it (possibly by a Golang implementation).

Dockerfile.dev
```diff
@@ -27,6 +27,16 @@ RUN useradd -ms /bin/bash -l -u $USER_ID $USER_NAME && \
...
     echo "source /usr/local/lib/bazel/bin/bazel-complete.bash" >> /home/$USER_NAME/.bashrc
 
+RUN apt-get update && apt-get install -y python3 python3-pip \
+    gnupg wget software-properties-common
+
+RUN pip3 install prompt_toolkit jinja2 clang
+
+RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+RUN apt-add-repository "deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch-6.0 main"
+RUN apt-get update && apt-get install -y libclang-6.0-dev
```
Yi-Tseng: We already installed clang in the based build image (Dockerfile.build). You can specify the clang library when you invoke the main.py script

bazel/external/tai.BUILD
```diff
@@ -0,0 +1,46 @@
...
+        ],
+    cmd = "make -C $(location :stub) && cp $(location :stub)/libtai.so \
+    $(@D)",
+    visibility = [ '//visibility:public' ],
+    # TODO: need to find solution how to use tai directories in the 'cmd' script
+    tools = ["stub"]
+)
+
+genrule(
+    name = "meta_build",
```
Yi-Tseng: I'd suggest creating a `py_binary` rule for `main.py` so we don't need to install additional python libraries to the base image.

stratum/hal/lib/common/common.proto
```diff
@@ -250,6 +252,59 @@ enum FecMode {
...
+  ywrapper.Decimal64Value min = 6;
+  // The absolute time at which the minimum value occurred.
+  uint64 min_time = 7;
+}
+
+// NOTE. The name 'LazerFrequency' was chosen to avoid the conflict with
+// 'gnmi_publisher.cc:Frequency'.
+//
+// Optical channel frequency in MHz.
+message LazerFrequency {
```
bocon13: Do we mean Laser instead of Lazer? Or, does "lazer" have a special meaning in TAI?

Dockerfile.dev
bohdan-oheruk-plvision: Yes, I saw that you have already installed clang in(Dockerfile.build) but module (https://pypi.org/project/clang/) that is required by the "main.py" script expects clang with concrete version(up to 6.0.0, https://github.com/ethanhs/clang/issues/4)

bazel/external/tai.BUILD
bohdan-oheruk-plvision: Correct me if I'm wrong. By creating a new `py_binary` rule, you mean to add dependencies using `pip_import` (https://github.com/bazelbuild/rules_python#importing-pip-dependencies)?
If yes then we should use a new feature for import pip3 libraries that were recently merged (https://github.com/bazelbuild/rules_python/pull/256). 
And it looks like that our current bazel version does not support their changes. I tried to update our 'io_bazel_rules_python' to the latest version and got this error "io_bazel_rules_python/python/pip.bzl:44:14: unexpected keyword 'doc', for call to function repository_rule(implementation, attrs = None, local = False, environ = [])"

maksym-tropets-plvision:
> > > Unless the gNMI cli is somehow needed by the TAI code, it has to be a separate PR.
> > > On that note: I agree that Stratum should have a gNMI cli to query values, but unless there is a strong argument for using Golang, it will be in C++.
> > 
> > 
> > I think we can also move the current python-based gNMI CLI to another repo.
> 
> Didn't even know we already had one. That more reason to not add another one. Either extent the python cli or replace it (possibly by a Golang implementation).
> 
> @Yi-Tseng found an old C++ implementation and I submitted it as WIP in #74

        	The Go CLI has been added because we haven't had the working Python CLI in
		our mirror version. As soon as a new pull request (link to C++ CLI merge
		request) has been created, I don't see any reason to keep Go CLI too. So, reverted.

maksym-tropets-plvision:
> I believed that the dummy switch is designed for testing core components (hal, switch interface, ...)
> Please move TAI to a new target instead of putting it to dummy switch.
> I'd suggest add new target to bcm since we are targeting bcm based box with TAI support.

Currently, we don't have physical access to the Broadcom switch. We've planned to develop the feature for the dummy switch to be able to concentrate on gNMI and TAI design. Afterward, we've assumed to move the implementation to the Broadcom switch target and solve all hardware related problems during that migration.

bazel/external/tai.BUILD
Yi-Tseng: Actually Python2 works. You can try using this rules.
See this commit: https://github.com/Yi-Tseng/oopt-tai/commit/8697145db0a6ce47e2e6483680699c8cc2399ce5

stratum/hal/lib/common/common.proto
maksym-tropets-plvision: Nothing special for TAI case, just a spelling mistake. We mean "laser" here.
Fixed.

bazel/external/tai.BUILD
bohdan-oheruk-plvision: I tried your example, and faced the following error:

`clang.cindex.TranslationUnitLoadError: Error parsing translation unit.
Target //stratum/hal/bin/dummy:stratum_dummy failed to build`.

Looks like main.py inc/tai.h arg can't be found during 'dummy_switch' compilation.
That is caused by hardcoded paths to the file arguments for 'main.py'.

Your example is great, so I made some [fixes and improvements](https://github.com/Yi-Tseng/oopt-tai/pull/1) for your commit to avoid compilation errors.
And here in Stratum, I have updated bazel make the system to be synced with your repo temporarily.
Is my understanding proper that you might push these changes to the official TAI repo?
If yes, then I will update bazel in Stratum to use TAI official repo once these changes are in there.

bohdan-oheruk-plvision: Please let me know if you are ok with the changes.
If you don’t have any concerns I would like to push these changes to the official TAI repo and update Stratum with the reference to official repo instead of mine.

Yi-Tseng: Hi @bohdan-oheruk-plvision, sorry for the delay. I was traveling last week.
The patch from you looks good to me. I think we can use your changes in Stratum before they add Bazel changes to upstream. Also, feel free to push those changes to the official repo.

pudelkoM:
To me the code structure of this TAI implementation is not completely clear yet.
There seems to be a lot of code interfacing with TAI somehow. We generally contain such code to a single class (`*_wrapper`). 

To ilustrate my point I started the diagram below ([Source](https://docs.google.com/drawings/d/1xcwKeCeMJ3-fWhNKKdnVE1QhFRbjn7o_m5_TRBdxqhI/edit?usp=sharing)). Please correct and fill the gaps where necessary. We need some kind of documentation anyway.

arch

stratum/hal/lib/tai/module.cc
```diff
@@ -0,0 +1,155 @@
...
+  tai_attribute_t attr;
+  tai_status_t status;
+
+  attr.id = TAI_MODULE_ATTR_LOCATION;
+  attr.value.charlist.count = static_cast<uint>(location.size());
+  attr.value.charlist.list = const_cast<char*>(location_.c_str());
+  list.push_back(attr);
+
+  char str[] = "my vendor";
+  attr.id = TAI_MODULE_ATTR_VENDOR_NAME;
```
Andrea-Campanella: As per Wataru's comments:
https://github.com/Telecominfraproject/oopt-tai/blob/master/inc/taimodule.h#L119
vendor name is a readonly attribute. You shouldn't set it when creating the module.
The stub library which you're using doesn't do the validation.

stratum/hal/lib/common/common.proto
```diff
@@ -250,6 +252,59 @@ enum FecMode {
...
 
+// The input optical power of this port in units of 0.01 decibel-milliwatt.
+// If the port is an aggregate of multiple physical channels, this attribute
+// is the total power or sum of all channels.
+//
+// The time-based values are the timestamps in nanoseconds relative to the
+// Unix Epoch (Jan 1, 1970 00:00:00 UTC)."
+message InputPower {
+  // The instantaneous value of the statistic.
+  ywrapper.Decimal64Value instant = 1;
```
pudelkoM: Is there a strong need to have this `Decimal64Value` type?

Current Stratum practice is to use a scalar type and have an implicit type/scale: https://github.com/stratum/stratum/blob/d026ea398ee8c5b6b5adfd9b2eea596446172115/stratum/hal/lib/phal/db.proto#L135

Without this all the adapter code could be dropped too.

maksym-tropets-plvision: Unfortunately, gNMI doesn't support plain double: https://github.com/openconfig/gnmi/blob/master/proto/gnmi/gnmi.proto#L108.

It was agreed in Slack chat to introduce the new "Decimal64" type into the common.proto. Now, I've already pushed that for review.

Yi-Tseng: I just noticed that we convert `Decimal64` to `tai_float_t` which is a `float` type. 
https://github.com/Telecominfraproject/oopt-tai/blob/master/inc/taitypes.h#L54

I agreed to add `Decimal64` in Slack chat because I thought we need a high precision number in TAI. But looks like we only need `float` level precision. Do we have a strong need to have this type if the `float` type is enough?

stratum/hal/lib/tai/taimanager.cc
```diff
@@ -0,0 +1,265 @@
...
+ * \param request it's gNMI request value that contains what value will be
+ * returned
+ * \param module_netif_pair it's pair of module id and related to it networkif
+ * id interface
+ * \return valid DataResponse or invalid ::util::Status
+ */
+::util::StatusOr<DataResponse> TAIManager::GetValue(
+    const DataRequest::Request& request,
+    const std::pair<uint64, uint32>& module_netif_pair) const {
+  LOG(INFO) << __FUNCTION__;
```
Yi-Tseng: Use DEBUG or remove this log if it is for debugging only

Yi-Tseng: Same as below

stratum/hal/lib/tai/taiadapterhost.cc
```diff
@@ -0,0 +1,252 @@
...
+  std::weak_ptr<Module> weak_module_ptr =
+      GetModule(objectPath.front().object_index);
+  if (weak_module_ptr.expired()) {
+    LOG(WARNING) << "Invalid object, std::weak_ptr expired";
+    return {};
+  }
+
+  std::shared_ptr<Module> module = weak_module_ptr.lock();
+  if (objectPath.size() > 1) {
+    std::weak_ptr<TAIObject> tai_object;
```
Yi-Tseng: unused variable?

stratum/hal/lib/tai/test/tai_manager_test.cc
```diff
@@ -0,0 +1,138 @@
...
+      ->set_digits(kDigits);
+  set_request.mutable_port()
+      ->mutable_output_power()
+      ->mutable_instant()
+      ->set_precision(kPrecision);
+
+  EXPECT_TRUE(TAIManager::IsRequestSupported(set_request));
+
+  auto kStatus = manager_->SetValue(set_request, module_netif_pair);
+  EXPECT_TRUE(kStatus.ok());
```
Yi-Tseng: Is there a mock class for the TAI adapter? I tried to trace this code but cannot find mocked value for module 0 interface 0.

Yi-Tseng: Same as other tests

stratum/hal/lib/tai/test/tai_manager_test.cc
```diff
@@ -0,0 +1,138 @@
...
+#include "stratum/hal/lib/tai/networkinterface.h"
+#include "stratum/hal/lib/tai/taiadapterhost.h"
+#include "stratum/hal/lib/tai/taimanager.h"
+
+namespace stratum {
+namespace hal {
+namespace tai {
+
+class TAIManagerTest : public ::testing::Test {
+ public:
```
pudelkoM: `protected`

stratum/hal/lib/tai/test/tai_manager_test.cc
```diff
@@ -0,0 +1,138 @@
...
+  static void SetUpTestCase() { manager_ = TAIManager::Instance(); }
```
pudelkoM: `SetUpTestCase()` and `TearDownTestCase()` are deprecated. Use `SetUp()` and `TearDown()`.

https://github.com/google/googletest/blob/78fdd6c00b8fa5dd67066fbb796affc87ba0e075/googletest/include/gtest/gtest.h#L426-L429

stratum/hal/lib/tai/test/tai_manager_test.cc
```diff
@@ -0,0 +1,138 @@
...
+  static void TearDownTestCase() {
+    manager_ = nullptr;
+    TAIManager::Delete();
+  }
+
+  static TAIManager *manager_;
```
pudelkoM: Don't make the manager a static raw pointer. Current practice is using a `protected std::unique_ptr<TAIManager> manager_;`

stratum/hal/lib/common/common.proto
maksym-tropets-plvision: I was referring to the OpenConfig model, which considers the decimal type there.

Do we rely on the OpenConfig YANG model in this situation?
Or is it better to go with the "float" and cause the precision loss explicitly?

pudelkoM: Won't we lose precision anyway at some point? Yi suggest doing the conversion earlier (p4_service.cc probably) rather than later (TAIManager.cc) and save us the additional data type.

maksym-tropets-plvision: Changed to
* `Decimal64` from the gNMI side
* and `float` from the Stratum side.

Please, review.

stratum/hal/lib/common/decimal_types_utils.h
```diff
@@ -0,0 +1,172 @@
+#ifndef STRATUM_HAL_LIB_COMMON_DECIMAL_TYPES_UTILS_H
```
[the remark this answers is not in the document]
maksym-tropets-plvision: Added license header. Please, review.

stratum/hal/lib/common/decimal_types_utils_test.cc
```diff
@@ -0,0 +1,125 @@
+#include "stratum/hal/lib/common/decimal_types_utils.h"
```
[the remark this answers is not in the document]
maksym-tropets-plvision: Added license header. Please, review.

stratum/hal/lib/tai/BUILD
```diff
@@ -0,0 +1,73 @@
...
+
+package(
+    default_visibility = STRATUM_INTERNAL,
+)
+
+load("@com_github_grpc_grpc//bazel:cc_grpc_library.bzl", "cc_grpc_library")
+
+stratum_cc_library(
+    name = "tai",
+    srcs = ["taimanager.cc",
```
pudelkoM: Place list items on new line.

stratum/hal/lib/tai/taiadapterhost.cc
```diff
@@ -0,0 +1,252 @@
...
+
+namespace stratum {
+namespace hal {
+namespace tai {
+
+/*!
+ * \brief modules_location variable used to store returned from TAI module
+ * location
+ */
+static std::vector<std::string> modules_location;
```
pudelkoM: TAI readme says that the `module_presence` callback has to be guarded by locks.

stratum/hal/lib/tai/taiadapterhost.cc
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  status =
+      tai_api_query(TAI_API_HOSTIF, reinterpret_cast<void**>(&api_.hostif_api));
+  if (TAI_STATUS_SUCCESS != status) {
+    LOG(ERROR) << "Failed to query HOSTIF API" << std::endl;
+    return;
+  }
+
+  for (const auto& location : modules_location) {
```
pudelkoM: You can't be sure that the presence callback as been called here. This must happen later. I'd make that a separate function.

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -2818,6 +2818,20 @@ void SetUpComponentsComponentName(const std::string& name, TreeNode* node) {
...
+    const std::string& name, TreeNode* node) {
+  auto poll_functor = [name](const GnmiEvent& /*event*/,
+                             const ::gnmi::Path& path,
+                             GnmiSubscribeStream* stream) {
+    return SendResponse(GetResponse(path, "OPTICAL_CHANNEL"), stream);
+  };
+
+  node->SetOnPollHandler(poll_functor)
+      ->SetOnTimerHandler(poll_functor);
+}
```
Yi-Tseng: Add new parameter `std::string& type` and pass this parameter to `SendResponse` call. So we can reuse this function in the future for other components.

maksym-tropets-plvision: Fixed. Please review.

stratum/hal/lib/dummy/dummy_phal.cc
```diff
@@ -12,34 +12,101 @@
...
 #include "stratum/glue/status/status.h"
+#include "stratum/hal/lib/common/constants.h"
+#include "stratum/lib/macros.h"
+
+#if defined(WITH_TAI)
+#include "stratum/hal/lib/phal/tai/tai_phal.h"
+#include "stratum/hal/lib/phal/tai/tai_switch_configurator.h"
+#include "stratum/hal/lib/phal/tai/tai_wrapper/tai_manager.h"
+#include "stratum/lib/utils.h"
+#endif  // defined(WITH_TAI)
```
pudelkoM: These kinds of includes/ifdefs should be part of the generic PHAL.

stratum/hal/lib/common/common.proto
```diff
@@ -995,6 +1092,15 @@ message FrontPanelPortInfo {
   HwState hw_state = 6;
 }
 
+// Optical channel configuration for TAI module.
+message TaiOpticalChannelInfo {
```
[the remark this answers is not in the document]
maksym-tropets-plvision: Fixed. Please review.

pudelkoM:
Compile error when using this: `bazel build --define bcm_sdk=lt --define phal_with_tai=false //stratum/hal/bin/bcm/standalone:stratum_bcm`

```In file included from stratum/hal/lib/phal/phal.cc:15:
./stratum/hal/lib/phal/phal.h:32:10: fatal error: 'stratum/hal/lib/phal/optics_adapter.h' file not found
#include "stratum/hal/lib/phal/optics_adapter.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.```

Adding `":optics_adapter"` to the phal deps should fix it.

File: stratum/hal/lib/bcm/BUILD [hunk 20]
```diff
@@ -156,6 +156,7 @@ stratum_cc_library(
         ":bcm_sdk_interface",
         ":bcm_serdes_db_manager",
         ":utils",
+        "//stratum/hal/lib/phal/tai/tai_wrapper:tai",
```
pudelkoM: No phal code/deps outside of phal!

stratum/hal/lib/bcm/bcm_chassis_manager.cc
```diff
@@ -153,6 +153,26 @@ BcmChassisManager::~BcmChassisManager() {
...
+/*!
+ * \brief BcmChassisManager::IsNodePortIdRelatedToTAI method check is (node_id,
+ * port_id) pair \param node_port_id related to TAI
+ * \return true if related, false if not related
+ */
+bool BcmChassisManager::IsNodePortIdRelatedToTAI(
+    const std::pair<uint64, uint32>& node_port_id) const {
+  const auto kIterator = node_port_id_to_module_netif.find(node_port_id);
+  return kIterator != node_port_id_to_module_netif.end();
+}
```
pudelkoM: I think this function is not needed. If a port is not related to TAI the specific function will fail and we can report this then.

stratum/hal/lib/phal/tai/tai_wrapper/tai_manager.h
```diff
@@ -0,0 +1,280 @@
...
+  if (return_code != TAI_STATUS_SUCCESS) {
+    std::stringstream error_msg;
+    error_msg << "Can't get the attribute. TAI return code: " << return_code;
+
+    LOG(ERROR) << error_msg.str();
+    RETURN_ERROR(ERR_INTERNAL) << error_msg.str();
+  }
+
+  return ::util::StatusOr<T>(TaiAttributeToResponse<T>(tai_attr));
+}
```
Yi-Tseng: move implememtations to `.cc` file, same as below

stratum/hal/lib/phal/tai/tai_optics_datasource.cc
```diff
@@ -0,0 +1,129 @@
...
+
+/*!
+ * \brief TaiOpticsDataSource::UpdateValues method updates attributes with 
+ * fresh values from TAI.
+ * \return ::util::OkStatus if success
+ */
+::util::Status TaiOpticsDataSource::UpdateValues() {
+  ASSIGN_OR_RETURN(auto tx_laser_frequency,
+                   tai_manager_->GetValue<uint64>(
+                       TAI_NETWORK_INTERFACE_ATTR_TX_LASER_FREQ, {0, 0}));
```
Yi-Tseng: The module/interface pare are hardcoded. Should pass the module/interface-id when we create this data source. Same as below

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -87,6 +89,35 @@ ::gnmi::SubscribeResponse GetResponse(const ::gnmi::Path& path,
...
 
+// Specialization for '::gnmi::Decimal64'.
+::gnmi::SubscribeResponse GetResponse(const ::gnmi::Path& path,
+                                      const ::gnmi::Decimal64& contents) {
+  ::gnmi::Decimal64* value = TypedDecimalInitializer<::gnmi::Decimal64>(
+      contents.digits(), contents.precision()).InitAllocated();
+
+  ::gnmi::SubscribeResponse resp = GetResponse(path);
+  resp.mutable_update()->mutable_update(0)->mutable_val()
+      ->set_allocated_decimal_val(value);
```
pudelkoM: Can be replaced by
```c++
::gnmi::SubscribeResponse resp = GetResponse(path);
resp.mutable_update()->mutable_update(0)->mutable_val()->mutable_decimal_val() = contents;
```

stratum/hal/lib/phal/BUILD
```diff
@@ -91,7 +91,7 @@ phal_defines = select({
 }) + select({
     ":with_tai": tai_defines,
     ":without_tai": [],
-    "//conditions:default": [],
+    "//conditions:default": tai_defines,
```
pudelkoM: Default will be no TAI. Most platforms don't have optical ports.

stratum/hal/lib/common/yang_parse_tree.cc
```diff
@@ -172,6 +173,22 @@ ::util::Status YangParseTree::ProcessPushedConfig(
     port_id_to_node_id[singleton.id()] = singleton.node();
     singleton_names.insert(singleton.name());
   }
+
+  std::unordered_set<std::string> optical_names;
```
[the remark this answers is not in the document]
maksym-tropets-plvision: Yeap, I've used the `unordered_set` in order not to break the code symmetry. Please resolve if you agree.

stratum/hal/lib/common/yang_parse_tree_paths.cc
maksym-tropets-plvision: The `contents` argument is not `T*` but `const T&`.
Could you please explain more detailed about what do you mean here?

maksym-tropets-plvision:
> I believed that the dummy switch is designed for testing core components (hal, switch interface, ...)
> Please move TAI to a new target instead of putting it to dummy switch.
> I'd suggest add new target to bcm since we are targeting bcm based box with TAI support.

Fixed by moving ports mapping to TAI Phal. Please note, that tai phal related code is enabled using preprocessor. That's why I've used std::funtction to put the part of the code in there. Please review.

stratum/hal/lib/bcm/bcm_switch.cc
```diff
@@ -401,6 +404,30 @@ ::util::Status BcmSwitch::RetrieveValue(uint64 /*node_id*/,
...
+        const std::pair<uint64, uint32> module_network_id =
+            bcm_chassis_manager_->GetModuleNetworkIds(
+                node_port_id.first, node_port_id.second);
+
+        ::util::Status status = phal_interface_->GetOpticalTransceiverInfo(
+            module_network_id.first, module_network_id.second,
+            resp_val.mutable_tai_optical_channel_info());
+        if (status.ok()) {
+          resp = resp_val;
+        }
```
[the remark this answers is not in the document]
maksym-tropets-plvision: Fixed by moving ports mapping to TAI Phal. Please note, that tai phal related code is enabled using preprocessor. That's why I've used std::funtction to put the part of the code in there. Please review.

stratum/hal/lib/bcm/BUILD
maksym-tropets-plvision: Fixed. Please review.

stratum/hal/lib/bcm/bcm_chassis_manager.cc
maksym-tropets-plvision: I've changed the logic a little and removed that function. Please review

stratum/hal/lib/phal/BUILD
maksym-tropets-plvision: That was my attempt to fix the build on CI. Though, it's fixed now through correct target deps, so removed this. Please review.

maksym-tropets-plvision:
> Compile error when using this: `bazel build --define bcm_sdk=lt --define phal_with_tai=false //stratum/hal/bin/bcm/standalone:stratum_bcm`
> 
> ```
> ./stratum/hal/lib/phal/phal.h:32:10: fatal error: 'stratum/hal/lib/phal/optics_adapter.h' file not found
> #include "stratum/hal/lib/phal/optics_adapter.h"
>          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 1 error generated.```
> 
> Adding `":optics_adapter"` to the phal deps should fix it.
> ```

Added `":optics_adapter"` to the phal deps. Build is successful. Please review.

stratum/hal/lib/bcm/bcm_switch.cc
```diff
@@ -401,6 +402,18 @@ ::util::Status BcmSwitch::RetrieveValue(uint64 /*node_id*/,
...
+
+      case DataRequest::Request::kOpticalChannelInfo: {
+        ::util::Status status = phal_interface_->GetOpticalTransceiverInfo(
+            req.optical_channel_info().node_id(),
+            req.optical_channel_info().port_id(),
+            resp_val.mutable_optical_channel_info());
+        if (status.ok()) {
+          resp = resp_val;
+        }
+        break;
```
Yi-Tseng: It can be `status.update(...)`. `case DataRequest::Request::kPortCounters` is an example (line 333)

stratum/hal/lib/common/yang_parse_tree_paths.cc
pudelkoM: I forgot a de-reference. Here's the full function:

```c++
// Specialization for '::gnmi::Decimal64'.
::gnmi::SubscribeResponse GetResponse(const ::gnmi::Path& path,
                                      const ::gnmi::Decimal64& contents) {
  ::gnmi::SubscribeResponse resp = GetResponse(path);
  *resp.mutable_update()
      ->mutable_update(0)
      ->mutable_val()
      ->mutable_decimal_val() = contents;
  return resp;
}
```

stratum/hal/lib/common/decimal_types_utils.h
```diff
@@ -0,0 +1,189 @@
+/*
```
Yi-Tseng: Can't we use default protobuf functions to create a Decimal64 message object?
Or create a utility function or macro to initialize the object?
`InitDecimal64(&val, digits, precision);`

stratum/hal/lib/tai/tai_manager.h
```diff
@@ -0,0 +1,79 @@
+/*
```
Yi-Tseng: What is the reason we need to have another abstraction layer that converts data source calls to TAI wrapper value?
In Stratum, we have abstraction which converts from "general" to "vendor-specific".
For example, trying to get SFP information from a slot/port number is general, but how to get that information is vendor-specific.
The TAI data source is already vendor-specific, it should be able to talk to the wrapper directly.

stratum/hal/lib/phal/tai/tai_wrapper/tai_manager.h
bohdan-oheruk-plvision: This is the template method. We should either define it in the header file or directly include the .cpp file with its implementation wherever we use it. What approach should we follow in this case?

stratum/hal/lib/common/decimal_types_utils.h
maksym-tropets-plvision: > Can't we use default protobuf functions to create a Decimal64 message object?

We can, and it leads to the boilerplate code:
```
::gmni::Decimal64* value = new Decimal64Value;
value->set_digits(1);
value->set_precision(2);
```

Anyway, if using this wrappers seems an overkill to you, we can use the naive approach. Please comment.

> Or create a utility function

This file contains such utility functions as `TypedDecimalInitializer` members. The reason they are put in the class is a compile-time type checking.

> or macro

Do we prefer using macros instead of template functions?

stratum/hal/lib/tai/tai_manager.h
bohdan-oheruk-plvision: We plan to have other features, for example, collecting statistic info (for output/input-power (min, max, average)), or subscription mechanism. All of them are supposed to be placed in `TAIManager` class.

stratum/hal/lib/common/decimal_types_utils.h
Yi-Tseng: I think native way to create the Decimal64 should be fine. No need to have another wrapper or matcher.

stratum/hal/bin/dummy/main.cc
```diff
@@ -25,10 +25,25 @@ namespace stratum {
...
+void InitTAI(const std::string& exec_path) {
+  setenv("TAI_MUX_PLATFORM_ADAPTER", "static", true);
+
+  std::string static_json_path;
+  static_json_path = exec_path.substr(0, exec_path.find_last_of("\\/"));
+
+  static_json_path.append("/static.json");
+  setenv("TAI_MUX_STATIC_CONFIG_FILE", static_json_path.c_str(), true);
+}
+
```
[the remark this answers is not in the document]
Yi-Tseng: You can use gflags to pass options to TAI PHAL if you need to setup the path of json path.
this is an example: https://github.com/stratum/stratum/blob/master/stratum/hal/lib/common/config_monitoring_service.cc#L51-L56

stratum/hal/lib/tai/tai_manager.h
Yi-Tseng: Ok, should be fine if you need another component to manage the state automatically.
@pudelkoM @Andrea-Campanella any thought?

stratum/hal/lib/phal/tai/tai_wrapper/tai_manager.h
pudelkoM: I think we missed the template. It's ok to leave as is for now.

stratum/docs/tai/README.md
```diff
@@ -0,0 +1,70 @@
...
+    slot: 1
+    port: 4
+    channel: 1
+    speed_bps: 100000000000  # 100Gbps
+    node: 1
+  }
+
+  module_location: 1
+  netif_location: 0
+}
```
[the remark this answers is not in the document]
maksym-tropets-plvision: Updated. Please review.

stratum/hal/bin/dummy/main.cc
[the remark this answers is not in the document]
maksym-tropets-plvision: Changed. Please review.

stratum/hal/lib/bcm/bcm_switch.cc
maksym-tropets-plvision: Changed. Please review.

PR #74, opened 2019-12-17
Replace Python gNMI cli with C++

pudelkoM:
- less dependency (static linked into a single binary)
 - better performance


stratum/hal/bin/dummy/BUILD
```diff
@@ -22,6 +22,7 @@ package(
 )
 
 load("//bazel:rules.bzl", "stratum_cc_binary", "HOST_ARCHES")
+load("@grpc_py_deps//:requirements.bzl", grpc_all_requirements = "all_requirements", grpc_requirement = "requirement")
```
pudelkoM: If we load multiple things, it's better to split them over multiple lines

SeanCondon:
Hi Max - I saw the title of this and wanted to let you know that we made a change to the `gnmi_cli` utility (in Go) coming from openconfig so that it would work with Stratum - it just required the option of turning off TLS
https://github.com/opennetworkinglab/gnmi/tree/master/cmd/gnmi_cli

It might be an alternative to having yet another "gnmi_cli"

pudelkoM:
That's great! Thanks for letting me know, Sean.

We might be able to drop the cli code completely then.

Yi-Tseng:
> Hi Max - I saw the title of this and wanted to let you know that we made a change to the `gnmi_cli` utility (in Go) coming from openconfig so that it would work with Stratum - it just required the option of turning off TLS
> https://github.com/opennetworkinglab/gnmi/tree/master/cmd/gnmi_cli
> 
> It might be an alternative to having yet another "gnmi_cli"

Hi @SeanCondon  Thanks for the update, there are few reasons I created this tool (Python version and C++ version) 
Is there any document for `gnmi_cli` tool?
Can we use gNMI path (e.g. /interfaces/interface[name=eth0]/state) instead of providing entire protobuf message to the tool ?

SeanCondon:
Hi Yi. I haven't tried the query option. I'll check if Adib has any experience of it. The reason we chose it is because it comes straight from the gNMI authors and we thought that's what customers would be likely to use.

We have some documentation for it on our own documentation site:
https://docs.onosproject.org/onos-config/docs/gnmi/

and more at:
https://github.com/onosproject/simulators/blob/master/docs/gnmi/gnmi_user_manual.md

PR #75, opened 2019-12-17
Add P4Runtime "Capabilities" RPC

Yi-Tseng:
c.c. @ccascone  (since I cannot assign him as reviewer)

### DIFF
--- a/stratum/lib/macros.h
+++ b/stratum/lib/macros.h
@@ -121,6 +121,14 @@ _error_impl
 // Replaced by glog CHECK_NOTNULL until it is added to Abseil
 #define ABSL_DIE_IF_NULL CHECK_NOTNULL
 
+// Stringify the result of expansion of a macro to a string
+// e.g:
+// #define A text
+// STRINGIFY(A) => "text"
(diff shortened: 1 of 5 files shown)


stratum/lib/macros.h
```diff
@@ -121,6 +121,14 @@ _error_impl
 // Replaced by glog CHECK_NOTNULL until it is added to Abseil
 #define ABSL_DIE_IF_NULL CHECK_NOTNULL
 
+// Stringify the result of expansion of a macro to a string
+// e.g:
+// #define A text
+// STRINGIFY(A) => "text"
+// Ref: https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html
+#define STRINGIFY_INNER(s) #s
```
pudelkoM: If we add a generally available macro here, https://github.com/stratum/stratum/blob/96e219342a38a9ec13c6f29fa55a58687307f4b6/stratum/hal/config/config_validator_main.cc#L29 can be removed.

Yi-Tseng: Will be fixed in #76

pudelkoM: Ok

PR #78, opened 2019-12-21
Fix onlp tests

### LINKED ISSUE #68: Make ONLP tests testable on CI
Currently they require root rights because of some issue in the ONLP libraries.

pudelkoM:
Fixes #68


stratum/hal/lib/phal/onlp/onlpphal.cc
```diff
@@ -118,11 +118,18 @@ ::util::Status OnlpPhal::VerifyChassisConfig(const ChassisConfig& config) {
...
 ::util::Status OnlpPhal::Shutdown() {
-  absl::WriterMutexLock l(&config_lock_);
-
   // TODO(unknown): add clean up code
 
-  initialized_ = false;
+  {
+    absl::WriterMutexLock l(&init_lock_);
+    if (this == singleton_) {
+      singleton_ = nullptr;
```
pudelkoM: This does not free the instance. On the other hand, I'm not sure if it's ever save to free the singleton, because it could still be in use somewhere. Other classes also don't free on `Shutdown`, so I'd skip it here too..

stratum/hal/bin/bcm/standalone/main.cc
```diff
@@ -92,7 +93,8 @@ int Main(int argc, char** argv) {
   // Create chassis-wide and per-node class instances.
   auto* bcm_diag_shell = BcmDiagShell::CreateSingleton();
   auto* bcm_sdk_wrapper = BcmSdkWrapper::CreateSingleton(bcm_diag_shell);
-  auto* onlpphal = stratum::hal::phal::onlp::OnlpPhal::CreateSingleton();
+  auto onlp_wrapper = phal::onlp::OnlpWrapper::Make().ConsumeValueOrDie();
```
pudelkoM: `auto*` for consistency

stratum/hal/bin/barefoot/main.cc
```diff
@@ -152,11 +153,12 @@ Main(int argc, char* argv[]) {
   std::unique_ptr<DeviceMgr> device_mgr(new DeviceMgr(unit));
 
   auto pi_node = pi::PINode::CreateInstance(device_mgr.get(), unit);
+  auto onlp_wrapper = OnlpWrapper::Make().ConsumeValueOrDie();
```
pudelkoM: I think OnlpWrapper should be a singleton (`CreateInstance`).

`onlp_sw_init` should probably not called multiple times.

pudelkoM:
> Should we remove logging(or use DEBUG level) from unit tests? (If we don't care about the output)

I think that's fine. `bazel test` only shows output when a test fails and when run with `--test_output=errors`. And then more information is good.

PR #80, opened 2020-01-07
Enable C++ style check and fix errors

pudelkoM:
Headers and c++ code should match all rules from cpplint (which based on Google C++ style guide).

Depends on #81 
Minimal subset from #48 
Closes #48

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,7 +167,7 @@ jobs:
       - checkout
       - run:
           name: Run cpplint
-          command: cpplint --linelength=110 --recursive stratum
+          command: cpplint --recursive stratum
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,7 +167,7 @@ jobs:
       - checkout
       - run:
           name: Run cpplint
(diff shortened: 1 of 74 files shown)


.circleci/config.yml
```diff
@@ -160,12 +160,22 @@ jobs:
...
       - *docker_push
 
+  cpp-style-check:
+    docker:
+      - image: stratumproject/build:build
+    steps:
+      - checkout
+      - run:
+          name: Run cpplint
+          command: cpplint --linelength=110 --recursive stratum
```
pudelkoM: No line length

stratum/hal/lib/common/openconfig_converter_test.cc
```diff
@@ -13,16 +13,15 @@
...
-#include <tuple>
-
 #include <google/protobuf/text_format.h>
 
-#include "stratum/hal/lib/common/openconfig_converter.h"
+#include <tuple>
 
+#include "gtest/gtest.h"
 #include "stratum/glue/status/status_test_util.h"
+#include "stratum/hal/lib/common/openconfig_converter.h"
```
pudelkoM: Move to top

PR #115, opened 2020-02-06
PhalCli: Better path parsing

pudelkoM:
I erroneously removed some crucial code paths before. This PR adds them and improves other aspects of the PhalCli.


bocon13:
Can we provide some brief user documentation. Update readme or help string..

stratum/hal/lib/phal/phal_cli.cc
```diff
@@ -46,11 +53,26 @@ ::util::StatusOr<PathQuery> ParseQuery(const std::string& query) {
...
-  for (const auto& field : query_fields) {
-    CHECK_RETURN_IF_FALSE(field != "")
+  for (const auto& query_field : query_fields) {
+    CHECK_RETURN_IF_FALSE(query_field != "")
         << "Encountered unexpected empty query field.";
-    PathQuery::PathEntry entry;
-    entry.set_name(field);
+
+    PathQuery::PathEntry entry;  // Protobuf type
+    RE2 field_regex(R"#((\w+)(\[(?:\d+|\@)\])?)#");
```
bocon13, line 61: I haven't seen the \@ before... does this match [@]? 

In line 72, it seems like if no number is passed in between the brackets, we get all entries. Seems like there are different queries if someone passes /cards vs. /cards[@] (the first in indexed with all, the other is not indexed). Is there a difference?

pudelkoM: Good question. This code is mostly copy&paste from Google's original legacy phal cli, which became onlp_cli: https://github.com/stratum/stratum/blob/a3ade5cf2f568ab57986c8981c8e4c77356f8b30%5E/stratum/hal/lib/phal/phal_cli.cc#L47-L65

I'm also not happy with the really bare bones repl (No history, completion or cursor movement). I'd rather use gFlags and let the shell handle it, like Yi's gNMI cli.

bocon13: Fair enough. I think this is probably fine for now. This is mostly for low level debugging anyway.

Thanks for the pointer. Looks like [@] means all indicies.

pudelkoM: Saw that and wanted to comment, but fixed to code instead.

PR #116, opened 2020-02-10
Check bash scripts

pudelkoM:
[Shellcheck](https://www.shellcheck.net/) checks for common errors and mistakes in bash scripts. Ours are not exactly security critical or handle untrusted input, but we could still make sure they follow best practices.

@ccascone you're a fan of having scripts for common tasks, any opinion on this?


bocon13:
Replace this with superlinter

PR #127, opened 2020-02-19
Use slot and port number for xceiver paths

Yi-Tseng:
Currently, we use node_id and port_id to query the transceiver information.
However, we should use the physical slot and port number to get the actual information

We may or may not define a new type for data request and set request in common.proto

```proto
message DataRequest {
  message Request {
    message PhyPort { // or something else...
      int32 slot = 1;
      int32 port = 2;
    }
    // Defines the data required to get a data for a port.
    message Port {
      uint64 node_id = 1;
      uint32 port_id = 2;
    }
...
```
or add slot/port umber to `Port` message

```proto
message DataRequest {
  message Request {
    // Defines the data required to get a data for a port.
    message Port {
      uint64 node_id = 1;
      uint32 port_id = 2;
      int32 slot = 3;
      int32 port = 4;
    }
```

Or we just reuse the node_id/port_id fields (may confuse developers in the future)


Yi-Tseng:
temporary close this issue since this is not a high priority task

PR #118, opened 2020-02-13
Add optical-related paths

Yi-Tseng:
This is the first part from #72 

@maksym-tropets-plvision we will break PR #72 to four smaller PR so we can merge it faster.

The first one is to add new paths and internal message structure(common.proto) to the Stratum to support optical device.
The second part will be TAI PHAL backend, TaiSwitchConfigurator, TaiDataSource, and OpticalAdapter.
<del>The third part will be TaiSwitchConfigurator, TaiDataSource, and OpticalAdapter</del>
The final part will be TaiManager/Wrapper

I put you and Bohdan as the author of the commit.
Let me know if there is any issue

Co-authored-by: Bohdan Oheruk <bohdan.oheruk@plvision.eu>
Co-authored-by: Yi Tseng <yi@opennetworking.org>

PR #135, opened 2020-02-20
Clean up yang parse tree after #118

pudelkoM:
Found a much better and cleaner solution.

### DIFF
--- a/stratum/hal/lib/common/yang_parse_tree_paths.cc
+++ b/stratum/hal/lib/common/yang_parse_tree_paths.cc
@@ -35,6 +35,14 @@ namespace hal {
 
 namespace {
 
+// A helper method that converts a float to gNMI Decimal64.
+// For use in as a 'process_func', hence no Status return type.
+::gnmi::Decimal64 ConvertDoubleToDecimal64OrDie(const float& value) {
+  auto status = FloatToDecimal64Value(value, kDefaultPrecision);
+  CHECK(status.ok());
+  return status.ConsumeValueOrDie();
+}
+
 // A helper method that prepares the gNMI message.
 ::gnmi::SubscribeResponse GetResponse(const ::gnmi::Path& path) {
   ::gnmi::Notification notification;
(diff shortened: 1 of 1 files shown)


stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -371,43 +379,48 @@ U GetValue(
...
-
-  // Retrieve value from the switch.
+  DataResponseWriter writer(
+      [&resp, data_response_get_inner_message_func,
+       data_response_has_inner_message_func,
+       optical_channel_info_has_inner_message_func,
+       optical_channel_info_get_inner_message_func,
+       inner_message_get_field_func](const DataResponse& in) {
+        if (!(in.*data_response_has_inner_message_func)()) return false;
+        auto oc_info = (in.*data_response_get_inner_message_func)();
```
Yi-Tseng: `oc_info` ->`data_req_inner_msg` ?

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -371,43 +379,48 @@ U GetValue(
 }
 
 // Optical-port-specific version.
```
Yi-Tseng: This becomes a general function. I think we can rename some variable

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -371,43 +379,48 @@ U GetValue(
...
+U GetValue(
     uint64 node_id, uint32 port_id, YangParseTree* tree,
-    const T& (OpticalChannelInfo::*instance_getter)() const,
-    bool (OpticalChannelInfo::*has_instance_getter)() const,
-    U (T::*inner_message_get_field_func)() const) {
+    const T& (DataResponse::*data_response_get_inner_message_func)() const,
+    bool (DataResponse::*data_response_has_inner_message_func)() const,
+    DataRequest::Request::Port* (
+        DataRequest::Request::*data_request_get_mutable_inner_message_func)(),
+    bool (T::*optical_channel_info_has_inner_message_func)() const,
```
Yi-Tseng: `data_req_has_inner_message_func` ?

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -371,43 +379,48 @@ U GetValue(
...
+    const V& (T::*optical_channel_info_get_inner_message_func)() const,
```
Yi-Tseng: `data_req_get_inner_message_func`?

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -371,43 +379,48 @@ U GetValue(
...
+      [&resp, data_response_get_inner_message_func,
+       data_response_has_inner_message_func,
+       optical_channel_info_has_inner_message_func,
+       optical_channel_info_get_inner_message_func,
+       inner_message_get_field_func](const DataResponse& in) {
+        if (!(in.*data_response_has_inner_message_func)()) return false;
+        auto oc_info = (in.*data_response_get_inner_message_func)();
+        if (!(oc_info.*optical_channel_info_has_inner_message_func)())
+          return false;
+        auto power = (oc_info.*optical_channel_info_get_inner_message_func)();
```
Yi-Tseng: `power` -> `data_req_inner_msg_field` ? (or something else

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -525,6 +538,30 @@ TreeNodeEventHandler GetOnPollFunctor(
   };
 }
 
+// Optical-port-specific version.
+// Optical port messages are two level nested messages (DataResponse::T::U).
```
Yi-Tseng: Same as here and below

stratum/hal/lib/common/yang_parse_tree_paths.cc
pudelkoM: It's not for the request, but for the response inner message.

Yi-Tseng: `inner_msg_has_inner_msg_func`? (or maybe someting else?

stratum/hal/lib/common/yang_parse_tree_paths.cc
pudelkoM: Done. I left the comment about it being optical port specific to match the others.

PR #143, opened 2020-02-25
Fix dependency for CDLang

Yi-Tseng:
Use `repo1.maven.org` instead of `central.maven.org`


Yi-Tseng:
> Do we actually test/need this anywhere?

QA team is now working on CDLang

PR #151, opened 2020-03-04
Return ENTRY_EXISTS error for group inserts

pudelkoM:
We returned the wrong error code, which prompted retries by ONOS instead of deletes.


Yi-Tseng:
Should we also return ENTRY_EXISTS at line 1138?

pudelkoM:
> Should we also return ENTRY_EXISTS at line 1138?

Yes, I removed the check. 
a) we already check beforehand: https://github.com/stratum/stratum/blob/455ede1d96f52976cbdeddc976b45968cb3fa7d2/stratum/hal/lib/bcm/bcm_node.cc#L855-L858

b) The `InsertIfNotPresent` will catch it.

PR #154, opened 2020-03-04
Split Barefoot Dockerfile to speed up image building

yoooou:


stratum/hal/bin/barefoot/docker/Dockerfile.builder
```diff
@@ -0,0 +1,68 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    python \
+    python-yaml \
+    python-pip \
+    libssl-dev \
+    libelf-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+# Copy stratum source code, include SDE and Linux headers tarball
+ADD . /stratum
```
Yi-Tseng: we don't need to copy everything now since we are not building the stratum in this container

yoooou: I'll try if copying only two tarballs works

yoooou: Several updates:
- SDE and header tarballs now can be placed anywhere on the build node and the script takes paths to the tarballs as arguments. When building SDE and kernel modules, the tarballs are copied to the directory where the Dockerfile is located and docker build also starts form there. Internally docker copies ONLY the tarballs to the builder container. The copied tarballs are then removed after the build.
- When building the runtime image, ONLY stratum code is copied to the builder container.
- "-i" options (to keep intermediate builder image) is removed so now it always saves the builder image locally. This is added to the README as a note. Let me know if you think it's better to not save the builder image by default.
- "-b" option (to specify a builder image and skip SDE/kernel modules building) is also removed for simplicity since when the tarballs stay the same and the corresponding builder image exists, building SDE/kernel modules only takes a second. So we don't really need to skip the build. Let me know if this makes sense.

PR #155, opened 2020-03-04
Implement SDKLT config file (.yaml) generator

pudelkoM:
I tested this with the AS7712 inside the Trellis topo. 

Blocked by #157


pudelkoM:
We'll have to wait with this until there is a way to enable loopback mode on ports.

stratum/hal/lib/bcm/bcm_chassis_manager.cc
```diff
@@ -1635,12 +1636,168 @@ bool BcmChassisManager::IsSingletonPortMatchesBcmPort(
...
+    // Key is a map (PC_PM_ID: serdes_core)
+    pc_pm << YAML::Key << YAML::BeginMap << YAML::Key << "PC_PM_ID"
+          << YAML::Value << bcm_port.serdes_core() << YAML::EndMap;
+
+    pc_pm << YAML::Value << YAML::BeginMap;
+
+    pc_pm << YAML::Key << "PM_OPMODE" << YAML::Value << YAML::Flow
+          << YAML::BeginSeq << "PC_PM_OPMODE_DEFAULT" << YAML::EndSeq;
+
+    pc_pm << YAML::Key << "SPEED_MAX" << YAML::Value << YAML::Flow
```
pudelkoM: `SPEED_MAX` in the `PC_PM` table should not be set to the requested speed, as it prevents changing from 40G to 100G at runtime.

stratum/hal/lib/bcm/bcm_chassis_manager.cc
```diff
@@ -1729,12 +1730,175 @@ bool BcmChassisManager::IsSingletonPortMatchesBcmPort(
...
+  pc_pm << YAML::BeginMap;
+  pc_pm << YAML::Key << "device";
+  pc_pm << YAML::Value << YAML::BeginMap;
+  for (size_t i = 0; i < max_num_units; ++i) {
+    pc_pm << YAML::Key << i;  // <unit>
+    pc_pm << YAML::Value << YAML::BeginMap;
+    pc_pm << YAML::Key << "PC_PM";
+    pc_pm << YAML::Value << YAML::BeginMap;
+    for (const auto& bcm_port : target_bcm_chassis_map.bcm_ports()) {
+      // Key is a map (PC_PM_ID: serdes_core)
```
Yi-Tseng, line 1748: I think we should also check `unit` from bcm_port, for example:
```cpp
if (bcm_port.unit() != i) {
  continue;
}
```

stratum/hal/lib/bcm/bcm_chassis_manager.cc
```diff
@@ -1729,12 +1730,175 @@ bool BcmChassisManager::IsSingletonPortMatchesBcmPort(
...
-  return ::util::OkStatus();
+  const size_t max_num_units = base_bcm_chassis_map.bcm_chips_size();
+
+  // PC_PM Table
+  YAML::Emitter pc_pm;
+  pc_pm << YAML::BeginDoc;
+  pc_pm << YAML::BeginMap;
+  pc_pm << YAML::Key << "device";
+  pc_pm << YAML::Value << YAML::BeginMap;
+  for (size_t i = 0; i < max_num_units; ++i) {
```
Yi-Tseng, line 1742: would be clear to use `unit` instead of `i`

PR #156, opened 2020-03-05
Fix bazel gRPC rules

Yi-Tseng:
Bazel gRPC rules requires python-dev and python3-dev installed


pudelkoM:
Really version 2 AND 3 at the same time?

PR #159, opened 2020-03-06
Use "-O0" for CI jobs

Yi-Tseng:


pudelkoM:
Reducing the optimization level in CI should reduce build and test times by about 60% and make them more reliable. This does not affect any binaries you build locally.

I did a few test run on my MacBook:

```
Bazel 2.2.0
8 GB RAM, 6 Cores
bazel test --config=asan --test_output=errors --define bcm_sdk=sdk --define phal_with_tai=false //stratum/hal/lib/bcm/...

-O0 -> INFO: Elapsed time: 189.020s, Critical Path: 34.07s
-O1 -> INFO: Elapsed time: 333.873s, Critical Path: 89.86s
-O2 -> INFO: Elapsed time: 518.296s, Critical Path: 182.04s
-O3 -> OoM killed
```

I did not monitor memory usage, but suspect that's one of the reasons our CI jobs get killed.

PR #172, opened 2020-03-30
Implement ONLPv2 platform test

### LINKED ISSUE #170: Create ONLPv2 platform test to check conformity
We should create a test that goes through all the ONLPv2 API calls we care about and makes sure they produce sensible values. This test would not be run as part of CI, but on the switch-under-test directly.

Bazel allows making a test [manual only](https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes) with the test attribute `manual`. `:all` or `/...` will not trigger it.

While unit tests should only test project internal code, this would help us finding errors in the platform implementation and could be part of the CCP.

pudelkoM:
This PR creates a ONLPv2 platform test to check conformity. 
Due to it's platform dependent manner, it has to be run manually and explicitly like so:

`bazel test --config=asan --test_output=errors //stratum/hal/lib/phal/onlp:onlp_platform_test`

Fixes #170

### DIFF
--- a/bazel/rules/test_rule.bzl
+++ b/bazel/rules/test_rule.bzl
@@ -16,7 +16,7 @@
 
 def stratum_cc_test(name, deps = None, srcs = None, data = None, copts = None,
                     defines = None, linkopts = None, size = "small",
-                    visibility = None):
+                    visibility = None, tags = None, linkstatic = None):
   native.cc_test(
     name = name,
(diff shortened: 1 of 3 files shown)


pudelkoM:
The recent changes to the ONLP wrapper make using the direct APIs more difficult.

Closing until we have a need for this.

PR #178, opened 2020-04-04
Support pushing chassis config via gNMI root path

Yi-Tseng:


stratum/hal/lib/common/yang_parse_tree.cc
```diff
@@ -146,6 +146,18 @@ void YangParseTree::SendNotification(const GnmiEventPtr& event) {
...
+
+  // Add the minimum nodes:
+  //   /interfaces/interface[name=*]/state/ifindex
+  //   /interfaces/interface[name=*]/state/name
+  //   /interfaces/interface/...
+  //   /
+  // The rest of nodes will be added once the config is pushed.
+  AddSubtreeAllInterfaces();
+  AddSubtreeAllComponents();
+  AddRoot();
```
pudelkoM, line 160: Also called later https://github.com/stratum/stratum/pull/178/files#diff-c4a583ef9a466f4a5199f2fe94500d9fR229
Do we need both?

Yi-Tseng: good catch, I think we can put them together, I'll remove the line 229

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -782,12 +783,16 @@ void SetUpRoot(TreeNode* node, YangParseTree* tree) {
...
-
+    // Deserialize the input proto to OpenConfig device format.
+    if (in.ParseFromString(typed_value->bytes_val())) {
+      // Convert the input proto into the internal format.
+      ASSIGN_OR_RETURN(*config->writable(),
+          OpenconfigConverter::OcDeviceToChassisConfig(in));
+    } else {
+      // Try parse it with ChassisConfig format.
+      RETURN_IF_ERROR(
+        ParseProtoFromString(typed_value->bytes_val(), config->writable()));
```
Yi-Tseng, line 794: indent

pudelkoM:
@Yi-Tseng 
Can you give an example how to push a ChassisConfig with the gnmi cli?

stratum/hal/lib/common/yang_parse_tree.cc
pudelkoM, line 160: Could this be moved to the end of function? Where `AddRoot()` was before.

Issue #187, opened 2020-04-09
P4 Write batch can hang


bocon13:
If you send a batch of writes to the P4Runtime Service, gRPC packs errors into the gRPC status response which goes into the HTTP2 header. If this response is sufficiently large (more than one TCP frame), then it will get split up into multiple HTTP2 continuation chunks.

The gRPC C++ server does not mark the original header correctly (with END_STREAM) if the header gets split into chunks, and as a result clients will hang indefinitely.

This problem can largely be avoid by using small write batches.

gRPC Issue:
https://github.com/grpc/grpc/issues/21436

pudelkoM:
Fixed by #191

PR #191, opened 2020-04-10
Updating gRPC dependency

bocon13:
Moving to grpc v1.28.1, plus a bug fix for #187


bocon13:
Added the reminder: https://github.com/stratum/stratum/issues/192

PR #195, opened 2020-04-11
Updating Mininet Dockerfile gRPC version to match Stratum's

bocon13:


pudelkoM:
Is this change still necessary? Current master builds in CI.

bocon13:
It shouldn't make a big difference. The version here now matches Stratum server version, so I'm inclined to take the upgrade.

PR #198, opened 2020-04-13
Build a Debian package for p4c-fpm

pudelkoM:
- Update rules_pkg dependency
- Fix rules_pkg load paths
- Create a deb target: p4c_fpm_dep
- Simplify docker image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,12 +160,14 @@ jobs:
       - *docker_push
 
   publish-docker-p4c-fpm:
-    machine: true
+    docker:
+      - image: stratumproject/build:build
     environment:
       - DOCKER_SCOPE: stratum/p4c_backends/fpm
       - DOCKER_FILE: Dockerfile
       - DOCKER_IMG: stratumproject/p4c-fpm
     steps:
+      - setup_remote_docker
       - checkout
       - *restore_bazel_cache
       - *set_bazelrc
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,19 +162,26 @@ jobs:
   publish-docker-p4c-fpm:
     machine: true
     environment:
-      - DOCKER_SCOPE: .
-      - DOCKER_FILE: stratum/p4c_backends/fpm/Dockerfile
+      - DOCKER_SCOPE: stratum/p4c_backends/fpm
+      - DOCKER_FILE: Dockerfile
       - DOCKER_IMG: stratumproject/p4c-fpm
     steps:
       - checkout
-      - *docker_login
+      - *restore_bazel_cache
       - *set_bazelrc
+      - run:
+          name: Build p4c-fpm
+          command: |
+            bazel build //stratum/p4c_backends/fpm:p4c_fpm_dep
+            cp bazel-bin/stratum/p4c_backends/fpm/p4c_fpm_dep.deb $DOCKER_SCOPE
+      - *docker_login
       - *docker_build
       - run:
           name: Test p4c-fpm Docker image
           command: |
             docker run $DOCKER_IMG -version
       - *docker_push
+      - *save_bazel_cache
 
   publish-docker-stratum-bcm:
     docker:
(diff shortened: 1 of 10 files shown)


stratum/hal/bin/bmv2/BUILD
```diff
@@ -46,7 +46,7 @@ stratum_cc_binary(
     data = ["dummy.json"],
 )
 
-load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_deb", "pkg_tar")
+load("@rules_pkg//:pkg.bzl:pkg.bzl", "pkg_deb", "pkg_tar")
```
pudelkoM, line 49: Fix

stratum/p4c_backends/fpm/Dockerfile
```diff
@@ -16,34 +16,19 @@
 
 # This Dockerfile expects the stratum root as its scope, hence you should build
 # from root e.g.:
+# TODO(max)
 # docker build -t <some tag> -f stratum/p4c_backends/fpm/Dockerfile .
```
pudelkoM, line 20: Fix docker command

stratum/p4c_backends/fpm/BUILD
```diff
@@ -1916,6 +1918,43 @@ cc_binary(
...
+pkg_tar(
+    name = "fpm_p4c_data",
+    deps = [
+        ":fpm_p4c_bin",
+        ":fpm_p4c_includes",
+    ],
+)
+
+pkg_deb(
+    name = "p4c_fpm_dep",
```
Yi-Tseng, line 1949: `p4c_fpm_deb`?

stratum/p4c_backends/fpm/Dockerfile
```diff
@@ -14,36 +14,23 @@
...
+FROM bitnami/minideb:stretch
 
 LABEL maintainer="Stratum dev <stratum-dev@lists.stratumproject.org>"
 LABEL description="Docker-based distribution of the Stratum p4c FPM backend"
 
 ENV RUNTIME_DEPS libgmp10 libgmpxx4ldbl gcc
 RUN install_packages ${RUNTIME_DEPS}
 
-COPY --from=builder /output /
+ADD ./p4c_fpm_deb.deb /
```
ccascone, line 32: The resulting docker image will have the deb package taking up space unnecessarily. Before, only the p4c-fpm binary and dependencies were copied to the final image. You might have simplified the Dockerfile, but image efficiency seems to have dropped. I don't know how big is the debian package, so I'm not sure whether such inefficiency can be ignored. You could use the same approach as before, i.e., install the deb package in the first stage and copy over the binary in the second stage.

pudelkoM: Done. Should I add the comments back in too?

.circleci/config.yml
```diff
@@ -226,9 +235,9 @@ workflows:
       - publish-docker-mininet:
           requires:
             - publish-docker-build
-      - publish-docker-p4c-fpm:
-          requires:
-            - publish-docker-build
+      - publish-docker-p4c-fpm
+          # requires:
+            # - publish-docker-build
```
pudelkoM, line 240: Since the build times are much lower now, we could think about running this job on all branches, but only push on master. Maybe change the `docker_push` part to include a check?

ccascone: It sounds like a good idea to me. You could else try to build a P4 program (e.g., main.p4) as part of the docker image check

stratum/p4c_backends/fpm/Dockerfile
ccascone: Up to you. I think it should be pretty straightforward for anyone with knowledge of docker multi-stage builds.

.circleci/config.yml
pudelkoM: I tied to get a short command, but it always resulted in this long string with lots of Bazel internal path in it.
We already compile main.p4 as part of CI (https://github.com/stratum/stratum/blob/master/stratum/pipelines/main/BUILD#L33-L41) so I'd leave that out for now.

PR #201, opened 2020-04-16
Dev/netcope

craigsdell:
This is the NP4 Intel FPGA SmartNIC integration PR.


stratum/hal/bin/np4intel/np4intel.bzl
```diff
@@ -0,0 +1,37 @@
...
+  hdrs = [], # see cc_library rule above 
+  shared_library = "netcope-bin/lib/libnp4atom.so",
+)
+
+""")
+
+np4intel_configure = repository_rule(
+    implementation=_impl,
+    local = True,
+    environ = ["NP4_INSTALL"])
```
pudelkoM, line 37: Convert to proper dependency, if possible.
Example for external SDKs: https://github.com/stratum/stratum/blob/07d6e325e3da6694bf3eb04e116d61630555d42e/bazel/deps.bzl#L272-L288

pudelkoM:
Can you fix the formatting of call C++ and bazel files?

clang-format and buildifier should take care of this.

stratum/hal/bin/np4intel/np4intel.bzl
pudelkoM: To include local files, `urls` supports the `file://` scheme:

```
        http_archive(
            name = "some_dpe",
            sha256 = "5aaeac84cbc6f1869c2ad99a973e107c67e0f99bb8c5c4e02282f205c2b3acf9",
            urls = ["file:///tmp/file.tar"],
            build_file = "@//bazel:external/some_dep.BUILD",
        )
```

craigsdell: Hi Max, I think I mentioned this in a previous TST meeting, I can't include the netcope tarball in the bazel build dependency as the tarball is actually installed as part of the docker build process by running a shell script supplied by Netcope in the tarball directory (see here https://github.com/craigsdell/stratum/blob/dev/netcope/stratum/hal/bin/np4intel/docker/Dockerfile.builder).  I guess we could put a checksum test in the docker build file, but right now Netcope don't have a standard tarball that they supply so we might need to approach them about that so that it's a consistent tarball supplied to all customers/users of stratum and we could include a checksum test in the docker build.

pudelkoM: The tarball was just an example, if you're current using an unpacked directory, [new_local_repository](https://docs.bazel.build/versions/master/be/workspace.html#new_local_repository) would be the right choice.

stratum/hal/bin/np4intel/main.cc
```diff
@@ -0,0 +1,190 @@
...
+
+        // Now log the return code message
+        if (rc != 0) {
+            RETURN_ERROR(ERR_INTERNAL) << "DPDK EAL Init failed";
+        }
+        LOG(INFO) << "DPDK EAL Init successful";
+    }
+
+    return ::util::OkStatus();
+}
```
pudelkoM, line 124: Here is an example of 4 spaces.

WORKSPACE
```diff
@@ -35,6 +35,11 @@ workspace(name = "com_github_stratum_stratum")
 #       Please do not push changes to this section upstream.
 # ---------------------------------------------------------------------------
 
+#local_repository(
+#    name = "com_github_p4lang_PI_np4",
+#    path = "/home/craigsid/np4_install/PI",
+#)
+
```
pudelkoM, line 42: remove?

stratum/hal/lib/np4intel/np4_chassis_manager.cc
```diff
@@ -0,0 +1,427 @@
...
+  if (initialized_) {
+    return MAKE_ERROR(ERR_INTERNAL)
+           << "RegisterEventWriters() can be called only before the class is "
+           << "initialized.";
+  }
+
+  port_status_change_event_channel_ = Channel<PortStatusChangeEvent>::Create(
+      kMaxPortStatusChangeEventDepth);
+
+  port_status_change_event_writer_ =
```
pudelkoM, line 372: warning: writing variable 'port_status_change_event_writer_' requires holding mutex 'port_status_change_event_writer_lock_' exclusively [-Wthread-safety-analysis]
  port_status_change_event_writer_ =
  ^

stratum/hal/lib/np4intel/np4_chassis_manager.cc
```diff
@@ -0,0 +1,427 @@
...
+::util::Status NP4ChassisManager::Shutdown() {
+  ::util::Status status = ::util::OkStatus();
+  {
+    absl::ReaderMutexLock l(&chassis_lock);
+    if (!initialized_) return status;
+  }
+  // It is fine to release the chassis lock here (it is actually needed to call
+  // UnregisterEventWriters or there would be a deadlock). Because initialized_
+  // is set to true, RegisterEventWriters cannot be called.
+  APPEND_STATUS_IF_ERROR(status, UnregisterEventWriters());
```
pudelkoM, line 416:
```
warning: calling function 'UnregisterEventWriters' requires holding mutex 'chassis_lock' exclusively [-Wthread-safety-analysis]
  APPEND_STATUS_IF_ERROR(status, UnregisterEventWriters());
                                 ^
```

stratum/hal/bin/np4intel/BUILD
```diff
@@ -0,0 +1,108 @@
...
+        ":stratum_np4intel_config",
+        ":stratum_np4intel_pipeline",
+        ":stratum_np4intel_service",
+    ],
+)
+
+pkg_deb(
+    name = "stratum_np4intel_debian",
+    architecture = "amd64",
+    built_using = "bazel (0.14.1)",
```
pudelkoM, line 81: Remove

stratum/hal/bin/np4intel/stratum_np4intel.service
```diff
@@ -0,0 +1,23 @@
...
+
+ExecStart=/usr/bin/stratum_np4intel \
+    --external_hercules_urls=${EXTERNAL_HERCULES_URL} \
+    --forwarding_pipeline_configs_file=${FORWARDING_PIPELINE_CONFIGS_FILE} \
+    --persistent_config_dir=${PERSISTENT_CONFIG_DIR} ${EXTRA_NP_CONFIG}
+
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
```
pudelkoM, line 23: Is this service actually used anywhere?

stratum/hal/bin/np4intel/update_config.py
```diff
@@ -0,0 +1,157 @@
...
+if __name__ == '__main__':
+    test = Test()
+    test.set_up_stream()
+    test.update_config()
+    if args.loopback:
+        packet_out = p4runtime_pb2.PacketOut()
+        packet_out.payload = "\xab" * 100
+        test.send_packet_out(packet_out)
+        test.get_packet_in()
+    test.tear_down_stream()
```
pudelkoM, line 157: Is this file specific to NP4? If not, tools like this are better stored out side of Stratum.

WORKSPACE
craigsdell: yeah maybe, I guess I left it as an example of how to include dev versions of the PI code.  Perhaps I could change my userid to say <userid> ? and add a comment to this effect?

stratum/hal/bin/np4intel/stratum_np4intel.service
craigsdell: yes this is not used anymore in the docker world, I will remove.

stratum/hal/bin/np4intel/docker/Dockerfile.builder
```diff
@@ -0,0 +1,132 @@
...
+    rm -f bazel-$BAZEL_VERSION-installer-linux-x86_64.sh
+
+# Install java and lcov for Bazel coverage
+ARG JDK_URL
+ENV JAVA_HOME /usr/local/lib/jvm
+RUN wget $JDK_URL -O $HOME/jdk_11.0.2.tar.gz && \
+    mkdir -p $JAVA_HOME && \
+    tar xf $HOME/jdk_11.0.2.tar.gz -C $JAVA_HOME --strip-components=1 && \
+    rm $HOME/jdk_11.0.2.tar.gz
+ENV PATH=$PATH:/usr/local/lib/jvm/bin
```
pudelkoM, line 76: Replace with `FROM stratumproject/build:build as builder` ?

setup_np4intel_dev_env.sh
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env bash
```
pudelkoM, line 1: I think this could be merged into `setup_dev_env.sh` with a few variables. No need to have 90% duplication.

setup_np4intel_dev_env.sh
```diff
@@ -0,0 +1,135 @@
...
+#
+
+if [[ $EUID -eq 0 ]]; then
+   echo "This script should not be run as root, run it as the user who owns the Stratum source directory"
+   exit 1
+fi
+
+PULL_DOCKER=NO
+MOUNT_SSH=NO
+BAZEL_CACHE=$HOME/.np4intel_cache
```
pudelkoM, line 25: Did you encounter errors when using the normal cache? Bazel should handle this.

stratum/hal/bin/np4intel/docker/Dockerfile.builder
craigsdell: No, the container is based on ubuntu because Netcope SDK requires either Ubuntu or CentOS for the install scripts to work etc.  Could see if we can get them to support Debian over time given it's pretty close, but for now, that is what we have.

setup_np4intel_dev_env.sh
craigsdell: I'd have to go back and check, but I remember having issues when building the normal stratum based on debian and the np4intel platform based on ubuntu.  It ends up sharing the cache between the two different build environments.

setup_np4intel_dev_env.sh
craigsdell: probably true, but I figured given that they both build on different docker container bases (i.e. stratum uses debian, stratum for np4intel uses Ubuntu) then it might make sense to have two different scripts, but I guess we could have an option instead.  You'll note there is also a different builder container that needs to be created... again because it's based on Ubuntu and also installs the NP4 SDK libraries.

stratum/hal/bin/np4intel/update_config.py
craigsdell: you're right, I've removed it, along with a couple of others that weren't needed

stratum/hal/bin/np4intel/BUILD
craigsdell: yes will remove this plus the pkg_tar files as we only really build and deploy as a container.

pudelkoM: oh, this was just about the `build_using` line. The package is fine, if you want to keep it.

stratum/hal/bin/np4intel/docker/Dockerfile.builder
pudelkoM: Ahh didn't see the FROM line.

stratum/hal/lib/np4intel/np4_chassis_manager.cc
craigsdell: done.  Note: that bmv2 code has the same issue will need to be fixed

stratum/hal/lib/np4intel/np4_chassis_manager.cc
craigsdell: done

stratum/hal/bin/np4intel/BUILD
craigsdell: done

stratum/hal/bin/np4intel/stratum_np4intel.service
craigsdell: removed

stratum/hal/bin/np4intel/update_config.py
craigsdell: removed

WORKSPACE
craigsdell: removed

stratum/hal/bin/np4intel/main.cc
craigsdell: fixed, I've run clang-format over all the new code.

setup_np4intel_dev_env.sh
craigsdell: I've merged them into the same setup_dev_env.sh script, they use different caches, docker files and image names.

setup_np4intel_dev_env.sh
craigsdell: Just checked it and it does cause problems using the same cache for normal debian builds and np4_intel ubuntu builds. I can get around the issue by doing a "bazel clean -expunge" and then a rebuild but that defeats the purpose of a cache.

PR #206, opened 2020-04-23
Add rules for py proto files and package them

### LINKED ISSUE #205: Missing P4Runtime Python bindings from mn-stratum image
Instead of `.py` files, the image contains `.cc` and `.h` files in the corresponding python package location.

To reproduce the issue:
```
docker pull opennetworking/mn-stratum:latest
docker run --rm -it --entrypoint bash opennetworking/mn-stratum:latest
root@a43a429b0a76:~# ls /usr/local/lib/python2.7/dist-packages/p4/v1/
__init__.py  p4data.pb.cc  p4data.pb.h  p4runtime.grpc.pb.cc  p4runtime.grpc.pb.h  p4runtime.pb.cc  p4runtime.pb.h
```

pudelkoM:
Fixes #205 

We now use the official P4Runtime docker image and files in the Mininet container. This solves the issues with file placement and import paths when build with Bazel.


bazel/external/p4runtime.BUILD
```diff
@@ -101,3 +101,14 @@ py_grpc_library(
     srcs = [":p4runtime_proto"],
     deps = [":p4runtime_py_proto"],
 )
+
+py_proto_library(
+    name = "p4data_py_proto",
+    deps = [":p4data_proto"],
+)
+
+py_grpc_library(
```
ccascone, line 110: This should not be a grpc library. p4data doesn't expose any grpc service. The only grpc service is provided by p4runtime.proto.
```suggestion
py_proto_library(
```

tools/mininet/BUILD
```diff
@@ -0,0 +1,32 @@
...
+
+load("@rules_pkg//:pkg.bzl", "pkg_tar", "pkg_deb")
+
+licenses(["notice"])  # Apache v2
+
+package(
+    default_visibility = ["//visibility:public"],
+)
+
+pkg_tar(
```
bocon13, line 25: we might consider a wheel rather than a tarball here

pudelkoM: The `py_wheel` rules are experimental and not available in the released version yet. I also did not see a way to structure the files like we need them.

pudelkoM:
Current content with the latest image from dockerhub:

```
docker run --rm -it --entrypoint bash opennetworking/mn-stratum:latest
root@9693a49556a2:~# ls -lR /usr/local/lib/python2.7/dist-packages/p4/
/usr/local/lib/python2.7/dist-packages/p4/:
total 8
-rw-r--r-- 1 root root    0 Jan  1  2000 __init__.py
drwxr-xr-x 3 root root 4096 Jan  1  2000 config
drwxr-xr-x 2 root root 4096 Jan  1  2000 v1

/usr/local/lib/python2.7/dist-packages/p4/config:
total 4
-rw-r--r-- 1 root root    0 Jan  1  2000 __init__.py
drwxr-xr-x 2 root root 4096 Jan  1  2000 v1

/usr/local/lib/python2.7/dist-packages/p4/config/v1:
total 144
-rw-r--r-- 1 root root 74689 Jan  1  2000 p4info_pb2.py
-rw-r--r-- 1 root root 67247 Jan  1  2000 p4types_pb2.py

/usr/local/lib/python2.7/dist-packages/p4/v1:
total 184
-rw-r--r-- 1 root root      0 Jan  1  2000 __init__.py
-rw-r--r-- 1 root root  17859 Jan  1  2000 p4data_pb2.py
-rw-r--r-- 1 root root 148315 Jan  1  2000 p4runtime_pb2.py
-rw-r--r-- 1 root root  12531 Jan  1  2000 p4runtime_pb2_grpc.py
```

pudelkoM:
Fixing upstream is not a solution because of this issue https://github.com/grpc/grpc/issues/21830

pudelkoM:
I decided to use the official Docker image to generate the necessary Python code. This is a bit ugly, but considering the overall state of this CI job, still ok.

tools/mininet/Dockerfile
```diff
@@ -53,24 +53,19 @@ RUN cp --parents /usr/local/lib/libbmpi.so.0 /output
...
 # to run mininet, but it's useful to execute Python scripts acting as a
-# P4Runtime client, e.g. control plane apps or PTF tests.
-RUN bazel build @com_github_p4lang_p4runtime//:p4runtime_py_grpc
-
+# P4Runtime client, e.g. control plane apps or PTF tests. Also includes
+# Google API (e.g. status proto) needed by P4Runtime.
 ENV PYTHON_PACKAGE_BASE /output/usr/local/lib/python2.7/dist-packages
-RUN mkdir -p $PYTHON_PACKAGE_BASE/p4 && \
-    cp -r bazel-bin/external/com_github_p4lang_p4runtime/p4/* $PYTHON_PACKAGE_BASE/p4 && \
+COPY ./py_out /p4runtime
```
ccascone, line 59: I would add a comment explaining what `./py_out` is and how it should be provided. Also, since the build of this image now depends on the execution of some other steps first (to generate `py_out`), the instructions at the beginning of this file should be updated. You could simply say "Look at .circleci/config.yaml for an example of commands to build this image."

pudelkoM:
> LGTM, but if you would like to consider a prettier solution, instead of depending on the p4runtime's repo CI script, you could simply run the same protoc command inside the Dockerfile (like it is done here https://github.com/p4lang/p4runtime/blob/master/CI/compile_protos.sh#L63).
> Notice that the p4runtime's `.proto` files, as well as the `protoc` binary, should already be inside the Docker container you are building, downloaded by Bazel to build `stratum_bmv2`.

I'll look into it.

> LGTM, but if you would like to consider a prettier solution, instead of depending on the p4runtime's repo CI script, you could simply run the same protoc command inside the Dockerfile (like it is done here https://github.com/p4lang/p4runtime/blob/master/CI/compile_protos.sh#L63).
> Notice that the p4runtime's `.proto` files, as well as the `protoc` binary, should already be inside the Docker container you are building, downloaded by Bazel to build `stratum_bmv2`.

Good idea, but the `grpc_python_plugin` is missing in our image.

tools/mininet/Dockerfile
```diff
@@ -122,3 +128,4 @@ EXPOSE 50001-50100
 
 ENTRYPOINT ["mn", "--custom", "/root/stratum.py", "--switch", "stratum-bmv2", "--host", "no-offload-host", "--controller", "none"]
 
+RUN python -c "from p4.config.v1 import p4info_pb2;from google.rpc import status_pb2, code_pb2;from p4.v1 import p4runtime_pb2, p4runtime_pb2_grpc"
```
ccascone, line 131: I think it would make more sense to do this test as part of the CI script, otherwise, we create an unnecessary layer in the final image.

ccascone: This command currently fails on my machine, but it should succeed in CircleCI with this PR:
```➜  stratum git:(master) docker run --rm --entrypoint python opennetworking/mn-stratum:latest -c "from p4.config.v1 import p4info_pb2;from google.rpc import status_pb2, code_pb2;from p4.v1 import p4runtime_pb2, p4runtime_pb2_grpc"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: cannot import name code_pb2```

pudelkoM: Sure, I just wanted something to test on my machine.
Moved to CI

PR #212, opened 2020-04-26
 Move to IANA-reserved port 9339 for gNMI/gNOI

### LINKED ISSUE #199: Move to IANA-reserved port 9339 for gNMI/gNOI
We should move to port 9339 for gNMI/gNOI as the default listening port. We should also consider using the same for P4RT.

For existing users, we should deprecate / support 28000 for a little longer.

https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=gnmi

Yi-Tseng:
Fixes #199


stratum/hal/lib/common/hal.cc
```diff
@@ -32,7 +32,7 @@
 #include "absl/synchronization/mutex.h"
 
 // TODO(unknown): Use FLAG_DEFINE for all flags.
-DEFINE_string(external_stratum_urls, "",
+DEFINE_string(external_stratum_urls, stratum::kExternalStratumUrls,
```
pudelkoM, line 35: If we set a default here, we could remove all `--external_stratum_urls`` mentions from flag files, start script, etc.

PR #217, opened 2020-04-28
Add `m` utility to mn-stratum image

Yi-Tseng:
`m` utility allows user to attach to a mininet host or execute a specific
command in a host without using mininet CLI.

### DIFF
--- a/tools/mininet/examples/trellis/README.md
+++ b/tools/mininet/examples/trellis/README.md
@@ -378,6 +378,16 @@ PING 10.0.3.1 (10.0.3.1) 56(84) bytes of data.
 ^C
 ```
 
+#### Run a command on a Mininet host
+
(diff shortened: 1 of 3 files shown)


tools/mininet/examples/trellis/README.md
```diff
@@ -54,6 +54,7 @@ reference:
 | `make mn-log`       | Show the Mininet log (i.e., the CLI output)            |
 | `make netcfg`       | Push netcfg.json file (network config) to ONOS         |
 | `make reset`        | Reset the tutorial environment                         |
+| `make mn-host`      | Run a command in a Mininet host.                       |
```
ccascone, line 57: It would be nice to provide an example of how to use this new command (make target) in the instructions below, for example, to start a ping from one of the two hosts.

PR #220, opened 2020-05-01
Add chassis_config, chassis_map & phal_config for Inventec d7054

habeebmohammed:
Added the configs for d7054.

Inventec D7054 is Broadcom tomahawk_plus based platform 48*25G+ 6*100G.

Verified the configs can work. 
Attached is the log.
[Cypress_stratum_openNSA.txt](https://github.com/stratum/stratum/files/4562097/Cypress_stratum_openNSA.txt)


habeebmohammed:
> Add the switch [here](https://github.com/stratum/stratum/blob/master/stratum/hal/config/BUILD#L36) to enable testing of the config files.

Add it in the platform config test.

PR #243, opened 2020-05-22
Unifiy Docker configuration mount process

### LINKED ISSUE #237: Rework Stratum Docker container startup
The current implementation has several shortcomings:

- Requires un-versioned files on the host
- Exposes internal interfaces to the user

### Configuration files
**All** configuration files must reside on `/root` of the host by default: https://github.com/stratum/stratum/blob/master/stratum/hal/bin/barefoot/docker/start-stratum-container.sh#L19. Further, it must follow out exact folder structure, else Stratum will not pick it up.
The README does mention the chassis config this briefly, but does not explain the needed structure. If the user simply starts the container, **Stratum will not run**.

For ZTP this creates a dependency on the host (file)system.

### Exposed internal interfaces
Arguably, the flags to provide paths to configuration files were not indented to be used as a mechanism to supply or even *change* configurations. Nonetheless they are used for that today. Reason include: missing APIs, ease of use for development.
Until recently is was not possible to push a (plain) chassis config or set ports in loopback mode over gNMI. Now these interfaces are implemented and provide a better way to configure things (e.g. without restarts and loss of all flows).

These flags are internal and thus not stable. E.g. the BCM sdk config file flags has initially be used to pass a config file to the SDK. Since the automatic generation of said file, this is not the case anymore and will instead **overwrite** any existing config without warning.

### Proposed changes

- Supply and use sane default configuration files, but provide options to override them if needed.
- Do not expect certain files to be present on the host. Exceptions would be the start script itself.
- Pass additional flags given to the start script to Stratum

### LINKED ISSUE #242: Stop on insmod failure
https://github.com/stratum/stratum/blob/a1a0e10431990753b0ef4b00de12194f30ee0cf8/stratum/hal/bin/barefoot/deb/stratum-entrypoint#L34

Stratum should not start if the Kernel module could not be inserted.

### LINKED ISSUE #265: Canonicalize Docker startup/config behaviour between BF and BCM
Currently they require different configs folder to be mounted and have different flags.
We should unify them as much as possible for consistent UX, within reasonable limits.

pudelkoM:
This PR eliminates differences in the way the config is passed to the docker startup scripts between BCM and BF.

It also ensures the default ChassisConfig is used on BF, if there is a one.

Mounting the whole config directory is removed in favor of two new env variables:
- `CHASSIS_CONFIG`: override the the default chassis config
- `FLAG_FILE`: override the default flag file, BF only

The platform config files are now under `/etc/stratum/<platform>` instead of `/etc/stratum/stratum_configs/<platform>`.

`BF_SDE_INSTALL`  is not required anymore. The standard `SDE_INSTALL` is used directly now.

Manual install of `stratum_bf` is deprecated and removed from the README. Main reasons:
- It's hard to get the all the dependencies right
- Makes the configs un-versioned
- `dpkg -x` allows user to extract all files and binaries to a chosen location without installing anything, achieving the same effect as before

`stratum-entrypoint` renamed to `start-stratum.sh` for consistency with BCM and the `start-stratum-container.sh` script.

Closes #242
Closes #265 
Closes #237


stratum/hal/bin/barefoot/start-stratum.sh
```diff
@@ -0,0 +1,39 @@
...
+set -ex
+
+KDRV_PATH=${KDRV_PATH:-/usr/lib/modules/bf_kdrv.ko}
+PLATFORM=${PLATFORM:-x86-64-accton-wedge100bf-32x-r0}
+
+if [ -f "/etc/onl/platform" ]; then
+    PLATFORM="$(cat /etc/onl/platform)"
+fi
+
+# Setup port map
```
Yi-Tseng, line 14: Set up

stratum/hal/bin/barefoot/start-stratum.sh
```diff
@@ -0,0 +1,39 @@
...
+rm /usr/share/port_map.json || true
```
Yi-Tseng, line 15: Use `rm -f` to ignore error?

stratum/hal/bin/barefoot/start-stratum.sh
```diff
@@ -0,0 +1,39 @@
...
+ln -s /etc/stratum/stratum_configs/$PLATFORM/port_map.json /usr/share/port_map.json
+
+# Load Kernel module
+rmmod bf_kdrv || true
+insmod $KDRV_PATH intr_mode="msi"
```
Yi-Tseng, line 20: This will fail if we are using docker images like `stratumproject/stratum-bf:9.1.0` since we don't include kernel module in it.

pudelkoM: I see.

stratum/hal/bin/barefoot/start-stratum.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
pudelkoM, line 1: Drop .sh sufix

stratum/hal/bin/barefoot/deb/stratum-entrypoint
```diff
@@ -34,9 +34,13 @@ if [ -f "$KDRV_PATH" ]; then
...
+        exit 255
     fi
 else
     echo "Cannot find $KDRV_PATH, skip installing the Kernel module."
 fi
 
-exec /usr/bin/stratum_bf -flagfile=$FLAG_FILE $@
+exec /usr/bin/stratum_bf \
+    -flagfile=$FLAG_FILE \
+    -chassis_config_file=/etc/stratum/$PLATFORM/chassis_config.pb.txt \
```
pudelkoM, line 45: Flag files do not allow bash variables.

Should we remove them altogether?

stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -9,14 +9,21 @@ if [ -n "$PLATFORM" ]; then
...
               -v /etc/onl:/etc/onl"
 fi
 
-CONFIG_DIR=${CONFIG_DIR:-/root}
+if [ -n "$FLAG_FILE" ]; then
+    FLAG_FILE_MOUNT="-v $FLAG_FILE:/etc/stratum/stratum.flags"
+fi
+
+if [ -n "$CHASSIS_CONFIG" ]; then
+    CHASSIS_CONFIG_MOUNT="-v $CHASSIS_CONFIG:/etc/stratum/$PLATFORM/chassis_config.pb.txt"
```
abhilashendurthi, line 24: If `$PLATFORM` is not initialized while calling `start-stratum-container.sh`, `CHASSIS_CONFIG_MOUNT` will mount the file to `/etc/stratum/chassis_config.pb.txt`. Since this isn't the default location for chassis_config, `stratum-entrypoint.sh` will continue using the older chassis_config from respective platform folder.

pudelkoM: All scripts now require the presence of `$PLATFORM` (or `/etc/onl/platform`)

stratum/hal/bin/barefoot/README.md
```diff
@@ -9,208 +9,167 @@ SPDX-License-Identifier: Apache-2.0
...
+```
 
-## Building Stratum from scratch
+The Docker image contains pre-built Stratum binary, latest Barefoot Software
+Development Environment (SDE) libraries, and default configuration files for all
+supported platforms.
 
-If you want or need to build Stratum yourself or if you need to make changes to how the BF SDE is built, you can follow
-these instructions.
+Building a Docker image yourself and more us covered in the [Docker README](./docker/README.md).
```
ccascone, line 24:
```suggestion
Building a Docker image yourself and more is covered in the [Docker README](./docker/README.md).
```

stratum/hal/bin/barefoot/README.md
```diff
@@ -9,208 +9,167 @@ SPDX-License-Identifier: Apache-2.0
...
 
-Barefoot's P4Studio Build tool comes with a default Stratum profile, which takes care of
-installing all the necessary dependencies and builds the SDE with the
-appropriate flags.
-
-Please follow these steps:
+Before you can build Stratum, the Barefoot SDE needs to be installed.
 
- 1. Extract the SDE: `tar -xzvf bf-sde-<SDE_VERSION>.tgz`
+Note: To do this, you will need to have SELA with Intel (Barefoot) to access P4
```
ccascone, line 40:
```suggestion
Note: To do this, you will need to have a SELA with Intel (Barefoot) to access P4
```

stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -26,10 +37,12 @@ DOCKER_IMAGE_TAG=${DOCKER_IMAGE_TAG:-$SDE_VERSION-$KERNEL_VERSION}
 docker run -it --privileged \
```
ccascone, line 37: I might be wrong, but I think the container should be removed on stop, or at least I don't see any references about restarting a stopped stratum container
```suggestion
docker run --rm -it --privileged \
```

pudelkoM: Good point. Added to all containers.

stratum/hal/bin/barefoot/BUILD
```diff
@@ -47,7 +47,7 @@ stratum_cc_binary(
 pkg_tar(
     name = "stratum_binaries",
     srcs = [
-        "deb/stratum-entrypoint",
+        "deb/stratum-entrypoint.sh",
```
pudelkoM, line 50: I think the extension is unnecessary, but for consistency either all or none should have them. For now it's all.

stratum/hal/bin/bcm/standalone/README.md
```diff
@@ -40,14 +40,14 @@ Note the **ONLPv2**!
...
 
-## Pre-build Docker image
+## Pre-built Docker image
 
 Stratum for Broadcom switches can be run inside Docker on the switch itself.
 As part of CI, we publish Stratum with a pre-compiled binary and a set of default configuration files as a [Docker container](https://hub.docker.com/repository/docker/stratumproject/stratum-bcm).
 
- - `cd stratum/hal/bin/bcm/standalone`
+ - `cd stratum/hal/bin/bcm/standalone/docker`
  - `docker pull stratumproject/stratum-bcm:latest`
```
Yi-Tseng, line 49: Should we remove the `docker pull` command? The `docker run` command should pull the image if the Docker can't find the image.

stratum/hal/bin/barefoot/README.md
```diff
@@ -264,39 +223,3 @@ To start a shell session, you can use (requires Docker):
...
-
-You can find the Debian package here:
-`bazel-bin/stratum/hal/bin/barefoot/stratum_bf_deb.deb`
-
-The Debian package includes some necessary libraries like PI so you don't need to install these libraries into the target device.
-
-To install the package, simply upload the deb file to your device and use `apt` command:
-
-```bash
-[sudo] apt update
```
pudelkoM, line 290: We should keep this.
A completely fresh switch has no cache and will not resolve the dependencies.

stratum/hal/bin/barefoot/README.md
```diff
@@ -264,39 +223,3 @@ To start a shell session, you can use (requires Docker):
...
-[sudo] apt install -y ./stratum_bf_deb.deb
```
pudelkoM, line 291: Also the `-y`

stratum/hal/bin/bcm/standalone/README.md
pudelkoM: It's still needed to get updates. I moved it to the top and added a comment.

stratum/hal/bin/barefoot/README.md
```diff
@@ -246,6 +209,25 @@ will configure device port 132 in 100G mode with Reed-Solomon (RS) FEC.
 
 FEC can also be configured when adding a port through gNMI.
 
+### Managing Stratum with systemd
+
+Systemd provides service management and Stratum has been integrated into it.
+
+Start Stratum service manually:
```
Yi-Tseng, line 216: `Start/Stop Stratum service manually`? Since there is a `# stop` comment

Yi-Tseng: And `Enable/Disable.....` below

PR #244, opened 2020-05-24
Bump BMv2, PI, and P4Runtime to fix action profile crash in stratum_bmv2

ccascone:
A bug in BMv2 caused the stratum_bmv2 process to crash when packets were processed by an action selector left in an inconsistent state. The issue is described here:
https://github.com/p4lang/behavioral-model/issues/893

This was causing PTF tests for fabric.p4 to fail:
https://github.com/opennetworkinglab/fabric-p4test/issues/72

To fix stratum_bmv2, this change bumps the bmv2 commit, which requires bumping PI and P4Runtime as well. P4Runtime is updated from 1.1.0-rc1 to backward-compatible 1.2.0-dev.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -2,8 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 ARG BAZEL_VERSION=3.0.0
-ARG PI_COMMIT=1539ecd8a50c159b011d9c5a9c0eba99f122a845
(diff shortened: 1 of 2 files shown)


pudelkoM:
Will do a quick check with the BCM side of Stratum.

pudelkoM:
No issues on BCM

PR #246, opened 2020-05-27
Add SDE 9.2.0 support

Yi-Tseng:


pudelkoM:
I can confirm that this PI commit hash works.

PR #248, opened 2020-05-28
Symlink SerDes firmware in tarball

### LINKED ISSUE #247: Stratum Tofino in Docker hangs on startup
When using the docker container of Stratum bf, it takes around one minute until bf_switchd actually runs.

Log:

```
Skipping P4 program load for dev_id 0
detecting.. IOMMU not enabled on the platform

[~1 min hang]

bf_switchd: dev_id 0 initialized
bf_switchd: initialized 1 devices
```

pudelkoM:
By default copies of the files are created. On startup the Tofino driver compares the links and updates the firmware if needed. When these links are not present, the update is always performed, leading to a ~1 min hang on startup.

Fixes #247


stratum/hal/lib/barefoot/barefoot.bzl
```diff
@@ -67,8 +67,20 @@ pkg_tar(
...
-         glob(["barefoot-bin/share/tofino_sds_fw/**"]),
+         glob(["barefoot-bin/share/tofino_sds_fw/credo/**"]) +
+         [
+            "barefoot-bin/share/tofino_sds_fw/avago/firmware/pcie_serdes.0x1066_2147.rom",
+            "barefoot-bin/share/tofino_sds_fw/avago/firmware/sbus_master.0x101D_2001.rom",
+            "barefoot-bin/share/tofino_sds_fw/avago/firmware/serdes.0x106F_204D.rom",
+            "barefoot-bin/share/tofino_sds_fw/avago/firmware/serdes.0x1091_244D.rom",
+         ],
   mode = "0644",
+  symlinks = {
```
bocon13, line 78: Can these be dynamically generated in a macro? I don't remember if globs are expanded in macro code or not.

stratum/hal/lib/barefoot/barefoot.bzl
```diff
@@ -67,8 +67,20 @@ pkg_tar(
   name = "bf_shareable_files",
   srcs = glob(["barefoot-bin/share/microp_fw/**"]) +
          glob(["barefoot-bin/share/bfsys/**"]) +
-         glob(["barefoot-bin/share/tofino_sds_fw/**"]),
+         glob(["barefoot-bin/share/tofino_sds_fw/credo/**"]) +
```
pudelkoM, line 70: `glob` has an `exclude` [param](https://docs.bazel.build/versions/master/be/functions.html#glob). Could simplify this rule.

pudelkoM:
https://github.com/bazelbuild/rules_pkg/issues/115

pudelkoM:
> Also, the current workaround in Stratum for that issue will no longer work in Bazel 6.0.0. See: [bazelbuild/bazel@a5e66fa](https://github.com/bazelbuild/bazel/commit/a5e66fafc9d1f9eb192bc157396040b6c95be5cf)

Thanks for the heads up, I'll take a look.

PR #252, opened 2020-06-02
Add sensible defaults to most flags

pudelkoM:
With the recent move to Debian packages we have default locations for most config files. When going through our start scripts/flag files, I found that most use the same value, often even the default.

This simplifies the startup scripts significantly.

**No flags or options are removed or changed by this PR**


stratum/hal/bin/barefoot/main.cc
```diff
@@ -25,7 +25,7 @@ int switch_pci_sysfs_str_get(char *name, size_t name_size);
 
 using ::pi::fe::proto::DeviceMgr;
 
-DEFINE_string(bf_sde_install, "",
+DEFINE_string(bf_sde_install, "/usr",
```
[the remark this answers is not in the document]
Yi-Tseng: The Debian package installs SDE to `/usr`

stratum/hal/bin/np4intel/docker/configs/stratum.flags
```diff
@@ -1,10 +1,6 @@
 -define phal_with_onlp=false
--external_stratum_urls=0.0.0.0:28000
--grpc_max_recv_msg_size=256
--persistent_config_dir=/stratum_configs/
 -forwarding_pipeline_configs_file=/stratum_configs/pipeline_config.pb.txt
 -dpdk_config=/stratum_configs/dpdk_config.pb.txt
 -chassis_config_file=/stratum_configs/chassis_config.pb.txt
 -write_req_log_file=/stratum_logs/p4_writes.pb.txt
```
Yi-Tseng, line 5: Change to default path like `/etc/stratum` and `/var/log/stratum` ?

pudelkoM: We could and should going forward. But I'm not familiar with the platform so I put of any changes there until we can try them ourself.

pudelkoM:
Tested stratum_bcm on a hardware switch with default configs.

stratum/hal/bin/np4intel/docker/configs/stratum.flags
pudelkoM: @craigsdell Please check there changes and add more if you think it also apply to the NP4 platform

PR #268, opened 2020-06-17
Clean up SDE build file

pudelkoM:
Backport of a internal BFRT change. Since we don't know when/if the BFRT version will be merged, I'll backport as much as possible.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -30,7 +30,7 @@ ENV PKG_DEPS pkg-config zip zlib1g-dev unzip python wget ca-certificates \
     lcov curl autoconf automake libtool libgmp-dev libpcap-dev \
     libboost-thread-dev libboost-filesystem-dev libboost-program-options-dev \
     gnupg2 software-properties-common python-pip python-dev python3-dev \
-    apt-transport-https libelf-dev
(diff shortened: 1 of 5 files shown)


Yi-Tseng:
Sounds good to me, should we also include `libbz2-dev` to our dev image?

pudelkoM:
> Sounds good to me, should we also include `libbz2-dev` to our dev image?

Yes, we need it to build libarchive. Would you include it here? I'd do it in the `bfrt` branch so we get it once we need it.

Yi-Tseng:
> > Sounds good to me, should we also include `libbz2-dev` to our dev image?
> 
> Yes, we need it to build libarchive. Would you include it here? I'd do it in the `bfrt` branch so we get it once we need it.

I think we can include it here so the CI will upload a new container image

pudelkoM:
> > > Sounds good to me, should we also include `libbz2-dev` to our dev image?
> > 
> > 
> > Yes, we need it to build libarchive. Would you include it here? I'd do it in the `bfrt` branch so we get it once we need it.
> 
> I think we can include it here so the CI will upload a new container image

Done, also properly fixed in `bfrt`

PR #269, opened 2020-06-17
Use upstream Bazel rules of P4Runtime

pudelkoM:
P4Rruntime reworked its Bazel rules (https://github.com/p4lang/p4runtime/pull/297), so we should be able to consume them directly now.


pudelkoM:
I did a quick tests with P4Runtime shell against stratum_bcm and found no issues.

PR #279, opened 2020-06-22
Automatically allocate huge pages in docker start script

pudelkoM:
I feel this should be handled in the `start-stratum.sh` script instead, so it also applies the the Debian package.

Is it possible to mount huge pages from inside the container?


stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -20,10 +20,18 @@ if [ -d "/etc/onl" ]; then
               -v /etc/onl:/etc/onl"
 fi
 
+# Prepare the host.
+free_hugepages=$(cat /sys/kernel/mm/hugepages/hugepages-2048kB/free_hugepages)
+if [[ $free_hugepages -lt 128 ]]; then
```
bocon13, line 25: Print a warning instead and exit.

bocon13:
@pudelkoM I just pushed a change. Please take a look.

I did some local testing, but we should verify on the PDP switch as well.

bocon13:
@bocon13 to test on hardware

pudelkoM (commented):
Just tried this on a switch (Debian version). Worked both with 128 free pages and no prior setup.

bocon13:
Found two issues: exit wasn't working on failure, sysctl was not working on the switch in Docker

Retested deb and container on hardware, and it works (when huge pages are and aren't allocated)

bocon13 (approved):

PR #281, opened 2020-06-24
Add libdru_sim.so for runs against tofino model

pudelkoM:
It is `dlopen`'d when using the tofino model as a backend.


Yi-Tseng:
@pudelkoM cant you test it before we merge it?

pudelkoM:
Tested against fabric-tna PTF suite.

PR #283, opened 2020-06-26
Add default config for Inventec D5264Q28D

Yi-Tseng:


pudelkoM:
Could you also add an entry in the build file?

PR #286, opened 2020-06-30
Use "exec form" for all Docker entrypoints

Yi-Tseng:


pudelkoM:
Only exec form passes arguments to the binary.

https://docs.docker.com/engine/reference/builder/#entrypoint

https://stackoverflow.com/a/32729888

stratum/hal/bin/bcm/standalone/docker/Dockerfile
```diff
@@ -11,4 +11,4 @@ RUN rm ./stratum_bcm_deb.deb
 
 EXPOSE 28000/tcp
 EXPOSE 9339/tcp
-ENTRYPOINT start-stratum.sh
+ENTRYPOINT ["/usr/bin/start-stratum.sh"]
```
pudelkoM, line 14: We should try the relative paths first, before hardcoding them. Otherwise we have the same information in two places (here and the BUILD file)

Yi-Tseng: make sense, since we placed the script in the default system binary path.

PR #288, opened 2020-07-09
Support querying counter from all interfaces

Yi-Tseng:
Tested on dummy switch only


stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -3709,6 +3709,60 @@ void YangParseTreePaths::AddSubtreeAllInterfaces(YangParseTree* tree) {
...
+                    [&event, &stream](const TreeNode& leaf) {
+                      return (leaf.GetOnPollHandler())(event, stream);
+                    });
+                // Notify the client that all nodes have been processed.
+                APPEND_STATUS_IF_ERROR(
+                    status, YangParseTreePaths::SendEndOfSeriesMessage(stream));
+                return status;
+              });
+
+
```
pudelkoM, line 3765: remove newline

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -3709,6 +3709,60 @@ void YangParseTreePaths::AddSubtreeAllInterfaces(YangParseTree* tree) {
                 return status;
               });
```
pudelkoM, line 3711: remove newline, see code above

pudelkoM:
Works on a real switch, too.

Usage example:
`bazel run //tools/gnmi:gnmi-cli -- --grpc-addr 10.128.13.241:28000 sub-sample /interfaces/interface[name="*"]/state/counters --interval 5000`

PR #290, opened 2020-07-10
Add a new way to include Barefoot SDE dependencies

Yi-Tseng:
The developer can package the entire SDE install directory to a tarball file and copy it to the build system. And Bazel can use that tarball to build stratum

TODO: doc


Yi-Tseng:
duplicate with new one

PR #292, opened 2020-07-14
Disable P4 write logging if flag is empty

### LINKED ISSUE #291: Disable to P4 write request logging
Do not create or write P4 write request logs to a file if `-max_num_controllers_per_node` is not set(empty)
This can improve the write performance a bit since it blocks each write requests

Yi-Tseng:
fixes #291


pudelkoM:
Overall idea seems good to me. With the default path we set, this feature stays enabled by default.

Why not move the check into the actual logging function? https://github.com/stratum/stratum/blob/master/stratum/hal/lib/common/p4_service.cc#L209-L231
Would solve this for all call sites.

Yi-Tseng:
> Overall idea seems good to me. With the default path we set, this feature stays enabled by default.
> 
> Why not move the check into the actual logging function? https://github.com/stratum/stratum/blob/master/stratum/hal/lib/common/p4_service.cc#L209-L231
> Would solve this for all call sites.
> 
> ```c++
> void LogWriteRequest(uint64 node_id, const ::p4::v1::WriteRequest& req,
>                      const std::vector<::util::Status>& results,
>                      const absl::Time timestamp) {
>   if (!FLAGS_write_req_log_file.empty()) {
>     return;
>   }
>   if (results.size() != req.updates_size()) {
>     LOG(ERROR) << "Size mismatch: " << results.size()
>                << " != " << req.updates_size() << ". Did not log anything!";
>     return;
>   }
> ```

make sense, done.

stratum/hal/lib/common/p4_service.cc
```diff
@@ -209,6 +209,10 @@ ::grpc::Status ToGrpcStatus(const ::util::Status& status,
 void LogWriteRequest(uint64 node_id, const ::p4::v1::WriteRequest& req,
                      const std::vector<::util::Status>& results,
                      const absl::Time timestamp) {
+  // Skip logging if flag is empty.
```
pudelkoM, line 212: I'll leave it up to you if this falls under: [https://google.github.io/styleguide/cppguide.html#Implementation_Comment_Donts](https://google.github.io/styleguide/cppguide.html#Implementation_Comment_Donts)

PR #294, opened 2020-07-15
Reject ChassisConfig changes that modify the port layout

### LINKED ISSUE #293: gNMI Get fails with DEADLINE_EXCEEDED on stratum_bf
**Describe the bug**
A gNMI-related thread seems to crash (or at least a crash-like stack trace is printed, see stratum.log) when ONOS issues a Get RPC to read interface counters. This happens consistently when restarting the stratum_bf container after ONOS has already established a first successful connection (including a pipeline push, interface discovery, etc).

[stratum.log](https://github.com/stratum/stratum/files/4916840/stratum.log)
[chassis-config.pb.txt](https://github.com/stratum/stratum/files/4916844/stratum-leaf1-chassis-config.pb.txt)
[onos-grpc.log](https://github.com/stratum/stratum/files/4916848/onos-grpc.log) (look for `callId=1300` for an example of Get RPC that fails)

**Environment**
latest `stratumproject/stratum-bf:9.0.0-4.14.49-OpenNetworkLinux` container on Wedge 100BF-32X execute with:

    CHASSIS_CONFIG=/root/chassis_config.pb.txt ./start-stratum-container.sh --bf-sim

Switch:
```bash
root@aether-tofino:~# cat /etc/onl/platform
x86-64-accton-wedge100bf-65x-r0
root@aether-tofino:~# uname -a
Linux aether-tofino 4.14.49-OpenNetworkLinux #1 SMP Thu Jun 25 17:42:47 UTC 2020 x86_64 GNU/Linux
root@aether-tofino:~# lsmod
Module                  Size  Used by
bf_kdrv                28672  2
x86_pkg_temp_thermal    16384  0
gpio_ich               16384  0
```

**To Reproduce**
Restart the Stratum container after ONOS has already connected to it once.

**Expected behavior**
The Get RPC should be completed successfully instead of a DEADLINE_EXCEEDED exception.

pudelkoM:
Stratum internals (like the yang tree) cannot handle such changes at runtime. Therefore we save the new config and require a reboot. Individual port config changes (speed, admin state) are still possible.

Tested with bfrt, but at least one run with stratum_bf would be good, too.

Steps:
- Start without chassis config: `start-stratum.sh  --chassis_config_file=/tmp/foo.pb.txt`
- Push new, full config: `bazel run //tools/gnmi:gnmi-cli -- --grpc-addr 10.128.13.223:28000 --bytes_val_file /stratum/stratum/hal/config/x86-64-accton-wedge100bf-32x-r0/chassis_config.pb.txt --replace set "/"`
- Check that all port are there: `pm show`
- Delete some ports from the config
- Push again. Expected error: `The switch is already initialized, but we detected the newly pushed config requires a change in the port layout. The stack needs to be rebooted to finish config push.`
- Restart Stratum
- Check ports, should be less than before
- Re-add ports to config
- Push and restart
- Check port all ports again

Fixes #293

### DIFF
--- a/stratum/hal/lib/barefoot/bf_chassis_manager.cc
+++ b/stratum/hal/lib/barefoot/bf_chassis_manager.cc
@@ -351,6 +351,38 @@ BFChassisManager::~BFChassisManager() = default;
 
 ::util::Status BFChassisManager::VerifyChassisConfig(
     const ChassisConfig& config) {
+  if (config.trunk_ports_size()) {
+    return MAKE_ERROR(ERR_INVALID_PARAM)
+           << "Trunk ports are not supported on Tofino.";
+  }
+  if (config.port_groups_size()) {
+    return MAKE_ERROR(ERR_INVALID_PARAM)
+           << "Port groups are not supported on Tofino.";
+  }
+
+  // If the class is initialized, we also need to check if the new config will
+  // require a change in the port layout. If so, report reboot required.
+  if (initialized_) {
+    std::map<uint64, std::map<uint32, PortKey>>
+        node_id_to_port_id_to_singleton_port_key;
+
+    for (const auto& singleton_port : config.singleton_ports()) {
+      uint32 port_id = singleton_port.id();
+      uint64 node_id = singleton_port.node();
+
+      node_id_to_port_id_to_singleton_port_key[node_id][port_id] =
+          PortKey(singleton_port.slot(), singleton_port.port(),
+                  singleton_port.channel());
+    }
+
+    if (node_id_to_port_id_to_singleton_port_key !=
+        node_id_to_port_id_to_singleton_port_key_) {
+      return MAKE_ERROR(ERR_REBOOT_REQUIRED)
+             << "The switch is already initialized, but we detected the "
+             << "newly pushed config requires a change in the port layout. "
+             << "The stack needs to be rebooted to finish config push.";
+    }
+  }
   return ::util::OkStatus();
 }
 
--- a/stratum/hal/lib/barefoot/bf_switch.cc
+++ b/stratum/hal/lib/barefoot/bf_switch.cc
@@ -88,9 +88,9 @@ BFSwitch::~BFSwitch() {}
     for (const auto& entry : node_id_to_unit) {
       uint64 node_id = entry.first;
       int unit = entry.second;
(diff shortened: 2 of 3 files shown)


stratum/hal/lib/barefoot/bf_switch.cc
```diff
@@ -43,6 +43,9 @@ BFSwitch::BFSwitch(PhalInterface* phal_interface,
 BFSwitch::~BFSwitch() {}
 
 ::util::Status BFSwitch::PushChassisConfig(const ChassisConfig& config) {
+  // Verify the config first. No need to continue if verification is not OK.
+  // Push config to PHAL first and then the rest of the managers.
```
pudelkoM, line 47: This comment should be two lines below.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -354,7 +354,40 @@ ::util::Status BFChassisManager::PushChassisConfig(
...
+  }
+
+  // If the class is initialized, we also need to check if the new config will
+  // require a change in the port layout. If so, report reboot required.
+  if (initialized_) {
+    std::map<uint64, std::map<uint32, PortKey>>
+        node_id_to_port_id_to_singleton_port_key;
+
+    for (const auto& singleton_port : config.singleton_ports()) {
+      LOG(WARNING) << "Checking new port " << singleton_port.ShortDebugString();
```
pudelkoM, line 374: remove

stratum/hal/lib/barefoot/bf_switch.cc
```diff
@@ -63,8 +66,39 @@ ::util::Status BFSwitch::PushChassisConfig(const ChassisConfig& config) {
...
+  if (!ret.ok()) {
+    if (ret.status().error_code() != ERR_NOT_INITIALIZED) {
+      APPEND_STATUS_IF_ERROR(status, ret.status());
+    }
+  } else {
+    const auto& node_id_to_unit = ret.ValueOrDie();
+    for (const auto& entry : node_id_to_unit) {
+      uint64 node_id = entry.first;
+      int unit = entry.second;
+      ASSIGN_OR_RETURN(auto* bfrt_node, GetPINodeFromUnit(unit));
```
Yi-Tseng, line 91: pi_node

PR #295, opened 2020-07-15
Fix formatting in barefoot code

pudelkoM:
No other code changes.


stratum/hal/lib/barefoot/bf_pal_wrapper.h
```diff
@@ -22,11 +22,11 @@ class BFPalWrapper : public BFPalInterface {
  public:
   static constexpr int32 kDefaultMtu = 10 * 1024;  // 10K
```
pudelkoM, line 23: Should be in constants.h

PR #303, opened 2020-07-30
Add compat code to enable basic MacOS support

pudelkoM:
While Stratum (the SwitchOS) will probably never run on MacOS, we have related tools for development and deployment that could.
This PR allows us to build and ship native tools instead of docker images. Tested with `gnmi-cli`.

### DIFF
--- a/stratum/glue/status/posix_error_space.cc
+++ b/stratum/glue/status/posix_error_space.cc
@@ -2,15 +2,147 @@
 // Copyright 2018-present Open Networking Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-
 #include "stratum/glue/status/posix_error_space.h"
 
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


stratum/glue/status/posix_error_space.cc
```diff
@@ -2,16 +2,148 @@
...
 #include <string>
 
 #include "stratum/glue/logging.h"
 
+// Fake missing error codes on MacOS. Note, that they still have to be unique
+// and must not overlap with existing ones.
+#ifdef __APPLE__
+#ifndef ENOMEDIUM
+#define ENOMEDIUM (ELAST + __COUNTER__)
+#endif
```
pudelkoM, line 20: Since macros can't use preprocessor statements themselves, this block is written out by hand.

PR #304, opened 2020-07-31
Move tools under stratum root

pudelkoM:
While not all tools are C++ programs, some rely on Stratum libraries. Moving them here allows the visibility of the packages to remain stratum-private.

This is in preparation for:
- Adding more tools (pipeline builder & pusher)
- Building C++ with common compiler flags


stratum/hal/config/BUILD
```diff
@@ -11,7 +11,7 @@ load(
 load(":platform_config_test.bzl", "platform_config_test")
 load("@rules_pkg//:pkg.bzl", "pkg_tar", "pkg_deb")
 
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = STRATUM_INTERNAL)
```
pudelkoM, line 14: @Yi-Tseng Do you remember why you changed this to public?

pudelkoM:
I'm not sure I we should move all tools under `/stratum`. The mininet, TLS and onos-cfg don't depend on Stratum and we have external dependencies on them.

stratum/hal/config/BUILD
Yi-Tseng: I think I was trying to use create some tools with Bazel to use these configs. But I didn't add those tests to this repo in the end and forgot to revert this change.

Yi-Tseng: I think I was trying to use create some tools with Bazel to use these configs. But I didn't add those tests to this repo in the end and forgot to revert this change.

pudelkoM: Ok, with this change new tools can live under `/stratum` and we can leave this private.

PR #305, opened 2020-07-31
More MacOS fixes

pudelkoM:


stratum/glue/net_util/ipaddress.cc
```diff
@@ -9,7 +9,6 @@
 
 #include <arpa/inet.h>
 #include <assert.h>
-#include <byteswap.h>
```
pudelkoM, line 12: ipaddress.cc includes `bits.h`, which has the fix.

PR #306, opened 2020-08-03
Add Stratum wide default values for copts, linkopts and testing copts

### LINKED ISSUE #102: Set bazel copts and/or linkopts on Stratum targets
Setting flags on all our targets is preferable to global flags via `.bazelrc`, because the later applies to ALL code, including external. This limits the options we can actually use. Enabling all warnings for example results in an useless torrent of compiler output from code we have no influence over. External code also does not always support instrumentation and breaks if compiled with, e.g., ASAN.

[Abseil](https://github.com/abseil/abseil-cpp/tree/master/absl/copts) does it this way:
```python
cc_library(
    name = "compressed_tuple",
    hdrs = ["internal/compressed_tuple.h"],
    copts = ABSL_DEFAULT_COPTS,
    linkopts = ABSL_DEFAULT_LINKOPTS,
    deps = [
        "//absl/utility",
    ],
)
```

This is a huge change LoC wise, but could probably be automated fairly easily.

Some of the flags we could set:

- `-Wno-unused-function`
- `-Wno-unused-lambda-capture`
- `-Werror=thread-safety-reference`
- `-Werror=enum-compare`
- `-Werror=deprecated-declarations`

pudelkoM:
The specific flag values are taken verbatim from [Abseil](https://github.com/abseil/abseil-cpp/blob/518f17501e6156f7921fbb9b68a1e420bcb10bc5/absl/copts/GENERATED_copts.bzl).

This will allow us to transition from global copts in `.bazelrc` to per-target specific options. 
Compiler matching is performed based on the `"@bazel_tools//tools/cpp:compiler"` flag value. This values is always defined, even when the `--compiler=foo` CLI flag is not set.

The default set of options is applied to all `stratum_cc_library` and `stratum_cc_binary` targets. Currently it can not be overwritten, just extended. Since not all targets use the `stratum_*` rules yet, we can use that as a toggle to disable extended checking for impacted targets, until fixed later.

Fixes #102


bazel/defs.bzl
```diff
@@ -10,7 +10,28 @@
...
+    "ABSL_LLVM_TEST_FLAGS",
+)
+
+EMBEDDED_ARCHES = ["x86"]
+HOST_ARCHES = ["x86"]
+STRATUM_INTERNAL = ["//stratum:__subpackages__"]
+STRATUM_DEFAULT_COPTS = select({
+    "//stratum:llvm_compiler": ABSL_LLVM_FLAGS,
+    "//conditions:default": ABSL_GCC_FLAGS,
+})
```
pudelkoM, line 30: Add some flags that we care about, like
`-Wimplicit-fallthrough` and `-Wno-sign-compare`

bazel/defs.bzl
```diff
@@ -10,7 +10,63 @@
...
+    "-Wno-unused-variable",
+]
+STRATUM_DISABLED_COMPILER_WARNINGS_GCC = []
+STRATUM_DISABLED_COMPILER_WARNINGS_LLVM = [
+    "-Wno-gnu-zero-variadic-macro-arguments",
+    "-Wno-missing-variable-declarations",
+    "-Wno-suggest-destructor-override",
+    "-Wno-unused-lambda-capture",
+    "-Wno-unused-template",
+]
```
pudelkoM, line 42: I'm not super sold on the warnings, since these still apply to included external headers, and therefore produce a lot of text. But the output is much smaller for iterative cycles on the same files and might catch something there.

For egregious violations we still have the `-Werror` hammer.

stratum/tools/gnmi/BUILD
```diff
@@ -4,6 +4,8 @@
 load(
     "//bazel:rules.bzl",
     "HOST_ARCHES",
+    "STRATUM_DEFAULT_COPTS",
+    "STRATUM_DEFAULT_LINKOPTS",
```
pudelkoM, line 8: Revert, the defaults automatically apply to all `stratum_cc_*` targets now.

bazel/defs.bzl
```diff
@@ -10,7 +10,65 @@
...
+    "ABSL_LLVM_FLAGS",
+    "ABSL_LLVM_TEST_FLAGS",
+)
+
+EMBEDDED_ARCHES = ["x86"]
+HOST_ARCHES = ["x86"]
+STRATUM_INTERNAL = ["//stratum:__subpackages__"]
+
+# Compiler warnings for different toolchains.
+STRATUM_DISABLED_COMPILER_WARNINGS_COMMON = [
```
Yi-Tseng, line 29: add `-Wall`

pudelkoM:
The latest commits fixing random code are just for staging in here. We'll have to make them separate PRs later.

bazel/defs.bzl
```diff
@@ -10,7 +10,72 @@
...
+})
+
+# Compiler warnings that are threated as errors.
+STRATUM_COMPILER_ERRORS_COMMON = [
+    "-Werror=implicit-fallthrough",
+    "-Werror=ignored-attributes",
+]
+STRATUM_COMPILER_ERRORS_GCC = []
+STRATUM_COMPILER_ERRORS_LLVM = [
+    "-Werror=thread-safety-analysis",
```
pudelkoM, line 58: `-Wreorder-ctor`

bazel/defs.bzl
```diff
@@ -10,7 +10,76 @@
...
+})
+
+# Exported default compiler options for use.
+STRATUM_DEFAULT_COPTS = select({
+                            "//stratum:llvm_compiler": ABSL_LLVM_FLAGS,
+                            "//conditions:default": ABSL_GCC_FLAGS,
+                        }) + \
+                        STRATUM_COMPILER_WARNINGS_COMMON + \
+                        STRATUM_DISABLED_COMPILER_WARNINGS + \
+                        STRATUM_COMPILER_ERRORS
```
pudelkoM, line 76: I feel like this selection is getting too complex. Maybe one complete list per compiler is the way to go.

bazel/defs.bzl
```diff
@@ -10,7 +10,76 @@
...
+    "-Wno-unused-template",
+]
+STRATUM_DISABLED_COMPILER_WARNINGS = STRATUM_DISABLED_COMPILER_WARNINGS_COMMON + select({
+    "//stratum:llvm_compiler": STRATUM_DISABLED_COMPILER_WARNINGS_LLVM,
+    "//conditions:default": STRATUM_DISABLED_COMPILER_WARNINGS_GCC,
+})
+
+# Compiler warnings that are threated as errors.
+STRATUM_COMPILER_ERRORS_COMMON = [
+    "-Werror=ignored-attributes",
```
pudelkoM, line 53: `-Wreturn-type`

bazel/rules/binary_rule.bzl
```diff
@@ -1,22 +1,38 @@
...
+    cc_binary(
+        name = name,
+        deps = deps,
+        srcs = srcs,
+        data = data,
+        args = args,
+        copts = STRATUM_DEFAULT_COPTS + copts,
+        defines = defines,
+        includes = includes,
+        linkopts = linkopts,
```
pudelkoM, line 35: Add default link opts, even if empty right now.

bazel/defs.bzl
```diff
@@ -10,7 +10,77 @@
...
+    "-Werror=reorder",
+]
+STRATUM_COMPILER_ERRORS_LLVM = [
+    "-Werror=implicit-fallthrough",  # TODO(max): move to common after gcc 7
+    "-Werror=reorder-ctor",
+    "-Werror=thread-safety-analysis",
+]
+STRATUM_COMPILER_ERRORS = STRATUM_COMPILER_ERRORS_COMMON + select({
+    "//stratum:llvm_compiler": STRATUM_COMPILER_ERRORS_LLVM,
+    "//conditions:default": STRATUM_COMPILER_ERRORS_GCC,
```
pudelkoM, line 67: Match on gcc, empty default

bocon13 (dismissed):

bazel/defs.bzl
pudelkoM: GCC is not detected by Bazel: https://github.com/bazelbuild/bazel/issues/12707

But using gcc by default seems common practice and safe.

stratum/hal/lib/barefoot/BUILD
```diff
@@ -66,6 +66,7 @@ stratum_cc_library(
     name = "bf_chassis_manager",
     srcs = ["bf_chassis_manager.cc"],
     hdrs = ["bf_chassis_manager.h"],
+    copts = ["-Wno-thread-safety-analysis"],  # TODO(antonin)
```
pudelkoM, line 69: This is ok, even though gcc does not support it:

> However, if the -Wno- form is used, the behavior is slightly different: no diagnostic is produced for -Wno-unknown-warning unless other diagnostics are being produced.

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

PR #307, opened 2020-08-05
Change log level in timer daemon class

Yi-Tseng:


stratum/lib/timer_daemon.cc
```diff
@@ -68,9 +68,9 @@ bool TimerDaemon::Execute() {
   if (front != nullptr) {
     // Execute the timer's action!
     if (front->ExecuteAction() != ::util::OkStatus()) {
-      LOG(ERROR) << "something went wrong";
+      VLOG(0) << "something went wrong";
```
pudelkoM, line 71: Since all timer actions return a status, I'd actually log the status as error here.

```suggestion
      LOG(ERRRO) << "Error executing action: " << status;
```

PR #309, opened 2020-08-07
Update bmv2 README

pudelkoM:
Building outside of the provided development container is not supported anymore. If running `stratum_bmv2` outside of it is a valid use case, we must build a debian package with all dependencies for distribution.


pudelkoM:
> For those who don't want to build stratum_bmv2, we should mention that binaries are distributed as part of the mn-stratum docker image, which is built by CI from master.

Good point. I remembered the mininet image when looking for references of `stratum_bmv2`.

PR #311, opened 2020-08-18
Add node info to SwitchInterface and expose via gNMI

### LINKED ISSUE #166: Add gNMI paths to support device description discovery
```
/components/component/state/mfg-name
/components/component/state/mfg-date
/components/component/state/hardware-version
/components/component/state/software-version
/components/component/state/serial-no
```

Reference: https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform.yang#L338-L388

pudelkoM:
This PR adds a `NodeInfo` message to the `SwitchInterface` to retrieve data about the switching ASIC. It implements basic support for Tofino (no chip detection) and dummy, and full support for Broadcom.

There are a few open questions:
- <s>The component name in the gNMI path is constructed from the node name, which is not set in most of the chassis configs</s> `node-<id>` if node name is empty

New paths:
- `/components/component[name="dummy node 1"]/state/part-no` => TOMAHAWK, TRIDENT2
- `/components/component[name="dummy node 1"]/state/mfg-name` => Broadcom, Barefoot
- `/components/component[name="dummy node 1"]/state/type` => INTEGRATED_CIRCUIT

gNMI command:
```bash
bazel run //stratum/tools/gnmi:gnmi-cli -- --grpc-addr 127.0.0.1:28000 get /components/component[name="dummy node 1"]/state
```

Fixes #166

### DIFF
--- a/stratum/hal/lib/common/common.proto
+++ b/stratum/hal/lib/common/common.proto
@@ -1134,7 +1134,7 @@ message Alarm {
 }
 
 message NodeInfo {
-  string vendor = 1;
+  string vendor_name = 1;
   string chip_name = 2;
 }
 
--- a/stratum/hal/lib/common/common.proto
+++ b/stratum/hal/lib/common/common.proto
@@ -1133,6 +1133,11 @@ message Alarm {
   bool status = 4;          // required.
 }
 
+message NodeInfo {
+  string vendor = 1;
+  string chip_name = 2;
+}
+
 // Wrapper around an optional debug info for a node/chip.
 message NodeDebugInfo {
   string debug_string = 1;
@@ -1214,6 +1219,7 @@ message DataRequest {
       Port fec_status = 18;
       OpticalNetworkInterface optical_transceiver_info = 19;
       Port loopback_status = 20;
+      Node node_info = 21;
     }
   }
   repeated Request requests = 1;
@@ -1245,6 +1251,7 @@ message DataResponse {
     FecStatus fec_status = 18;
     OpticalTransceiverInfo optical_transceiver_info = 19;
     LoopbackStatus loopback_status = 20;
+    NodeInfo node_info = 21;
   }
 }
 
--- a/stratum/hal/lib/common/common.proto
+++ b/stratum/hal/lib/common/common.proto
@@ -1133,6 +1133,7 @@ message Alarm {
   bool status = 4;          // required.
 }
 
+// Wrapper around the optional node info.
 message NodeInfo {
   string vendor_name = 1;
   string chip_name = 2;
--- a/stratum/hal/lib/common/yang_parse_tree_paths.cc
+++ b/stratum/hal/lib/common/yang_parse_tree_paths.cc
@@ -2950,8 +2950,7 @@ void SetUpComponentsComponentName(const std::string& name, TreeNode* node) {
 
 ////////////////////////////////////////////////////////////////////////////////
 // /components/component[name=<name>]/state/type
-void SetUpComponentsComponentStateType(const std::string& name,
-                                       const std::string& type,
+void SetUpComponentsComponentStateType(const std::string& type,
                                        TreeNode* node) {
   auto poll_functor = [type](const GnmiEvent& /*event*/,
                              const ::gnmi::Path& path,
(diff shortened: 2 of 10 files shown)


stratum/hal/lib/common/common.proto
```diff
@@ -1133,6 +1133,11 @@ message Alarm {
   bool status = 4;          // required.
 }
 
+message NodeInfo {
+  string vendor = 1;
```
pudelkoM, line 1137: vendor_name ?

ccascone:
How about using the node ID in gNMI paths instead of the node name? I think using the node name here might lead to user errors more easily (e.g., "node 1" vs "Node 1"). Moreover, the node ID is already set in most chassis config files (I would assume it's mandatory), and it makes it easier for a controller to correlate P4Runtime device IDs to the corresponding ASIC info.

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -3587,15 +3621,18 @@ void YangParseTreePaths::AddSubtreeNode(const Node& node, YangParseTree* tree) {
...
+  SetUpComponentsComponentIntegratedCircuitConfigNodeId(node.id(), tree_node,
+                                                        tree);
+  tree_node = tree->AddNode(GetPath("components")(
+      "component", node.name())("integrated-circuit")("state")("node-id")());
+  SetUpComponentsComponentIntegratedCircuitStateNodeId(node.id(), tree_node,
+                                                       tree);
+  tree_node = tree->AddNode(GetPath("components")(
+      "component", node.name())("integrated-circuit")("state")("part-no")());
+  SetUpComponentsComponentIntegratedCircuitStatePartNo(node.id(), tree_node,
+                                                       tree);
```
Yi-Tseng, line 3635: We can add `augment` to the Stratum model(`openconfig-platform-stratum.yang`) since paths described here do not exist.
https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform.yang#L688-L702
http://www.yang-central.org/twiki/pub/Main/YangDocuments/draft-ietf-netmod-yang-02.html#rfc.section.4.2.8

stratum/public/yang/openconfig-platform-stratum.yang
```diff
@@ -76,4 +76,13 @@ module openconfig-platform-stratum {
       description "Stratum does not support components/component/backplane";
       deviate not-supported;
   }
+
+  augment "/oc-platform:components/oc-platform:component/oc-platform:integrated-circuit/oc-platform:state" {
+      leaf part-no {
+          type string;
+      }
+      leaf mfg-name {
+          type string;
```
Yi-Tseng, line 85:
```yang
description "Name for the manufacturer of the component."
```

stratum/public/yang/openconfig-platform-stratum.yang
```diff
@@ -76,4 +76,13 @@ module openconfig-platform-stratum {
       description "Stratum does not support components/component/backplane";
       deviate not-supported;
   }
+
+  augment "/oc-platform:components/oc-platform:component/oc-platform:integrated-circuit/oc-platform:state" {
+      leaf part-no {
+          type string;
```
Yi-Tseng, line 82:
```yang
description "Part number for the component (e.g. Chip model or name)."
```

Yi-Tseng:
My original thought is to use the following paths:

```
# Exists in OpenConfig model
/components/component[name="node-1"]/type -> return "INTEGRATED_CIRCUIT"
/components/component[name="node-1"]/mfg-name -> return "Barefoot"
/components/component[name="node-1"]/part-no -> return "BFN_XXXXX"

# Hercules extension
/components/component[name="node-1"]/integrated-circuit/state/node-id -> return 1
```

pudelkoM:
> LGTM overall.
> One place also needs to be modified:
> https://github.com/stratum/stratum/blob/master/stratum/hal/lib/common/yang_parse_tree.cc#L199
> Here we store node names to a set to check if there is a duplicate name. Will return error if all node names are empty. (Won't happen now, but may happen in the future)

The check should be removed. The node name is optional in the protobuf, and we use the unique ID anyway.

Yi-Tseng:
> The check should be removed. The node name is optional in the protobuf, and we use the unique ID anyway.
Agree, since we don't use that name to create the three. Feel free to remove that in this patch.

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -3579,28 +3644,40 @@ void YangParseTreePaths::AddSubtreeInterfaceFromOptical(
...
+  tree_node = tree->AddNode(
+      GetPath("components")("component", name)("state")("mfg-name")());
+  SetUpComponentsComponentStateMfgName(node.id(), tree_node, tree);
 }
 
 void YangParseTreePaths::AddSubtreeChassis(const Chassis& chassis,
                                            YangParseTree* tree) {
-  const std::string& name = chassis.name();
+  const std::string& name =
+      !chassis.name().empty() ? chassis.name() : "chassis";
```
pudelkoM, line 3680: I could not find a unique id in the Chassis proto message. Might be fine, since most configs have this name and there is only on chassis per yang tree.

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -3377,7 +3436,10 @@ void YangParseTreePaths::AddSubtreeInterfaceFromTrunk(
 void YangParseTreePaths::AddSubtreeInterfaceFromSingleton(
     const SingletonPort& singleton, const NodeConfigParams& node_config,
     YangParseTree* tree) {
-  const std::string& name = singleton.name();
+  const std::string& name =
+      !singleton.name().empty()
+          ? singleton.name()
+          : absl::StrFormat("%d/%d", singleton.port(), singleton.slot());
```
Yi-Tseng, line 3442: You also need to include the channel number since we can have multiple interfaces per channel on same physical port+slot

pudelkoM: So the complete order would be slot/port/channel?

stratum/hal/lib/common/yang_parse_tree_paths.cc
Yi-Tseng: looks good to me

pudelkoM:
Checking for duplicates or invalid names at yang tree creation is far too late. The chassis config has to be verified at the edge, when it is pushed by the user.

PR #313, opened 2020-08-24
Disable file logging on Stratum docker

pudelkoM:
This PR disables logging to disk on the Docker builds of Stratum.

By setting `-log_dir=/dev/null`, no logs will be created.
There are minor problems with this solution. glog will complain that it can not create sub-directories and print this message once for each level:
```
COULD NOT CREATE A LOGGINGFILE 20200824-165925.31866!W0824 16:59:25.207049 31866 credentials_manager.cc:45] Using insecure server credentials
```
This could be a problem for downstream log parsers, like fluentd.

Alternatives:
`-minloglevel` cannot be used, as it completely disables logging, even to stderr.

`-max_log_size=0` can be used to limit file size to 1MB. If we are concerned about polluting disk. Still writes the logs to disk, if performance is a concern.


pudelkoM:
Since these changes do not provide a good experience out-of-the-box, they will not become the defaults.

The discussion around production configurations is continued in #322

PR #317, opened 2020-08-25
Address concerns about yang chip type feature

pudelkoM:
PR #311 added the IC component to the yang tree. This PR fixes some errors around that.

- Return the BCM chip numbers instead of the nick names (e.g. BCM56965)
- <s>Make names in ChassisConfig mandatory</s> Breaks too much existing code / convention and is redundant with the id field
- <s>Removes the name generation in yang tree paths, name are now taken verbatim only</s> Names are again taken from the chassis config, or generated if not present.
- <s>Fixes the supplied configs</s>
- Validate that component names, if given, are unique


stratum/hal/lib/common/config_monitoring_service.cc
```diff
@@ -131,6 +134,60 @@ ::util::Status ConfigMonitoringService::PushChassisConfig(
...
+  for (const auto& singleton_port : config.singleton_ports()) {
+    CHECK_RETURN_IF_FALSE(!singleton_port.name().empty())
+        << "SingletonPort name cannot be empty.";
+  }
+  CHECK_RETURN_IF_FALSE(AreUniqueNames(config.singleton_ports()));
+  CHECK_RETURN_IF_FALSE(
+      gtl::STLIsUnique(config.singleton_ports(),
+                       [](const SingletonPort& a, const SingletonPort& b) {
+                         return a.name() == b.name();
+                       }));
```
pudelkoM, line 169: Which implementation should we chose?

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -3437,10 +3450,10 @@ void YangParseTreePaths::AddSubtreeInterfaceFromSingleton(
     const SingletonPort& singleton, const NodeConfigParams& node_config,
     YangParseTree* tree) {
   const std::string& name =
-      !singleton.name().empty()
-          ? singleton.name()
-          : absl::StrFormat("%d/%d/%d", singleton.slot(), singleton.port(),
-                            singleton.channel());
+      singleton.name().empty()
+          ? absl::StrFormat("%d/%d/%d", singleton.slot(), singleton.port(),
```
Yi-Tseng, line 3454: Should we also add a prefix to the singleton port? since we have the prefix for the optical port, node, and chassis.

Yi-Tseng (dismissed):
Some minor things.

stratum/hal/lib/common/yang_parse_tree_paths.cc
pudelkoM: Good question. This scheme comes from the name we currently have in all shipped ChassisConfigs, so I'd assume that they do not cause issues.

stratum/hal/lib/common/config_monitoring_service.cc
```diff
@@ -131,6 +134,39 @@ ::util::Status ConfigMonitoringService::PushChassisConfig(
...
+::util::Status ConfigMonitoringService::VerifyChassisConfig(
+    const ChassisConfig& config) {
+  // Validate the names of the components, if given.
+  CHECK_RETURN_IF_FALSE(AreUniqueNames(config.nodes()));
+  CHECK_RETURN_IF_FALSE(AreUniqueNames(config.singleton_ports()));
+  CHECK_RETURN_IF_FALSE(gtl::STLIsUnique(
+      config.singleton_ports(),
+      [](const SingletonPort& a, const SingletonPort& b) {
+        return !a.name().empty() && !b.name().empty() && a.name() == b.name();
+      }));
```
pudelkoM, line 162: Decide on implementation approach.

stratum/hal/lib/common/yang_parse_tree_paths.cc
Yi-Tseng: ok, that's fine to me

stratum/hal/lib/common/config_monitoring_service.cc
```diff
@@ -131,6 +134,39 @@ ::util::Status ConfigMonitoringService::PushChassisConfig(
   return ::util::OkStatus();
 }
 
+namespace {
+// Helper function to determine whether a container of protobuf messages have an
+// unique name field.
+template <typename T>
+bool AreUniqueNames(const T& values) {
```
pudelkoM, line 141: `HaveUniqueNames`

stratum/hal/lib/common/config_monitoring_service.cc
pudelkoM: @bocon13 @Yi-Tseng Any thoughts on these approaches?

`STLIsUnique` has O(n^2) complexity but uses no additional memory. `AreUniqueNames` uses a set but is amortized O(n).

Yi-Tseng (dismissed):

stratum/hal/lib/common/config_monitoring_service.cc
Yi-Tseng: I think we can use `AreUniqueNames` since we won't have a huge set of port configs (up to 260 for 65portx10G) and we can also keep code simple.

stratum/hal/lib/common/config_monitoring_service.cc
Yi-Tseng: Maybe "The collection **`contains`** unique names?"

stratum/hal/lib/common/config_monitoring_service.h
```diff
@@ -60,6 +60,10 @@ class ConfigMonitoringService final : public ::gnmi::gNMI::Service {
                                    std::unique_ptr<ChassisConfig> config)
       LOCKS_EXCLUDED(config_lock_);
 
+  // Verifies platform independent properties of the given ChassisConfig proto.
+  // Note that PushChassisConfig calls VerifyChassisConfig at the beginning.
```
Yi-Tseng, line 64: We calls the `VerifyChassisConfig` before calling `PushChassisConfig`. But this comment shows one function calls another function.
How about `Note that VerifyChassisConfig will/must be called before PushChassisConfig`?

Yi-Tseng (dismissed):

stratum/hal/lib/common/config_monitoring_service.h
pudelkoM: It actually is called by the `ConfigMonitoringService::PushChassisConfig`: https://github.com/stratum/stratum/blob/32a71f141f9292f30dc4183c5c9007f876779aee/stratum/hal/lib/common/config_monitoring_service.cc#L113-L125

We also call it in `DoSet`, before handing the config to the `switch_interface_`.

PR #320, opened 2020-09-01
Add support for wildcard query to get node-id

Yi-Tseng:
TODO: Test on HW

### DIFF
--- a/stratum/hal/lib/common/yang_parse_tree.cc
+++ b/stratum/hal/lib/common/yang_parse_tree.cc
@@ -188,7 +188,8 @@ bool YangParseTree::IsWildcard(const std::string& name) const {
 
 ::util::Status YangParseTree::PerformActionForAllNonWildcardNodes(
     const gnmi::Path& path, const gnmi::Path& subpath,
-    const std::function<::util::Status(const TreeNode& leaf)>& action) const {
+    const std::function<::util::Status(const TreeNode& leaf)>& action,
+    bool skip_if_not_exists = false) const {
   const auto* root = root_.FindNodeOrNull(path);
   ::util::Status ret = ::util::OkStatus();
   for (const auto& entry : root->children_) {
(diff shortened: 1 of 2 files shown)


stratum/hal/lib/common/yang_parse_tree.cc
```diff
@@ -188,7 +188,8 @@ bool YangParseTree::IsWildcard(const std::string& name) const {
 
 ::util::Status YangParseTree::PerformActionForAllNonWildcardNodes(
     const gnmi::Path& path, const gnmi::Path& subpath,
-    const std::function<::util::Status(const TreeNode& leaf)>& action) const {
+    const std::function<::util::Status(const TreeNode& leaf)>& action,
+    bool skip_if_not_exists = false) const {
   const auto* root = root_.FindNodeOrNull(path);
   ::util::Status ret = ::util::OkStatus();
   for (const auto& entry : root->children_) {
```
pudelkoM, line 195: Here should be a `nullptr` check, right?

stratum/hal/lib/common/yang_parse_tree.cc
```diff
@@ -199,11 +200,15 @@ ::util::Status YangParseTree::PerformActionForAllNonWildcardNodes(
...
     if (leaf == nullptr) {
-      // Should not happen!
-      ::util::Status status = MAKE_ERROR(ERR_INTERNAL)
-                              << "Found node without "
-                              << subpath.ShortDebugString() << " leaf!";
-      APPEND_STATUS_IF_ERROR(ret, status);
+      // This will happen if the subpath does not exist in the path.
+      // For example, trying to query node-id from all components
+      // but some components do not have node-id leaf.
+      if (!skip_if_not_exists) {
```
pudelkoM, line 206: I would be ok with always dropping the error.

Yi-Tseng: Sounds good, will clean up this one

pudelkoM:
Seems to work without errors now.

pudelkoM (approved):

PR #328, opened 2020-09-16
Report packetout errors to controller

pudelkoM:
A previous proposal added StreamErrors to the stream channel to report asynchronous errors like dropped packetOuts back to the controller. This PR enables this functionality in Stratum.


stratum/hal/lib/common/p4_service.cc
```diff
@@ -536,13 +536,29 @@ ::grpc::Status P4Service::StreamChannel(
...
+          ::p4::v1::StreamMessageResponse resp;
+          auto stream_error = resp.mutable_error();
+          stream_error->set_canonical_code(ERR_PERMISSION_DENIED);
+          stream_error->set_message("Controller is not master.");
+          *stream_error->mutable_packet_out()->mutable_packet_out() =
+              req.packet();
+          stream->Write(resp);  // Best effort.
+          break;
+        }
         // If master, try to transmit the packet. No error reporting.
```
Yi-Tseng, line 550: remove "No error reporting"?

stratum/hal/lib/common/p4_service_test.cc
```diff
@@ -157,6 +157,8 @@ class P4ServiceTest : public ::testing::TestWithParam<OperationMode> {
       metadata_id: 666666
       value: "\x12"
   )";
+  static constexpr char kTestPacketMetadata4[] = R"(
+  )";
```
Yi-Tseng, line 161: can an empty string `""` work?

Yi-Tseng (dismissed):

stratum/hal/lib/common/p4_service_test.cc
pudelkoM: I'm sure it would, but this form keeps the code consistent with the above lines.

stratum/hal/lib/common/p4_service.cc
```diff
@@ -536,13 +536,32 @@ ::grpc::Status P4Service::StreamChannel(
...
+        // If this stream is not the master stream generate a stream error.
+        if (!IsMasterController(node_id, connection_id)) {
+          ::p4::v1::StreamMessageResponse resp;
+          auto stream_error = resp.mutable_error();
+          stream_error->set_canonical_code(
+              ::grpc::StatusCode::PERMISSION_DENIED);
+          stream_error->set_message("Controller is not master.");
+          *stream_error->mutable_packet_out()->mutable_packet_out() =
+              req.packet();
+          stream->Write(resp);  // Best effort.
```
pudelkoM, line 548: I think this code could be a function, like: https://github.com/stratum/stratum/blob/9f5bd2b285badbef11e81eca6c31d4a3c4342843/stratum/hal/lib/common/config_monitoring_service.cc#L337-L348

Yi-Tseng: sounds good. But I am not sure if we can share this code between gNMI and P4Runtime (maybe use the template?)

pudelkoM: It's not that easy, because we have to personalize the error before sending. PacketOut errors include the packet, for example.

Yi-Tseng (approved):

bocon13 (approved):
lgtm

PR #330, opened 2020-09-17
Disable logging to files by default

pudelkoM:
During an unrelated code change I found that the `-logtostderr` flag disables logging to files on disk.
Something we previously though not to be possible (#313).

This PR provides two options to implement this:
- Set a Stratum-wide (inc. cli apps) default before parsing the flags
  - We should pick this if we're certain that we don't care about disk logging in the majority of the cases
- <s>Supply the flag in the start script of the relevant applications
  - This option is better when the flag only makes sense for a few apps</s>

The first option is favored by the team.

In both cases it can be overwritten by the user.


Yi-Tseng (approved):

Yi-Tseng (approved):

stratum/glue/init_google.h
```diff
@@ -14,13 +13,24 @@
...
+  CHECK(!::gflags::SetCommandLineOptionWithMode("logtostderr", "true",
+                                                ::gflags::SET_FLAGS_DEFAULT)
+             .empty());
+  CHECK(!::gflags::SetCommandLineOptionWithMode("colorlogtostderr", "true",
+                                                ::gflags::SET_FLAGS_DEFAULT)
+             .empty());
+  CHECK(!::gflags::SetCommandLineOptionWithMode("stderrthreshold", "0",
+                                                ::gflags::SET_FLAGS_DEFAULT)
+             .empty());
   ParseCommandLineFlags(argc, argv, remove_flags);
```
pudelkoM, line 34: For future reference: we might have to split this up in `ParseCommandLineNonHelpFlags` and `HandleCommandLineHelpFlags`. `SetCommandLineOptionWithMode` would be done in between.

PR #331, opened 2020-09-18
Bump bazel version and change default compiler for CI

### LINKED ISSUE #56: Consider using clang toolchain
https://clang.llvm.org/comparison.html
TL;DR: Clang is much faster and uses far less memory than GCC.

Should consider move to clang since Stratum project is getting bigger and CI may crashed due to some resource limit.
Also, clang tool chain provides `AddressSanitizer` which helps us to debug Stratum when it crashed.
Moreover, the development time can be reduced.

### LINKED ISSUE #210: Explore tools for helping users manage Bazel version
For example:
https://github.com/grpc/grpc/blob/master/tools/bazel
https://github.com/bazelbuild/bazelisk

Yi-Tseng:
This pull request includes the following changes:
 - Use the Bazelisk tool instead of Bazel
 - Adds `.bazelversion` file to ensure the developer is using the correct Bazel version
 - Update the Bazel version to 3.5.0
 - Change the default compiler to clang/clang++ for CI

Also fixes #210 and fixes #56


Dockerfile.build
```diff
@@ -42,11 +43,12 @@ RUN add-apt-repository "$LLVM_REPO_NAME"
...
-RUN wget https://github.com/bazelbuild/bazel/releases/download/$BAZEL_VERSION/bazel-$BAZEL_VERSION-installer-linux-x86_64.sh && \
-    chmod +x bazel-$BAZEL_VERSION-installer-linux-x86_64.sh && \
-    ./bazel-$BAZEL_VERSION-installer-linux-x86_64.sh && \
-    rm -f bazel-$BAZEL_VERSION-installer-linux-x86_64.sh
+ARG USE_BAZEL_VERSION
+ARG BAZELISK_VERSION
+RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v${BAZELISK_VERSION}/bazelisk-linux-amd64 && \
+    mv bazelisk-linux-amd64 /usr/local/bin/bazel && \
+    chmod +x /usr/local/bin/bazel && \
+    bazel version
```
Yi-Tseng, line 51: Use `bazel version` to check the version and download the bazel binary when building the base container image.

pudelkoM (approved):

PR #341, opened 2020-09-25
Make bazel query work

### LINKED ISSUE #209: Get all Bazel targets to "run"
So that queries over entry tree don't fail

Yi-Tseng:
Fix targets which are not building

Fixes #209

### DIFF
--- a/stratum/pipelines/ptf/BUILD
+++ b/stratum/pipelines/ptf/BUILD
@@ -4,7 +4,7 @@
 
 licenses(["notice"])  # Apache v2
 
-load("@build_stack_rules_proto//python:python_grpc_library.bzl", "python_grpc_library")
+# load("@build_stack_rules_proto//python:python_grpc_library.bzl", "python_grpc_library")
 load(":scapy_exec.bzl", "scapy_exec")
 
 exports_files(["LICENSE"])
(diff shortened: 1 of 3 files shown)


stratum/pipelines/ptf/BUILD
```diff
@@ -25,10 +25,10 @@ filegroup(
...
 )
 
-python_grpc_library(
-    name = "ptf_py_lib",
-    deps = ["@com_google_googleapis//google/rpc:code_proto"],
-)
+# python_grpc_library(
+#     name = "ptf_py_lib",
+#     deps = ["@com_google_googleapis//google/rpc:code_proto"],
+# )
```
Yi-Tseng, line 31: Before we generate our own python library for `code.proto`. but now we can use `code_py_proto` from upstream directly.

stratum/pipelines/ptf/ptf_exec.bzl
```diff
@@ -23,8 +23,8 @@ def ptf_exec( name, device, pipeline_name, data=None, skip_test=False, skip_bmv2
         py_args.append(" --skip-bmv2-start")
 
     # Provide directories in the form pypi__<module>_<version>, without the "@" prefix and the "//:pkg" suffix
-    py_args.append(" --protobuf-dir ")
-    py_args.append(grpc_requirement("protobuf")[1:-6])
+    # py_args.append(" --protobuf-dir ")
+    # py_args.append(grpc_requirement("protobuf")[1:-6])
```
Yi-Tseng, line 27: Need to double check

PR #343, opened 2020-09-30
Stratum bf sim mode

pudelkoM:
This PR introduces the OperationMode argument to the tofino code base. This will allow adapting behavior depending on whether Stratum is running in hardware or tofino model.

### DIFF
--- a/stratum/hal/bin/barefoot/main_bfrt.cc
+++ b/stratum/hal/bin/barefoot/main_bfrt.cc
@@ -79,10 +79,20 @@ namespace barefoot {
     LOG(INFO) << "switchd started successfully";
   }
 
-  int device_id(0);
   // TODO(antonin): The SDE expects 0-based device ids, so we instantiate
   // components with "device_id" instead of "node_id".
-  const OperationMode mode = OPERATION_MODE_STANDALONE;
+  int device_id(0);
+
+  auto bf_pal_wrapper = BFPalWrapper::GetSingleton();
+  ASSIGN_OR_RETURN(bool on_sw_model,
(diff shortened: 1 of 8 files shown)


stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -51,7 +53,8 @@ ::util::Status BfrtTableManager::PushForwardingPipelineConfig(
     ASSIGN_OR_RETURN(
         P4Annotation annotation,
         p4_info_manager->GetSwitchStackAnnotations(reg.preamble().name()));
-    if (annotation.register_reset_interval_ms()) {
+    if (annotation.register_reset_interval_ms() &&
+        mode_ == OPERATION_MODE_STANDALONE) {
```
ccascone, line 57: It would be helpful to print a warn log saying that register reset is disabled because of simulation mode

stratum/hal/bin/barefoot/main_bfrt.cc
```diff
@@ -79,13 +79,24 @@ ::util::Status Main(int argc, char* argv[]) {
     LOG(INFO) << "switchd started successfully";
   }
 
-  int device_id(0);
   // TODO(antonin): The SDE expects 0-based device ids, so we instantiate
   // components with "device_id" instead of "node_id".
+  int device_id(0);
+
+  auto bf_pal_wrapper = BFPalWrapper::GetSingleton();
+  ASSIGN_OR_RETURN(bool on_sw_model,
```
Yi-Tseng, line 87: `bool is_sw_model`

PR #348, opened 2020-10-01
Move register clear work to thread and batch all

pudelkoM:


stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -102,6 +133,71 @@ ::util::Status BfrtTableManager::VerifyForwardingPipelineConfig(
...
+                                       min_reset_interval_ms);
+      ::p4::v1::RegisterEntry register_entry;
+      register_entry.set_register_id(reg.preamble().id());
+      register_entry.mutable_data()->set_bitstring(clear_value);
+      register_entry.clear_index();
+      // TODO(max): remove after testing
+      if (mode_ == OPERATION_MODE_SIM) {
+        register_entry.mutable_index()->set_index(0);
+      }
+      ::util::Status status =
```
Yi-Tseng, line 166: I guess you want to append the status from `WriteRegisterEntry` to the outer status(line141)?

stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -102,6 +133,71 @@ ::util::Status BfrtTableManager::VerifyForwardingPipelineConfig(
...
+  CHECK_RETURN_IF_FALSE(session != nullptr) << "Unable to create session.";
+  ::util::Status status = ::util::OkStatus();
+  for (;;) {
+    auto t1 = absl::Now();
+    uint64 min_reset_interval_ms = 10000;
+    RETURN_IF_BFRT_ERROR(session->beginBatch());
+    for (const auto& reg : registers) {
+      P4Annotation annotation;
+      {
+        absl::ReaderMutexLock l(&lock_);
```
Yi-Tseng, line 149: Why do we need this lock?

stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -48,6 +58,26 @@ ::util::Status BfrtTableManager::PushForwardingPipelineConfig(
...
+      data->registers.push_back(reg);
+    }
+    int ret = pthread_create(&register_reset_thread_, nullptr,
+                             &BfrtTableManager::RegisterClearThreadFunc, data);
+    if (ret != 0) {
+      RETURN_ERROR(ERR_INTERNAL) << "Failed to spawn register reset thread.";
+    }
+  } else {
+    LOG(FATAL)
+        << "Multiple pipeline pushes are not supported with register resets.";
```
pudelkoM, line 77: Fix this before we merge this.

Maybe use the timer daemon with a single event instead.

Yi-Tseng: I think that's the reason we need the lock when we push the pipeline

PR #349, opened 2020-10-01
Call gRPC server Shutdown() from a separate thread

bocon13:
If the thread that called gRPC Server Wait() is the one that handles the
Ctrl-C, then the thread will try to re-enter the gRPC Server mutex.

This seems to work fine with pthread_mutex (behavior is undefined), but
results in deadlock in Abseil's mutex, which becomes the default in
gRPC 1.32.0.

There should be no harm in doing this for gRPC < 1.32.0.


bocon13:
Update to 1.33.2 and check that bfrt_server branch keeps running

bocon13:
Starting with gRPC 1.32, we see the following when killing stratum with Ctrl-C...

```
^CI20201118 03:38:20.235855 2164777 hal.cc:244] Received signal: Interrupt
[mutex.cc : 1358] RAW: Potential Mutex deadlock: 
	@ 0x55fe4bef7fc3 absl::lts_2020_09_23::DebugOnlyDeadlockCheck()
	@ 0x55fe4bef8290 absl::lts_2020_09_23::Mutex::Lock()
	@ 0x55fe4bb0a711 gpr_mu_lock
	@ 0x55fe4b7ec7ac grpc::CoreCodegen::gpr_mu_lock()
	@ 0x55fe4b230426 grpc::internal::MutexLock::MutexLock()
	@ 0x55fe4b805cf2 grpc::Server::ShutdownInternal()
	@ 0x55fe4b1e74f3 grpc::ServerInterface::Shutdown<>()
	@ 0x55fe4b1e2098 stratum::hal::Hal::HandleSignal()
	@ 0x55fe4b1de66d stratum::hal::(anonymous namespace)::SignalRcvCallback()
	@ 0x7f0326f4f210 
	@ 0x55fe4befeaef absl::lts_2020_09_23::synchronization_internal::Waiter::Wait()
	@ 0x55fe4befe813 AbslInternalPerThreadSemWait
	@ 0x55fe4befd0e2 absl::lts_2020_09_23::synchronization_internal::PerThreadSem::Wait()
	@ 0x55fe4befd578 absl::lts_2020_09_23::Mutex::DecrementSynchSem()
	@ 0x55fe4befc1fb absl::lts_2020_09_23::CondVar::WaitCommon()
	@ 0x55fe4befc3af absl::lts_2020_09_23::CondVar::Wait()
	@ 0x55fe4bb0a7fa gpr_cv_wait
	@ 0x55fe4b7ec860 grpc::CoreCodegen::gpr_cv_wait()
	@ 0x55fe4b8078c4 grpc::internal::CondVar::Wait()
	@ 0x55fe4b80785a grpc::internal::CondVar::Wait()
	@ 0x55fe4b8061d8 grpc::Server::Wait()
	@ 0x55fe4b1e1e53 stratum::hal::Hal::Run()
	@ 0x55fe4b1dcc78 stratum::hal::dummy_switch::Main()
	@ 0x55fe4b1dcf1c main
	@ 0x7f0326f300b3 __libc_start_main

[mutex.cc : 1368] RAW: Acquiring 0x55fe4e3438e8    Mutexes held:  0x55fe4e3438e8
[mutex.cc : 1370] RAW: Cycle: 
[mutex.cc : 1384] RAW: mutex@0x55fe4e3438e8 stack: 
	@ 0x55fe4bef7fc3 absl::lts_2020_09_23::DebugOnlyDeadlockCheck()
	@ 0x55fe4bef8290 absl::lts_2020_09_23::Mutex::Lock()
	@ 0x55fe4bb0a711 gpr_mu_lock
	@ 0x55fe4b7ec7ac grpc::CoreCodegen::gpr_mu_lock()
	@ 0x55fe4b230426 grpc::internal::MutexLock::MutexLock()
	@ 0x55fe4b805cf2 grpc::Server::ShutdownInternal()
	@ 0x55fe4b1e74f3 grpc::ServerInterface::Shutdown<>()
	@ 0x55fe4b1e2098 stratum::hal::Hal::HandleSignal()
	@ 0x55fe4b1de66d stratum::hal::(anonymous namespace)::SignalRcvCallback()
	@ 0x7f0326f4f210 
	@ 0x55fe4befeaef absl::lts_2020_09_23::synchronization_internal::Waiter::Wait()
	@ 0x55fe4befe813 AbslInternalPerThreadSemWait
	@ 0x55fe4befd0e2 absl::lts_2020_09_23::synchronization_internal::PerThreadSem::Wait()
	@ 0x55fe4befd578 absl::lts_2020_09_23::Mutex::DecrementSynchSem()
	@ 0x55fe4befc1fb absl::lts_2020_09_23::CondVar::WaitCommon()
	@ 0x55fe4befc3af absl::lts_2020_09_23::CondVar::Wait()
	@ 0x55fe4bb0a7fa gpr_cv_wait
	@ 0x55fe4b7ec860 grpc::CoreCodegen::gpr_cv_wait()
	@ 0x55fe4b8078c4 grpc::internal::CondVar::Wait()
	@ 0x55fe4b80785a grpc::internal::CondVar::Wait()
	@ 0x55fe4b8061d8 grpc::Server::Wait()
	@ 0x55fe4b1e1e53 stratum::hal::Hal::Run()
	@ 0x55fe4b1dcc78 stratum::hal::dummy_switch::Main()
	@ 0x55fe4b1dcf1c main
	@ 0x7f0326f300b3 __libc_start_main

[mutex.cc : 1389] RAW: dying due to potential deadlock
*** Aborted at 1605670700 (unix time) try "date -d @1605670700" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGABRT (@0x3ec00210829) received by PID 2164777 (TID 0x7f0326f04800) from PID 2164777; stack trace: ***
    @     0x7f032712b3c0 (unknown)
    @     0x7f0326f4f18b gsignal
    @     0x7f0326f2e859 abort
    @     0x55fe4bf42548 (anonymous namespace)::RawLogVA()
    @     0x55fe4bf4264b absl::lts_2020_09_23::raw_logging_internal::RawLog()
    @     0x55fe4bef7ef4 absl::lts_2020_09_23::DeadlockCheck()
    @     0x55fe4bef7fc3 absl::lts_2020_09_23::DebugOnlyDeadlockCheck()
    @     0x55fe4bef8290 absl::lts_2020_09_23::Mutex::Lock()
    @     0x55fe4bb0a711 gpr_mu_lock
    @     0x55fe4b7ec7ac grpc::CoreCodegen::gpr_mu_lock()
    @     0x55fe4b230426 grpc::internal::MutexLock::MutexLock()
    @     0x55fe4b805cf2 grpc::Server::ShutdownInternal()
    @     0x55fe4b1e74f3 grpc::ServerInterface::Shutdown<>()
    @     0x55fe4b1e2098 stratum::hal::Hal::HandleSignal()
    @     0x55fe4b1de66d stratum::hal::(anonymous namespace)::SignalRcvCallback()
    @     0x7f0326f4f210 (unknown)
    @     0x7f032702489b syscall
    @     0x55fe4befedac absl::lts_2020_09_23::synchronization_internal::Futex::WaitUntil()
    @     0x55fe4befeaef absl::lts_2020_09_23::synchronization_internal::Waiter::Wait()
    @     0x55fe4befe813 AbslInternalPerThreadSemWait
    @     0x55fe4befd0e2 absl::lts_2020_09_23::synchronization_internal::PerThreadSem::Wait()
    @     0x55fe4befd578 absl::lts_2020_09_23::Mutex::DecrementSynchSem()
    @     0x55fe4befc1fb absl::lts_2020_09_23::CondVar::WaitCommon()
    @     0x55fe4befc3af absl::lts_2020_09_23::CondVar::Wait()
    @     0x55fe4bb0a7fa gpr_cv_wait
    @     0x55fe4b7ec860 grpc::CoreCodegen::gpr_cv_wait()
    @     0x55fe4b8078c4 grpc::internal::CondVar::Wait()
    @     0x55fe4b80785a grpc::internal::CondVar::Wait()
    @     0x55fe4b8061d8 grpc::Server::Wait()
    @     0x55fe4b1e1e53 stratum::hal::Hal::Run()
    @     0x55fe4b1dcc78 stratum::hal::dummy_switch::Main()
    @     0x55fe4b1dcf1c main
Aborted (core dumped)
```

pudelkoM:
What's really interesting: Both the blocking gRPC server `Wait` and the signal handler are on the same stack and run in the same thread:

```
    @     0x55fe4b1e74f3 grpc::ServerInterface::Shutdown<>()
    @     0x55fe4b1e2098 stratum::hal::Hal::HandleSignal()
    @     0x55fe4b1de66d stratum::hal::(anonymous namespace)::SignalRcvCallback()
...
    @     0x55fe4b8061d8 grpc::Server::Wait()
```

Easy to see why we get a mutex cycle this way.

Maybe use an alternate stack for the signal handler?

bocon13:
Reading the docs a bit more closely, it looks like the gRPC server shutdown is supposed to be called from a separate thread.  Spawning a new thread for the task is a bit clunky, but solves the deadlock warning.

```
E20201118 06:30:20.073416 2294566 hal.cc:223] Stratum external facing services are listening to 0.0.0.0:28000, 0.0.0.0:9339, 0.0.0.0:9559, localhost:28000...
^CI20201118 06:30:23.066975 2294566 hal.cc:244] Received signal: Interrupt
I20201118 06:30:23.068991 2294566 hal.cc:165] Shutting down HAL...
I20201118 06:30:23.069049 2294566 dummy_switch.cc:217] UnregisterEventNotifyWriter
I20201118 06:30:23.069128 2294566 dummy_switch.cc:104] Shutdown
I20201118 06:30:23.069320 2294566 phaldb_service.cc:56] PhalDbService shutdown completed successfully.
I20201118 06:30:23.070062 2294566 dummy_chassis_mgr.cc:41] Shutdown
I20201118 06:30:23.070879 2294566 dummy_box.cc:224] Shutting down the DummyBox.
I20201118 06:30:23.071316 2294566 main.cc:47] See you later!
```

stratum/hal/lib/common/hal.cc
```diff
@@ -247,7 +247,23 @@ void Hal::HandleSignal(int value) {
   // with no deadline will block forever, as it waits for all the active RPCs
   // to finish. To fix this, we give a deadline set to "now" so the call returns
   // immediately.
-  external_server_->Shutdown(std::chrono::system_clock::now());
+  // Per the grpc::Server docs, some other thread must call Shutdown() for Wait()
+  // to ever return. So, we try to call shutdown it a separate thread.
+  pthread_t shutdown_tid;
+  int ret = pthread_create(&shutdown_tid, nullptr,
```
pudelkoM: The list of functions safe to call from a signal handler is quite short (`man signal-safety`). `pthread_create` is not one of them :(

stratum/hal/lib/common/hal.cc
```diff
@@ -247,7 +243,25 @@ void Hal::HandleSignal(int value) {
   // with no deadline will block forever, as it waits for all the active RPCs
   // to finish. To fix this, we give a deadline set to "now" so the call returns
   // immediately.
-  external_server_->Shutdown(std::chrono::system_clock::now());
```
bocon13, line 250: @pudelkoM is this what you were thinking?

pudelkoM:
Can we pinpoint this change to a specific gRPC version yet? I wonder what changed upstream.

bocon13:
I started noticing it in gRPC 1.32.0. Could also be a change in the abseil version?

pudelkoM:
> change in the abseil version?

I didn't even notice. The PR title does not say it, nor the description. Why do you do this to me 😂

But yes, that's seems more likely.

bocon13:
> > change in the abseil version?
> 
> I didn't even notice. The PR title does not say it, nor the description. Why do you do this to me 😂
> 
> But yes, that's seems more likely.

But the first commit message says it :) I made the change after creating the PR hoping it would solve the problem.
I think the issue is still present in the older version of absl with gRPC 1.32.0. I took a look at the ::grpc::Server implementation between 1.30 and 1.32, and it wasn't obvious to me that there was a change that would be causing this.

```
E20201118 07:20:15.012745  2770 hal.cc:223] Stratum external facing services are listening to 0.0.0.0:28000, 0.0.0.0:9339, 0.0.0.0:9559, localhost:28000...
^CI20201118 07:20:16.093868  2770 hal.cc:244] Received signal: Interrupt
[mutex.cc : 1355] RAW: Potential Mutex deadlock: 
	@ 0x564d925bde8d absl::lts_2020_02_25::DebugOnlyDeadlockCheck() <--- note the older version
	@ 0x564d925be174 absl::lts_2020_02_25::Mutex::Lock()
	@ 0x564d921dd725 gpr_mu_lock
	@ 0x564d91ec662a grpc::CoreCodegen::gpr_mu_lock()
	@ 0x564d918fd34a grpc::internal::MutexLock::MutexLock()
	@ 0x564d91edfb70 grpc_impl::Server::ShutdownInternal()
	@ 0x564d918b3b4b grpc::ServerInterface::Shutdown<>()
	@ 0x564d918ae4d8 stratum::hal::Hal::HandleSignal()
	@ 0x564d918aaaad stratum::hal::(anonymous namespace)::SignalRcvCallback()
	@ 0x7fcc29355210 
	@ 0x564d925c44bb absl::lts_2020_02_25::synchronization_internal::Waiter::Wait()
	@ 0x564d925c41df AbslInternalPerThreadSemWait
	@ 0x564d925c2af8 absl::lts_2020_02_25::synchronization_internal::PerThreadSem::Wait()
	@ 0x564d925c2eda absl::lts_2020_02_25::Mutex::DecrementSynchSem()
	@ 0x564d925c1f6d absl::lts_2020_02_25::CondVar::WaitCommon()
	@ 0x564d925c2121 absl::lts_2020_02_25::CondVar::Wait()
	@ 0x564d921dd80e gpr_cv_wait
	@ 0x564d91ec66de grpc::CoreCodegen::gpr_cv_wait()
	@ 0x564d91ee165c grpc::internal::CondVar::Wait()
	@ 0x564d91ee15f2 grpc::internal::CondVar::Wait()
	@ 0x564d91ee0056 grpc_impl::Server::Wait()
	@ 0x564d918ae293 stratum::hal::Hal::Run()
	@ 0x564d918a90b8 stratum::hal::dummy_switch::Main()
	@ 0x564d918a935c main
	@ 0x7fcc293360b3 __libc_start_main
```

bocon13:
I found the change that causes the error: https://github.com/grpc/grpc/pull/23372

In gRPC 1.32.0, they switched the default from `pthread_mutex_t` and `pthread_cond_t` to `absl::Mutex` and `absl::CondVar` respectively. The pthread mutex is a PTHREAD_MUTEX_DEFAULT, so the behavior of re-entering the lock in the same thread (to handle the signal) is undefined; for our platform, it appears to work just fine. The Abseil mutex on the other hand does not like to be re-entered.

bocon13:
It seems like our success before was relying on undefined platform behavior. Abseil now makes that more concrete -- deadlock. Given that we can't really inspect the state of the mutex because it's owned by gRPC, I think the right thing to do is to call the Shutdown() method from a separate caller; otherwise, kill the process to avoid deadlocking.

stratum/hal/lib/common/hal.cc
bocon13: I guess this makes sense... this probably means that we need a fix inside of gRPC?

pudelkoM:
I tried moving the signal handler to a different stack with `sigstack`, but to no avail:

```
I20201118 11:15:50.399549  9728 hal.cc:242] Received signal: Interrupt
[mutex.cc : 1356] RAW: Potential Mutex deadlock:
	@ 0x1721e3f absl::lts_2020_02_25::DebugOnlyDeadlockCheck()
	@ 0x172186f absl::lts_2020_02_25::Mutex::Lock()
	@ 0x10fc1ed gpr_mu_lock
	@ 0xd5e629 grpc::CoreCodegen::gpr_mu_lock()
	@ 0x4fa771 grpc::internal::MutexLock::MutexLock()
	@ 0xd78878 grpc::Server::ShutdownInternal()
	@ 0x4a159b grpc::ServerInterface::Shutdown<>()
	@ 0x49ecec stratum::hal::Hal::HandleSignal()
	@ 0x49fef0 stratum::hal::(anonymous namespace)::SignalRcvCallback()
	@ 0x7ff43822b0e0 __restore_rt

[mutex.cc : 1366] RAW: Acquiring 0x3e99c08    Mutexes held:  0x3e99c08
[mutex.cc : 1367] RAW: Cycle:
[mutex.cc : 1381] RAW: mutex@0x3e99c08 stack:
	@ 0x1721e3f absl::lts_2020_02_25::DebugOnlyDeadlockCheck()
	@ 0x172186f absl::lts_2020_02_25::Mutex::Lock()
	@ 0x10fc1ed gpr_mu_lock
	@ 0xd5e629 grpc::CoreCodegen::gpr_mu_lock()
	@ 0x4fa771 grpc::internal::MutexLock::MutexLock()
	@ 0xd78878 grpc::Server::ShutdownInternal()
	@ 0x4a159b grpc::ServerInterface::Shutdown<>()
	@ 0x49ecec stratum::hal::Hal::HandleSignal()
	@ 0x49fef0 stratum::hal::(anonymous namespace)::SignalRcvCallback()
	@ 0x7ff43822b0e0 __restore_rt

[mutex.cc : 1386] RAW: dying due to potential deadlock
```

pudelkoM:
Disabling the deadlock detection still causes the crash:

```cpp
  absl::SetMutexDeadlockDetectionMode(absl::OnDeadlockCycle::kIgnore);
```

```
I20201118 19:06:38.761687  8526 hal.cc:242] Received signal: Interrupt
[mutex.cc : 2043] RAW: Check waitp == nullptr || waitp->thread->waitp == nullptr || waitp->thread->suppress_fatal_errors failed: detected illegal recursion into Mutex code
*** Aborted at 1605726398 (unix time) try "date -d @1605726398" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGABRT (@0x214e) received by PID 8526 (TID 0x7f8befa30800) from PID 8526; stack trace: ***
    @     0x7f8bbaf7f0e0 (unknown)
    @     0x7f8bba9eafff gsignal
    @     0x7f8bba9ec42a abort
    @          0x177071b (anonymous namespace)::RawLogVA()
    @          0x1770513 absl::lts_2020_02_25::raw_logging_internal::RawLog()
    @          0x1723fde absl::lts_2020_02_25::Mutex::UnlockSlow()
    @          0x172e06b absl::lts_2020_02_25::CondVar::WaitCommon()
    @          0x172e2da absl::lts_2020_02_25::CondVar::Wait()
    @          0x10fc321 gpr_cv_wait
    @           0xd729ff grpc_core::CondVar::Wait()
    @           0xd724f5 grpc_core::CondVar::Wait()
    @           0xd9cc43 grpc::ThreadManager::Wait()
    @           0xd7d30f grpc::Server::SyncRequestThreadManager::Wait()
    @           0xd78b6c grpc::Server::ShutdownInternal()
    @           0x4a159b grpc::ServerInterface::Shutdown<>()
    @           0x49ecec stratum::hal::Hal::HandleSignal()
    @           0x49fef0 stratum::hal::(anonymous namespace)::SignalRcvCallback()
    @     0x7f8bbaf7f0e0 (unknown)
    @     0x7f8bbaa9c469 syscall
    @          0x1735bb4 absl::lts_2020_02_25::synchronization_internal::Futex::WaitUntil()
    @          0x1735754 absl::lts_2020_02_25::synchronization_internal::Waiter::Wait()
    @          0x1734f19 AbslInternalPerThreadSemWait
    @          0x17322dd absl::lts_2020_02_25::synchronization_internal::PerThreadSem::Wait()
    @          0x1731c23 absl::lts_2020_02_25::Mutex::DecrementSynchSem()
    @          0x172e0b0 absl::lts_2020_02_25::CondVar::WaitCommon()
    @          0x172e2da absl::lts_2020_02_25::CondVar::Wait()
    @          0x10fc321 gpr_cv_wait
    @           0xd5e6d1 grpc::CoreCodegen::gpr_cv_wait()
    @           0xd86ea0 grpc::internal::CondVar::Wait()
    @           0xd7bf86 grpc::internal::CondVar::Wait()
    @           0xd78e40 grpc::Server::Wait()
    @           0x49e41d stratum::hal::Hal::Run()
```

pudelkoM:
I had success with blocking signal delivery to the main thead (that `wait`s on the gRPC server):

```
::util::Status Hal::RegisterSignalHandlers() {
  ...
  sigset_t set;
  sigemptyset(&set);
  sigaddset(&set, SIGINT);
  sigaddset(&set, SIGTERM);
  sigaddset(&set, SIGUSR2);
  if (pthread_sigmask(SIG_BLOCK, &set, NULL) != 0) {
    RETURN_ERROR(ERR_INVALID_PARAM) << "pthread_sigmask";
  }
...
```

The signal then got delivered to another thread:
```
(lldb) thread list
Process 26143 stopped
  thread #1: tid = 26143, 0x00007f7b92db5469 libc.so.6`syscall + 25, name = 'stratum_bf'
* thread #2: tid = 26180, 0x000000000049ec80 stratum_bf`stratum::hal::Hal::HandleSignal(int), name = 'bf_timer_src', stop reason = breakpoint 1.1
```

and shutdown completed without errors:

```
I20201118 20:27:02.726862 26180 hal.cc:242] Received signal: Interrupt
I20201118 20:27:02.728950 26143 hal.cc:163] Shutting down HAL...
I20201118 20:27:02.730391 26143 timer_daemon.cc:109] The timer daemon has been stopped.
I20201118 20:27:02.730439 26143 main.cc:169] See you later!
```

pudelkoM:
Summary from call:

Calling `Shutdown` from signal handler is always bad, no matter which thread it runs on.
Try signal handling with semaphores. Handler calls `sem_post` (explicitly async-safe), other thread `sem_wait`s and calls shutdown.

bocon13:
Semaphore based implementation is complete.

stratum/hal/lib/common/hal.cc
```diff
@@ -337,12 +346,26 @@ ::util::Status Hal::InitializeServer() {
...
 ::util::Status Hal::RegisterSignalHandlers() {
+  // Initialize the gRPC server shutdown semaphore
+  CHECK_ERR(sem_init(&grpc_shutdown_sem, 0, 0));
+
+  // Start the gRPC server shutdown thread
+  {
+    pthread_t t;
+    pthread_attr_t attr;
+    CHECK_ERR(pthread_attr_init(&attr));
+    CHECK_ERR(pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED));
```
pudelkoM, line 357: Why not join in `Teardown`?

bocon13: Done.

stratum/hal/lib/common/BUILD
```diff
@@ -539,6 +539,7 @@ stratum_cc_test(
         "@com_github_google_glog//:glog",
         "@com_google_googletest//:gtest",
         "@com_google_absl//absl/strings",
+	"@com_google_absl//absl/time",
```
bocon13, line 542:
```suggestion
        "@com_google_absl//absl/time",
```

stratum/hal/lib/common/hal.cc
```diff
@@ -163,7 +187,33 @@ ::util::Status Hal::Teardown() {
   // of warmboot shutdown, the stack is first freezed by calling an RPC in
   // AdminService, which itself calls Freeze() method in SwitchInterface class.
   LOG(INFO) << "Shutting down HAL...";
+
   ::util::Status status = ::util::OkStatus();
+
```
bocon13, line 192: Consider moving the following to a separate function

stratum/hal/lib/common/hal.cc
```diff
@@ -235,19 +285,37 @@ ::util::Status Hal::Run() {
     }
   }
```
bocon13, line 287: Consider moving the following to a separate function

stratum/hal/lib/common/hal.cc
```diff
@@ -163,7 +187,33 @@ ::util::Status Hal::Teardown() {
...
+      APPEND_STATUS_IF_ERROR(status, error);
+    }
+    if (value == PTHREAD_CANCELED) {
+      LOG(ERROR) << "External server shutdown thread was canceled. This is "
+                    "okay in a test, but should otherwise be avoided.";
+    } else if (value != nullptr) {
+      ::util::Status error =
+          MAKE_ERROR(ERR_INTERNAL)
+          << "External server shutdown thread returned unexpected value.";
+      APPEND_STATUS_IF_ERROR(status, error);
```
bocon13, line 210: This branch is probably not necessary as nullptr is literally the only thing we return.

stratum/hal/lib/common/hal.cc
```diff
@@ -47,12 +50,36 @@ namespace hal {
 
 namespace {
 
+// Semaphore that is unlocked in the shutdown signal handler, which signals
+// a separate thread to shutdown the gRPC service. A semaphore is used here
```
bocon13, line 54:
```suggestion
// a separate thread to shutdown the external server. A semaphore is used here
```

stratum/hal/lib/common/hal.cc
```diff
@@ -47,12 +50,36 @@ namespace hal {
...
+// because sem_post is a async-signal-safe function and gRPC server
```
bocon13, line 55:
```suggestion
// because sem_post is a async-signal-safe function and external server
```

stratum/hal/lib/common/hal.cc
```diff
@@ -47,12 +50,36 @@ namespace hal {
...
+
+// Wait on the shutdown semaphore, then shutdown the external server.
+void* ExternalServerShutdownThread(void*) {
+  // Verify that shutdown semaphore is working by locking it.
+  CHECK_ERR(sem_trywait(&external_server_shutdown_sem));
+
+  // Wait...
+  while (sem_wait(&external_server_shutdown_sem) == -1) {
+    if (errno != EINTR) {
+      LOG(ERROR) << "Failed to wait gRPC server shutdown semaphore. "
```
bocon13, line 74:
```suggestion
      LOG(ERROR) << "Failed to wait external server shutdown semaphore. "
```

stratum/hal/lib/common/hal.cc
```diff
@@ -322,13 +388,16 @@ ::util::Status Hal::InitializeServer() {
 #undef CHECK_IS_NULL  // should not be used in any other method.
 
 ::util::Status Hal::RegisterSignalHandlers() {
+  // Initialize the gRPC server shutdown semaphore
```
bocon13, line 391:
```suggestion
  // Initialize the external server shutdown semaphore
```

stratum/hal/lib/common/hal.cc
```diff
@@ -339,9 +408,16 @@ ::util::Status Hal::RegisterSignalHandlers() {
...
-    signal(e.first, e.second);
+    sighandler_t h = signal(e.first, e.second);
+    if (h == SIG_ERR) {
+      LOG(ERROR) << "Failed to register old signal " << strsignal(e.first);
+    }
   }
   old_signal_handlers_.clear();
+  LOG(INFO) << "Unregistered Hal signal handlers.";
+
+  // Destroy the gRPC server shutdown semaphore as it is no longer needed
```
bocon13, line 419:
```suggestion
  // Destroy the external server shutdown semaphore as it is no longer needed
```

stratum/hal/lib/common/hal.cc
```diff
@@ -235,19 +285,37 @@ ::util::Status Hal::Run() {
     }
   }
 
+  // Clear any pervious interrupts by locking semaphore
```
bocon13, line 288:
```suggestion
  // Clear any pervious interrupts by locking semaphore
```
```suggestion
  // Clear any previous interrupts by locking semaphore
```

stratum/hal/lib/common/hal.cc
```diff
@@ -235,19 +285,37 @@ ::util::Status Hal::Run() {
...
   // NOTE: Seems like if there is an active stream Read(), calling Shutdown()
   // with no deadline will block forever, as it waits for all the active RPCs
   // to finish. To fix this, we give a deadline set to "now" so the call returns
   // immediately.
-  external_server_->Shutdown(std::chrono::system_clock::now());
+  if (external_server_ != nullptr) {
+    external_server_->Shutdown(std::chrono::system_clock::now());
+    LOG(INFO) << "External server shutdown complete.";
+  } else {
+    LOG(ERROR) << "Failed to shutdown external server (nullptr).";
```
bocon13, line 317:
```suggestion
    LOG(ERROR) << "Failed to shutdown nonexistent external server.";
```

stratum/hal/lib/common/hal.cc
```diff
@@ -235,19 +285,37 @@ ::util::Status Hal::Run() {
     }
   }
 
+  // Clear any previous interrupts by locking semaphore
+  while (sem_trywait(&external_server_shutdown_sem) == 0)
+    ;
+  PCHECK(errno == EAGAIN);  // semaphore is locked
```
bocon13, line 291: Not actually sure that we want to do this. A client might have killed the process before we started, but it does replicate the previous behavior of not killing the server if the signal comes before we start the gRPC server.

stratum/hal/lib/common/hal.cc
```diff
@@ -235,19 +285,37 @@ ::util::Status Hal::Run() {
     }
   }
 
+  // Clear any previous interrupts by locking semaphore
+  while (sem_trywait(&external_server_shutdown_sem) == 0)
+    ;
```
bocon13, line 290:
```suggestion
  while (sem_trywait(&external_server_shutdown_sem) == 0) {}
```

stratum/hal/lib/common/hal.h
```diff
@@ -146,6 +146,9 @@ class Hal final {
   // serviced by ConfigMonitoringService and P4Service. Owned by the class.
   std::unique_ptr<::grpc::Server> external_server_;
 
+  // Thread that waits to shutdown the external server. Owned by the class.
+  pthread_t external_server_shutdown_thread_;
```
pudelkoM, line 150: Variable members come before pointers.

bocon13:
New test failure in the coverage job, but it passes in the unit test run:

```
INFO: Build completed, 1 test FAILED, 4140 total actions
//stratum/hal/lib/common:hal_test                                        FAILED in 6.6s
  /tmp/bazel-cache/output-root/f85b6fb5740e6e8c7efea142eec4b6e8/execroot/com_github_stratum_stratum/bazel-out/k8-fastbuild/testlogs/stratum/hal/lib/common/hal_test/test.log
```

stratum/hal/lib/common/hal.cc
pudelkoM: I feel the same about the other checks too.

Instead of trying to expect everything, focus on the single one implemented path. Any deviation is an error.

What is the user supposed to do with a log message like this: `External server shutdown thread was canceled. This is okay in a test, but should otherwise be avoided.` ?

stratum/hal/lib/common/hal_test.cc
```diff
@@ -64,6 +67,12 @@ class FakeProcmonService final : public procmon::ProcmonService::Service {
 };
 
 class HalTest : public ::testing::Test {
+ public:
+  void ShutdownExternalServerAndThread() {
+    CHECK_ERR(pthread_cancel(hal_->external_server_shutdown_thread_));
```
pudelkoM, line 72: `pthread_cancel` for thread management is an anti pattern. Also does not work with C++11 threads.

pudelkoM:
Fixed by #652
We still need to do the gRPC update.

PR #351, opened 2020-10-02
Cleanup register resets

pudelkoM:
This PR moves the register reset annotations behind the `incompatible_enable_register_reset_annotations` flag.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager.cc
@@ -22,6 +22,8 @@ DEFINE_uint32(
     bfrt_table_sync_timeout_ms,
     stratum::hal::barefoot::kDefaultSyncTimeout / absl::Milliseconds(1),
     "The timeout for table sync operation like counters and registers.");
+DEFINE_bool(incompatible_enable_register_reset_annotations, false,
+            "Enables handling of annotions to reset registers.");
 
 namespace stratum {
 namespace hal {
@@ -31,7 +33,6 @@ BfrtTableManager::BfrtTableManager(OperationMode mode,
                                    const BfrtIdMapper* bfrt_id_mapper)
     : mode_(mode),
       register_timer_descriptors_(),
-      register_reset_thread_(0),
       bfrt_info_(nullptr),
       p4_info_manager_(nullptr),
       bfrt_id_mapper_(ABSL_DIE_IF_NULL(bfrt_id_mapper)) {}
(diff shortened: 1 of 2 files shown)


ccascone:
@Yi-Tseng and I tried https://github.com/stratum/stratum/pull/348/commits/5dd8958df23f2c546fe8f5723e3bb28f44f9eb75 on HW and the register reset didn't seem to work. When generating traffic matching a watchlist rule we didn't see INT reports being generated every second. We then deployed the stratum-bfrt image based on the previous commit https://github.com/stratum/stratum/pull/348/commits/e456e5d23dffd3167957ee263e7041dfd6b803fa and it worked. So we believe that changes introduced in https://github.com/stratum/stratum/pull/348/commits/5dd8958df23f2c546fe8f5723e3bb28f44f9eb75 cause the reset logic to stop working.

ccascone:
I confirm the issue goes away when deploying an image based on this PR. The register reset logic works as expected when passing the `incompatible_enable_register_reset_annotations` flag.

stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -133,69 +70,89 @@ ::util::Status BfrtTableManager::VerifyForwardingPipelineConfig(
...
+  if (intervals_ms.empty()) {
+    return ::util::OkStatus();
+  }
+  std::sort(intervals_ms.begin(), intervals_ms.end());
+  auto last = std::unique(intervals_ms.begin(), intervals_ms.end());
+  intervals_ms.erase(last, intervals_ms.end());
+  if (intervals_ms.size() != 1) {
+    RETURN_ERROR(ERR_INVALID_PARAM)
+        << "Inconsistent register reset intervals are not supported.";
+  }
```
Yi-Tseng, line 110: This logic makes sense to me.
Was wondering if we can simplify to this:

```cpp
auto last = std::unique(intervals_ms.begin(), intervals_ms.end());
if (last != (intervals_ms.begin() + 1)) {
  //return error
}
```
```

Yi-Tseng (approved):

stratum/hal/lib/barefoot/bfrt_table_manager.cc
Yi-Tseng: But we can merge this patch now, I don't really find a big issue.

pudelkoM: I really like you idea, but have two concerns:

- I don't want to test this change again
- It makes the code in more unreadable and harder to understand

PR #356, opened 2020-10-07
Clean up glue/gtl files

pudelkoM:
Some of these files where imported from other projects before we adopted a style guide. This PR brings them up to the same standard by applying our formatting tools.


Yi-Tseng (approved):

bocon13 (approved):
lgtm

PR #357, opened 2020-10-08
Cleanup dummy switch

Yi-Tseng:
- Rename namespace "stratum::hal::dummy_switch" to
   "stratum::hal::dummy"
 - Replace Python dummy cli with C++ one.
 - Replace port counter sim to C++


stratum/hal/bin/dummy/dummy_cli.cc
```diff
@@ -0,0 +1,276 @@
...
+  port_qos_counters->set_out_dropped_pkts(strtoull(argv[8], nullptr, 10));
+
+  return request;
+}
+
+const absl::flat_hash_set<std::string> node_port_states = {
+    "oper_status",          "admin_status",          "mac_address",
+    "port_speed",           "negotiated_port_speed", "lacp_router_mac",
+    "lacp_system_priority", "port_counters",         "forwarding_viability",
+    "health_indicator"};
```
pudelkoM, line 149: If you add a `,` after the last item, the bracket will be on the next line.

PR #365, opened 2020-10-20
Add support for BFRuntime gRPC server in stratum_bfrt

bocon13:
Depends on #373

Merge that first into master, then merge master into bfrt, and then merge this patch


bocon13:
superseded by #389

PR #368, opened 2020-10-21
Remove xz compression from supported tar formats

pudelkoM:
This PR removes xz compression from the possible tar formats of the BF pipeline.
xz would need additional dependencies inside the build docker image. Considering tar will go away soon, we have decided to remove xz instead. No documentation updates needed, as xz was never mentioned anywhere.


bocon13 (approved):
lgtm

PR #376, opened 2020-10-27
Decouple SDN and SDK port IDs for Barefoot targets

bocon13:
This patch decouples the ChassisConfig SingletonPort ID number from the BF SDK port number, which were previously required to match. The ID number is intended to be set by the SDN controller or generated from a network management subsystem. SDK port numbers are now read from the SDK when the ChassisConfig is pushed. With this dependency broken, users can now set ID numbers however they would like (but perhaps to little effect).

Ports are accessed via their ChassisConfig SingletonPort name in gNMI, and by the BF SDK port number in P4Runtime.

For P4Runtime, users are still expected to use the SDK port number which can be found in gNMI under:
`/interfaces/interface/state/ifindex`

**TODO**
- [x] Unit test for Chassis Configs including failures for known/implemented validation -- #412
- [x] Answer multi-device support question (do any users require more than 1?)
- [x] Update `PortIdFromPortKeyGet` to return `::util::StatusOr<uint32>` -- #402
- [x] Resolve if/when/how we want to chain/collect errors -- #405
- [x] Use correct error code in grouped errors (if needed) -- #405
- [x] Resolve expected behavior for `ifindex` GET if Switch Interface is incomplete -- #403 & #404 

Co-authored-by: Maximilian Pudelko pudelkoM@users.noreply.github.com


pudelkoM:
How does the control plane get the sdk port is this design?

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -343,6 +377,8 @@ ::util::Status BFChassisManager::PushChassisConfig(
   node_id_to_port_id_to_port_config_ = node_id_to_port_id_to_port_config;
   node_id_to_port_id_to_singleton_port_key_ =
       node_id_to_port_id_to_singleton_port_key;
+  node_id_to_port_id_to_sdk_port_id_ = node_id_to_port_id_to_sdk_port_id;
+  node_id_to_sdk_port_id_to_port_id_ = node_id_to_sdk_port_id_to_port_id_;
```
Yi-Tseng, line 381: should it be `node_id_to_sdk_port_id_to_port_id_ = node_id_to_sdk_port_id_to_port_id; ` ?
(there is one `_` in the end)

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
```diff
@@ -280,6 +281,23 @@ ::util::Status BFPalWrapper::PortLoopbackModeSet(int unit, uint32 port_id,
   return ::util::OkStatus();
 }
 
+::util::Status PortIdFromPortKeyGet(int unit, PortKey port_key,
+                                    uint32* sdk_port_id) {
+  std::string port_str = absl::StrFormat(
+      "%d/%d", port_key.port, port_key.channel);
```
Yi-Tseng, line 287: Note that the chassis config is using a 1-base channel number while SDE uses 0-base. 

In `stratum/hal/lib/common/utils.h` the description of the PortKey shows that there are three possible values of a port key:
non-zero: channelized
0: non-channelized
-1: the channel is not important at all

Yi-Tseng (commented):
I believed that we also need to change the behavior of the `interfaces/interface/state/ifindex` path so the controller can get the correct SDK port number when it connects to the switch.

Yi-Tseng:
> I believed that we also need to change the behavior of the `interfaces/interface/state/ifindex` path so the controller can get the correct SDK port number when it connects to the switch.

"Change the behavior" I mean we can add a new `request` type to `DataRequest` message (e.g., `sdk_port` or `ifindex`...)
And add new `response` to the `DataResponse` message.
So we can use the SwitchInterface to get the real SDK port ID via gNMI.

bocon13:
> How does the control plane get the sdk port is this design?

@pudelkoM The idea was to use `/interfaces/interface[name=<name>]/state/ifindex` and I've updated the PR accordingly.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: good catch

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
bocon13: good catch

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -276,6 +309,7 @@ ::util::Status BFChassisManager::PushChassisConfig(
     // TODO(antonin): we currently ignore slot, port and channel (note that
     // Stratum requires slot and port to be set). We require id to be set to the
     // Tofino device port.
+    // FIXME(bocon) ^^^^^^
```
bocon13, line 312: clean up

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -701,22 +779,27 @@ void BFChassisManager::ReadPortStatusChangeEvents() {
...
-      *state = event.state;
-      SendPortOperStateGnmiEvent(*node_id, event.port_id, event.state);
+      auto* state = gtl::FindOrNull(node_id_to_port_id_to_port_state_[*node_id],
+                                    *port_id);
+      LOG(INFO) << "State of port " << *port_id << " in node " << *node_id
+                << " (SDK port " << event.port_id << " ): "
+                << PrintPortState(event.state) << ".";
+      if (state == nullptr) {
+        *state = event.state;
+      } // else  FIXME(bocon)
```
bocon13, line 801: fix

stratum/hal/lib/barefoot/bf_chassis_manager_test.cc
```diff
@@ -106,9 +109,27 @@ class BFChassisManagerTest : public ::testing::Test {
...
+    EXPECT_CALL(*bf_pal_mock_, PortIdFromPortKeyGet(unit, port_key, _))
+        .WillRepeatedly(DoAll(
+            SetArgPointee<2>(port_id + kSdkPortOffset),
+            Return(::util::OkStatus())));
+  }
+
+  void RegisterSdkPortId(const SingletonPort* singleton_port) {
+    RegisterSdkPortId(singleton_port->id(), singleton_port->slot(),
+                      singleton_port->port(), singleton_port->channel(),
+                      kUnit); //FIXME
```
bocon13, line 129: address this

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
```diff
@@ -280,6 +282,46 @@ ::util::Status BFPalWrapper::PortLoopbackModeSet(int unit, uint32 port_id,
   return ::util::OkStatus();
 }
 
+namespace {
+
+// PortKey uses three possibile values for channel:
+//   > 0: port is channelized (first channel is 1)
+//   0: port is not channelized
+//   < 0: port channel is not important (e.g. for port groups)
+int GetZeroBasedChannel(const PortKey* port_key) {
```
Yi-Tseng, line 291: We can use reference here, like `const PortKey& port_key` so we don't need to pass the address to this function.
and replace `->` to just `.`

stratum/hal/lib/barefoot/bf_pal_mock.h
```diff
@@ -43,9 +43,14 @@ class BFPalMock : public BFPalInterface {
   MOCK_METHOD3(PortMtuSet, ::util::Status(int unit, uint32 port_id, int32 mtu));
 
   MOCK_METHOD2(PortIsValid, bool(int unit, uint32 port_id));
+
   MOCK_METHOD3(PortLoopbackModeSet,
-               ::util::Status(int uint, uint32 port_id,
+               ::util::Status(int unit, uint32 port_id,
```
Yi-Tseng, line 48: good catch, can you also modify the wrapper and the pal_interface?

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -891,12 +974,32 @@ ::util::StatusOr<int> BFChassisManager::GetUnitFromNodeId(
...
 }
 
+::util::StatusOr<uint32> BFChassisManager::GetSdkPortId(
+    uint64 node_id, uint32 port_id) const {
+  if (!initialized_) {
+    return MAKE_ERROR(ERR_NOT_INITIALIZED) << "Not initialized!";
+  }
+  const int* unit = gtl::FindOrNull(node_id_to_unit_, node_id);
+  CHECK_RETURN_IF_FALSE(unit != nullptr)
+      << "Node " << node_id << " is not configured or not known.";
```
pudelkoM, line 984: `unit` is not used later?

stratum/hal/lib/barefoot/bf_pal_wrapper.h
```diff
@@ -54,6 +54,9 @@ class BFPalWrapper : public BFPalInterface {
   ::util::Status PortLoopbackModeSet(int uint, uint32 port_id,
                                      LoopbackState loopback_mode) override;
 
+  ::util::Status PortIdFromPortKeyGet(int unit, PortKey port_key,
```
pudelkoM, line 57:
```suggestion
  ::util::Status PortIdFromPortKeyGet(int unit, const PortKey& port_key,
```

stratum/hal/lib/barefoot/bf_chassis_manager.h
```diff
@@ -73,6 +73,9 @@ class BFChassisManager {
   ::util::StatusOr<int> GetUnitFromNodeId(uint64 node_id) const
       SHARED_LOCKS_REQUIRED(chassis_lock);
 
+  ::util::StatusOr<uint32> GetSdkPortId(uint64 node_id, uint32 port_id) const
+      SHARED_LOCKS_REQUIRED(chassis_lock);
+
```
pudelkoM, line 78: Must be virtual for the mock (which we don't have right now).

pudelkoM: Can also be private, there are no calls from outside this class.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -578,31 +653,34 @@ ::util::Status BFChassisManager::ReplayPortsConfig(uint64 node_id) {
           << "fec_mode field should contain a value";
     }
 
-    RETURN_IF_ERROR(bf_pal_interface_->PortAdd(unit, port_id, *config.speed_bps,
+    ASSIGN_OR_RETURN(auto sdk_port_id, GetSdkPortId(node_id, port_id));
```
pudelkoM, line 656:
```
stratum/hal/lib/barefoot/bf_chassis_manager.cc:656:40: warning: calling function 'GetSdkPortId' requires holding mutex 'chassis_lock' [-Wthread-safety-analysis]
    ASSIGN_OR_RETURN(auto sdk_port_id, GetSdkPortId(node_id, port_id));
```

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -351,6 +391,14 @@ ::util::Status BFChassisManager::PushChassisConfig(
 
 ::util::Status BFChassisManager::VerifyChassisConfig(
     const ChassisConfig& config) {
+  std::map<uint64, int> node_id_to_unit;
```
pudelkoM, line 394: This map is not used anywhere.

stratum/hal/lib/barefoot/bf_chassis_manager.h
bocon13: I was thinking that we want to keep this public to enable future calls (e.g. to enable port translation in P4RT). But we can also make this public in the future if needed.

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
bocon13: Makes sense. Thanks!

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
```diff
@@ -280,6 +282,46 @@ ::util::Status BFPalWrapper::PortLoopbackModeSet(int unit, uint32 port_id,
...
+  // Non-channelized port uses channel 0, or return non-zero
+  return port_key->channel;
+}
+
+}  // namespace
+
+::util::Status BFPalWrapper::PortIdFromPortKeyGet(int unit, PortKey port_key,
+                                                  uint32* sdk_port_id) {
+  int channel = GetZeroBasedChannel(&port_key);
+  if (channel < 0) {
```
pudelkoM, line 305: `CHECK_RETURN_IF_FALSE`

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
```diff
@@ -280,6 +282,46 @@ ::util::Status BFPalWrapper::PortLoopbackModeSet(int unit, uint32 port_id,
   return ::util::OkStatus();
 }
 
+namespace {
+
+// PortKey uses three possibile values for channel:
```
pudelkoM, line 287:
```suggestion
// PortKey uses three possible values for channel:
```

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
```diff
@@ -280,6 +282,46 @@ ::util::Status BFPalWrapper::PortLoopbackModeSet(int unit, uint32 port_id,
...
+  int channel = GetZeroBasedChannel(&port_key);
+  if (channel < 0) {
+    RETURN_ERROR(ERR_INVALID_PARAM) << "Channel must be set for port "
+        << port_key.port << " on dev " << unit << ".";
+  }
+  std::string port_str = absl::StrFormat("%d/%d", port_key.port, channel);
+  bf_dev_port_t dev_port;
+  auto bf_status = bf_pal_port_str_to_dev_port_map(
+      static_cast<bf_dev_id_t>(unit),
+      const_cast<char*>(port_str.c_str()),
```
pudelkoM, line 313: Don't.
https://stackoverflow.com/questions/56434272/why-is-the-const-overload-of-stdstringdata-still-restricted-in-modern-c/56434399#56434399

bocon13: I was wondering if you'd catch this. :)

pudelkoM: There is `string_as_array` or `&str[0]`.

pudelkoM:
> This allows users to specify any port id that they'd like for the ChassisConfig and in gNMI.

I'm not sold on that wording. To me it's seems like we're ignoring the `id` param from the ChassisConfig completely.

Given this SingletonPort:

```
singleton_ports {
  id: 1000
  name: "someport1"
  slot: 1
  port: 1
  speed_bps: 100000000000
  config_params {
    admin_state: ADMIN_STATE_ENABLED
  }
  node: 1
}
```

The gNMI path is `get /interfaces/interface[name=someport1]/state/ifindex`
And the result: 
```
      elem {
        name: "ifindex"
      }
    }
    val {
      uint_val: 128
    }
```

`128` must also be used in P4RT. Where does the `1000` come into play?

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: Was used originally, but replaced call with map lookup and never cleaned this up. fixed

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: It is used below in line 429

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
bocon13: replaced with `snprintf`

bocon13:
> > This allows users to specify any port id that they'd like for the ChassisConfig and in gNMI.
> 
> I'm not sold on that wording. To me it's seems like we're ignoring the `id` param from the ChassisConfig completely.
> 
> Given this SingletonPort:
> 
> ```
> singleton_ports {
>   id: 1000
>   name: "someport1"
>   slot: 1
>   port: 1
>   speed_bps: 100000000000
>   config_params {
>     admin_state: ADMIN_STATE_ENABLED
>   }
>   node: 1
> }
> ```
> 
> The gNMI path is `get /interfaces/interface[name=someport1]/state/ifindex`
> And the result:
> 
> ```
>       elem {
>         name: "ifindex"
>       }
>     }
>     val {
>       uint_val: 128
>     }
> ```
> 
> `128` must also be used in P4RT. Where does the `1000` come into play?

That's accurate. Perhaps, I'll add something to the readme to explain how IDs are used and reword the commit message.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: good catch! clang ftw. moved call out of lambda

stratum/hal/lib/barefoot/bf_chassis_manager.cc
pudelkoM: Isn't that `node_id_to_unit_`, the class member with the `_` at the end?

stratum/hal/bin/barefoot/README.run.md
```diff
@@ -190,6 +189,87 @@ For brevity, we will just show examples
...
+
+#### Port Numbers
+
+Previously, the port `id` was required to match the BF device port ID for Stratum
+to function. Stratum now reads BF device port ID from the SDK using the `node`,
+`port`, and `channel` params provided in the `singleton_port` config, which means
+the port `id` can now be set to a user-selected value.
+
+This port ID (SDN port ID) can be read using gNMI:
+`/interfaces/interface[<name>]/state/id`
```
bocon13, line 261:
```suggestion
`/interfaces/interface[name=<name>]/state/id`
```

stratum/hal/bin/barefoot/README.run.md
```diff
@@ -190,6 +189,87 @@ For brevity, we will just show examples
...
+
+The BF device port ID (SDK port ID) can be read using gNMI:
+`/interfaces/interface[<name>]/state/ifindex`
```
bocon13, line 264:
```suggestion
`/interfaces/interface[name=<name>]/state/ifindex`
```

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: This is a good catch. We are in Verify, so we shouldn't be using the class member in this function. Updating the call in 429 to use the local variable.

stratum/hal/lib/barefoot/bf_chassis_manager.h
```diff
@@ -206,6 +209,19 @@ class BFChassisManager {
...
+  // and SDK port IDs are used in calls to the BF SDK. This map is updated
+  // as part of each config push.
+  std::map<uint64, std::map<uint32, uint32>>
+      node_id_to_port_id_to_sdk_port_id_ GUARDED_BY(chassis_lock);
+
+  // Map from node ID to another map from SDK port ID to (SDN) port ID.
+  // This contains the inverse mapping of: node_id_to_port_id_to_sdk_port_id_
+  // This map is updated as part of each config push.
+  std::map<uint64, std::map<uint32, uint32>>
+      node_id_to_sdk_port_id_to_port_id_ GUARDED_BY(chassis_lock);
```
pudelkoM, line 223: Need to add both to the constructor's [member initializer list](https://en.cppreference.com/w/cpp/language/constructor)

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -926,18 +926,22 @@ void SetUpInterfacesInterfaceStateLastChange(TreeNode* node) {
...
-void SetUpInterfacesInterfaceStateIfindex(uint32 port_id, TreeNode* node) {
+void SetUpInterfacesInterfaceStateIfindex(uint32 node_id, uint32 port_id,
+                                          TreeNode* node,
+                                          YangParseTree* tree) {
+  // Returns system assigned number for each interface
+  // If Stratum performs port translation (e.g. for the bcm target), we return
+  // the port_id. If Stratum does not perform port translation (e.g. for the
+  // bf target), we return the SDK port number.
+  auto poll_functor =
+      GetOnPollFunctor(node_id, port_id, tree,
```
bocon13, line 937: consider falling back to the port_id if nothing is returned...

stratum/hal/lib/common/common.proto
```diff
@@ -1171,6 +1171,11 @@ message FecStatus {
   FecMode mode = 1;
 }
 
+// Wrapper around the SDK port ID.
+message SdkPortId {
```
bocon13, line 1175:
```suggestion
message SdnPortIdOverride {
```

If the port translation is not performed by the target, return the port that should be used.

stratum/hal/lib/np4intel/np4_switch.cc
```diff
@@ -188,6 +189,12 @@ ::util::Status NP4Switch::RetrieveValue(uint64 /*node_id*/,
...
         break;
+      case DataRequest::Request::kSdkPortId:
+        // Stratum hides the SDK port ID, so we just return the SDN port ID
+        resp_val.mutable_sdk_port_id()->set_sdk_port_id(
+            req.sdk_port_id().port_id());
+        resp = resp_val;
+        break;
       default:
         // TODO(antonin)
         resp = MAKE_ERROR(ERR_INTERNAL) << "Not supported yet";
```
bocon13, line 200: Not supported

stratum/hal/lib/common/openconfig_converter.cc
```diff
@@ -246,9 +246,8 @@ SingletonPortToInterfaces(const SingletonPort &in) {
   interface_key.set_name(in.name());
   auto interface = interface_key.mutable_interface();
 
-  // SingletonPort.id -> /interfaces/interface/state/id(ifindex)
+  // SingletonPort.id -> /interfaces/interface/config/id
```
bocon13, line 249: comment

stratum/hal/lib/common/openconfig_converter.cc
```diff
@@ -343,9 +342,8 @@ TrunkPortToInterfaces(const ChassisConfig &root, const TrunkPort &in) {
   interface_key.set_name(in.name());
   auto trunk = interface_key.mutable_interface();
 
-  // SingletonPort.id -> /interfaces/interface/state/id(ifindex)
+  // SingletonPort.id -> /interfaces/interface/config/id
```
bocon13, line 345: comment

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -533,7 +613,8 @@ ::util::Status BFChassisManager::GetPortCounters(uint64 node_id, uint32 port_id,
     return MAKE_ERROR(ERR_NOT_INITIALIZED) << "Not initialized!";
   }
   ASSIGN_OR_RETURN(auto unit, GetUnitFromNodeId(node_id));
-  return bf_pal_interface_->PortAllStatsGet(unit, port_id, counters);
+  ASSIGN_OR_RETURN(auto sdk_port_id, GetSdkPortId(node_id, port_id));
+  return bf_pal_interface_->PortAllStatsGet(unit, sdk_port_id, counters);
   return ::util::OkStatus();
```
Yi-Tseng, line 618: unreachable statement.

bocon13: thanks

stratum/hal/bin/barefoot/README.run.md
```diff
@@ -190,6 +189,87 @@ For brevity, we will just show examples
...
+the port `id` can now be set to a user-selected value.
+
+This port ID (SDN port ID) can be read using gNMI:
+`/interfaces/interface[name=<name>]/state/id`
+
+The BF device port ID (SDK port ID) can be read using gNMI:
+`/interfaces/interface[name=<name>]/state/ifindex`
+
+In both cases, the `name` used in the gNMI paths should match the name provided
+in the `singleton_port` config.
```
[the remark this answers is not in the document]
bocon13: Do you mean singleton_ports? Or something else?

stratum/hal/lib/common/common.proto
```diff
@@ -1171,6 +1171,16 @@ message FecStatus {
...
 
+// Wrapper around the SDN port ID.
+// The target is expected to perform accept requests from the controller
+// using the ID assigned by the SDN management system. If the target
+// cannot accept requests using the provided ID, it can provide
+// and alterate ID that should be used by the controller to address
+// the interface.  
+message SdnPortIdOverride {
+  uint32 port_id = 1;
+}
```
pudelkoM, line 1182: I wanted to suggest to makes this part of `FrontPanelPortInfo` to make it easy to detect switches that don't set the value, but then I found that `bcm_switch` does not implement that one, too.

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -926,18 +926,63 @@ void SetUpInterfacesInterfaceStateLastChange(TreeNode* node) {
...
-void SetUpInterfacesInterfaceStateIfindex(uint32 port_id, TreeNode* node) {
+void SetUpInterfacesInterfaceStateIfindex(uint32 node_id, uint32 port_id,
+                                          TreeNode* node,
+                                          YangParseTree* tree) {
+  // Returns the port ID for the interface to be used by P4Runtime.
+  // If Stratum performs port translation (e.g. for the bcm target), we return
+  // the port_id provided by the ChassisConfig (also called SDN port ID).
+  // If Stratum does not perform port translation (e.g. for the bf target), we
+  // return the SDK port number.
+  auto on_poll_functor = [tree, port_id, node_id](const GnmiEvent& event,
```
bocon13, line 937: @pudelkoM to review this

stratum/hal/lib/barefoot/bf_pal_wrapper.h
```diff
@@ -51,9 +51,12 @@ class BFPalWrapper : public BFPalInterface {
 
   bool PortIsValid(int unit, uint32 port_id) override;
 
-  ::util::Status PortLoopbackModeSet(int uint, uint32 port_id,
+  ::util::Status PortLoopbackModeSet(int unit, uint32 port_id,
                                      LoopbackState loopback_mode) override;
 
+  ::util::Status PortIdFromPortKeyGet(int unit, const PortKey& port_key,
```
pudelkoM, line 57: This function should have signature:
```
::util::StatusOr<uint32> PortIdFromPortKeyGet(
      int unit, const PortKey& port_key)
```

> Prefer using return values over output parameters: they improve readability, and often provide the same or better performance.
> Use non-const pointers to represent **optional** outputs.

https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -236,15 +255,20 @@ ::util::Status BFChassisManager::PushChassisConfig(
...
   std::map<PortKey, HwState> xcvr_port_key_to_xcvr_state;
 
-  int unit(0);
-  for (auto& node : config.nodes()) {
-    unit_to_node_id[unit] = node.id();
-    node_id_to_unit[node.id()] = unit;
-    unit++;
+  {
+    int unit(0);
+    for (auto& node : config.nodes()) {
```
pudelkoM, line 264:
```suggestion
    int unit = 0;
    for (const auto& node : config.nodes()) {
```

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -261,21 +285,31 @@ ::util::Status BFChassisManager::PushChassisConfig(
...
+
     PortKey port_group_key(singleton_port.slot(), singleton_port.port());
     xcvr_port_key_to_xcvr_state[port_group_key] = HW_STATE_UNKNOWN;
   }
 
-  ::util::Status status = ::util::OkStatus();  // errors to keep track of.
+  // If there was an error parsing the ports, return early.
+  RETURN_IF_ERROR(status);
 
   for (auto singleton_port : config.singleton_ports()) {
```
pudelkoM, line 304:
```suggestion
  for (const auto singleton_port : config.singleton_ports()) {
```

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -360,21 +404,50 @@ ::util::Status BFChassisManager::VerifyChassisConfig(
...
-    std::map<uint64, std::map<uint32, PortKey>>
-        node_id_to_port_id_to_singleton_port_key;
+  // Validate singleton ports
+  std::map<uint64, std::map<uint32, PortKey>>
+      node_id_to_port_id_to_singleton_port_key;
+  {
+    std::map<uint64, int> node_id_to_unit;
+    // Map node ids to 0-based units
+    int unit(0);
+    for (auto& node : config.nodes()) {
```
pudelkoM, line 414:
```suggestion
    int unit = 0;
    for (const auto& node : config.nodes()) {
```

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -360,21 +404,50 @@ ::util::Status BFChassisManager::VerifyChassisConfig(
...
+            status, bf_pal_interface_->PortIdFromPortKeyGet(
+                        *unit, singleton_port_key, &unused_sdk_port_id));
+      }
+    }
+    if (!status.ok()) {
+      ::util::Status error = MAKE_ERROR(ERR_INVALID_PARAM)
+                             << "Error in singleton ports configuration.";
+      error.Update(status);
+      return error;
     }
```
pudelkoM, line 445:
```suggestion
    RETURN_IF_ERROR(status)
            << "Error in singleton ports configuration.";
```

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
```diff
@@ -280,6 +280,47 @@ ::util::Status BFPalWrapper::PortLoopbackModeSet(int unit, uint32 port_id,
...
+  //     Convert base-1 channel to base-0 channel if port is channelized
+  //     Otherwise, port is already 0 in the non-channelized case
+  const int channel =
+      (port_key.channel > 0) ? port_key.channel - 1 : port_key.channel;
+  CHECK_RETURN_IF_FALSE(channel >= 0)
+      << "Channel must be set for port " << port << " on dev " << unit << ".";
+
+  char port_string[MAX_PORT_HDL_STRING_LEN];
+  int r =
+      snprintf(port_string, MAX_PORT_HDL_STRING_LEN, "%d/%d", port, channel);
```
pudelkoM, line 305:
```suggestion
      snprintf(port_string, sizeof(port_string), "%d/%d", port, channel);
```

See: any other usage of `snprintf`

stratum/hal/lib/bmv2/bmv2_switch.cc
```diff
@@ -174,6 +174,7 @@ ::util::Status Bmv2Switch::RetrieveValue(uint64 /*node_id*/,
                                          std::vector<::util::Status>* details) {
   absl::ReaderMutexLock l(&chassis_lock);
   for (const auto& req : request.requests()) {
+    DataResponse resp_val;
```
pudelkoM, line 177: ???

bocon13: no longer needed

stratum/hal/lib/np4intel/np4_switch.cc
```diff
@@ -178,6 +178,7 @@ ::util::Status NP4Switch::RetrieveValue(uint64 /*node_id*/,
                                         std::vector<::util::Status>* details) {
   absl::ReaderMutexLock l(&chassis_lock);
   for (const auto& req : request.requests()) {
+    DataResponse resp_val;
```
bocon13, line 181: no longer needed
```suggestion
```

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -351,6 +391,10 @@ ::util::Status BFChassisManager::PushChassisConfig(
 
 ::util::Status BFChassisManager::VerifyChassisConfig(
     const ChassisConfig& config) {
+  if (config.nodes_size() != 1) {
```
pudelkoM, line 394: I'm not sure if we support more than one node yet, or if there even is a chip of that kind.

We should probably adjust the error message.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -351,6 +391,10 @@ ::util::Status BFChassisManager::PushChassisConfig(
 
 ::util::Status BFChassisManager::VerifyChassisConfig(
```
bocon13, line 392: add unit test

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: The code is set up to support multiple nodes, but there is an implicit mapping between Stratum node and SDK unit ID. That is, unit IDs are assigned sequentially based on the order of the nodes in the ChassisConfig which seems pretty brittle. Now does anyone use this? I don't know, but there definitely are vendors that build chassis switches or even 1U pizza boxes with more than one switching ASIC.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: I wanted an INVALID_PARAM error if there are any (rather than UNKNOWN), and couldn't figure out a better way to do that.

pudelkoM: `APPEND_ERROR` returns a new error object:

```
// Return a new error based on an existing error, with an additional string
// appended.  Otherwise behaves like MAKE_ERROR, including logging the error by
// default.
// Requires !status.ok().
// Example:
//   status = APPEND_ERROR(status) << ", more details";
//   return APPEND_ERROR(status) << ", more details";
#define APPEND_ERROR(status) \
  ::util::status_macros::MakeErrorStream((status), __FILE__, __LINE__)
```

bocon13: I believe the error code of the status doesn't change; it just appends to the error message. My goal was to make sure that the code of the returned error is ERR_INVALID_PARAM.

Also, I don't know if Update is doing what I want either.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
stolsma: And I dont know how it works or how the SDK handles it but there is a possibility to split the 2/4 pipelines from each other and configure them differently (as if they are different switches as I understood). It could be that the SDK gives the splitted pipelines different node id's....

pudelkoM:
> The other implementation was a little more convoluted, but the logically is subtly different. I'm not sure which one we want.
> 
> Older implementation would always return the port_id if an override port could not be determined and then log an error if we don't expect the error. This new implementation returns the port_id only if it receives a particular error, and it might not even return the right port ID if the callee chooses to update the sdn_port and return the UNIMPLEMENTED error.

This seems like a hint that having partial switch_interface implementations is bad design.

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -926,18 +926,47 @@ void SetUpInterfacesInterfaceStateLastChange(TreeNode* node) {
...
+      if (resp.has_sdn_port_id_override()) {
+        sdn_port = resp.sdn_port_id_override().port_id();
+        return true;
+      }
+      return false;
+    });
+
+    // Query the switch for the overridden port ID.
+    std::vector<::util::Status> details;
+    RETURN_IF_ERROR(tree->GetSwitchInterface()->RetrieveValue(
```
bocon13, line 958: Depending on if/how we intend to support P4RT translations, we might have another mechanism for getting this data into the functor in the future.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13, line 394: CHECK_RETURN_IF_FALSE

bocon13: needs to be positive

stratum/hal/lib/dummy/dummy_switch.cc
```diff
@@ -288,7 +288,12 @@ ::util::Status DummySwitch::RetrieveValue(uint64 node_id,
         break;
       }
       default:
```
bocon13, line 290: implement SdnPortId in all switch implementations to return identity

bocon13: #403

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: #405

bocon13: We should just check that the nodes_size() > 0

bocon13: Addressed in #405

stratum/hal/lib/barefoot/bf_pal_wrapper.cc
```diff
@@ -239,13 +239,46 @@ ::util::Status BFPalWrapper::PortLoopbackModeSet(int unit, uint32 port_id,
...
+  CHECK_RETURN_IF_FALSE(channel >= 0)
+      << "Channel must be set for port " << port << " on dev " << unit << ".";
+
+  char port_string[MAX_PORT_HDL_STRING_LEN];
+  int r = snprintf(port_string, sizeof(port_string), "%d/%d", port, channel);
+  if (r < 0 || r >= MAX_PORT_HDL_STRING_LEN) {
+    RETURN_ERROR(ERR_INVALID_PARAM) << "Failed to build port string"
+                                    << " for port " << port << " channel "
+                                    << channel << " on dev " << unit << ".";
+  }
```
bocon13, line 274:
```suggestion
  CHECK_RETURN_IF_FALSE(r > 0 && r < MAX_PORT_HDL_STRING_LEN)
      << "Failed to build port string" << " for port " << port << " channel "
      << channel << " on dev " << unit << ".";
```

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -561,7 +614,8 @@ ::util::Status BFChassisManager::ReplayPortsConfig(uint64 node_id) {
   LOG(INFO) << "Replaying ports for node " << node_id << ".";
 
   auto replay_one_port = [node_id, unit, this](
-                             uint32 port_id, const PortConfig& config,
+                             uint32 port_id,
+                             const PortConfig& config,
```
bocon13, line 618:
```suggestion
                             uint32 port_id, const PortConfig& config,
```

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: moved to issue

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -330,10 +359,11 @@ ::util::Status BFChassisManager::PushChassisConfig(
         continue;
       }
       auto unit = node_id_to_unit_[node_id];
+      uint32 sdk_port_id = node_id_to_port_id_to_sdk_port_id_[node_id][port_id];
       // remove ports which are no longer present in the ChassisConfig
-      LOG(INFO) << "Deleting port " << port_id << " in node " << node_id << ".";
-      APPEND_STATUS_IF_ERROR(status,
-                             bf_pal_interface_->PortDelete(unit, port_id));
+      LOG(INFO) << "Deleting port " << port_id << " in node " << node_id
```
bocon13, line 364:
```suggestion
      // TODO(bocon): Collect these errors and keep trying to remove old ports (don't fail on first one)
      LOG(INFO) << "Deleting port " << port_id << " in node " << node_id
```

pudelkoM:
Just found this gnmi path: `/interfaces/interface/state/hardware-port`
We have a DataRequest field for that: https://github.com/stratum/stratum/blob/master/stratum/hal/lib/common/common.proto#L1219 (That only the dummy implements)

bocon13:
> Just found this gnmi path: `/interfaces/interface/state/hardware-port`
> 
> > For non-channelized interfaces, references the hardware port corresponding to the base interface.
> 
> We have a DataRequest field for that: https://github.com/stratum/stratum/blob/master/stratum/hal/lib/common/common.proto#L1219 (That only the dummy implements)

I think this is supposed to be refer to the front panel port number...
https://github.com/openconfig/hercules/blob/51ed75ced12bbd651ce3934a3b3cdc5fb3b0b98e/docs/hercules-models.md#paths-used-for-node-interface-mapping
But I'm also not positive.

pudelkoM:
> But I'm also not positive.

Reading the description again together with the type (string reference to name), makes me think this is ONLY for channelized ports and should link to the main port, e.g. `/interfaces/interface[name="1/1"]/state/hardware-port` -> `"1/0"`

Not related to this issue.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -57,84 +57,95 @@ BFChassisManager::BFChassisManager(PhalInterface* phal_interface,
...
       node_id_to_port_id_to_singleton_port_key_(),
+      node_id_to_port_id_to_sdk_port_id_(),
+      node_id_to_sdk_port_id_to_port_id_(),
       xcvr_port_key_to_xcvr_state_() {}
 
 BFChassisManager::~BFChassisManager() = default;
 
 ::util::Status BFChassisManager::AddPortHelper(
-    uint64 node_id, int unit, uint32 port_id,
+    uint64 node_id, int unit, uint32 sdk_port_id,
```
pudelkoM, line 67: I keep wondering if this would not be better solved by a `BfPort`, like the `BcmPort`: https://github.com/stratum/stratum/blob/9f5bd2b285badbef11e81eca6c31d4a3c4342843/stratum/hal/lib/bcm/bcm.proto#L91

Encapsulates both the ID, node, and sdk port in one message to pass around.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -325,12 +353,12 @@ ::util::Status BFChassisManager::PushChassisConfig(
...
       // ADMIN_STATE_UNKNOWN (invalid)
-      APPEND_STATUS_IF_ERROR(
-          status, UpdatePortHelper(node_id, unit, port_id, singleton_port,
-                                   *config_old, &config));
+      RETURN_IF_ERROR(UpdatePortHelper(node_id, unit, sdk_port_id,
+                                       singleton_port, *config_old, &config));
     }
   }
 
+  // Clean up from old config
```
pudelkoM, line 361:
```suggestion
  // Clean up from old config.
```

stratum/hal/lib/barefoot/bf_chassis_manager.cc
bocon13: Let's address this in the wrapper

PR #377, opened 2020-10-28
Add SDN port ID to RetrieveValue for stratum_bfrt target

bocon13:
depends on #403


stratum/hal/lib/barefoot/bfrt_switch.cc
```diff
@@ -204,7 +204,8 @@ ::util::Status BfrtSwitch::RetrieveValue(uint64 node_id,
       case DataRequest::Request::kPortCounters:
       case DataRequest::Request::kAutonegStatus:
       case DataRequest::Request::kFrontPanelPortInfo:
-      case DataRequest::Request::kLoopbackStatus: {
+      case DataRequest::Request::kLoopbackStatus:
+      case DataRequest::Request::kSdkPortId: {
```
bocon13, line 208:
```suggestion
      case DataRequest::Request::kSdnPortId: {
```

PR #381, opened 2020-10-29
Fix Barefoot SDE and container builds

### LINKED ISSUE #382: Used experimental function in docker build
**Describe the bug**
When follow readme document to build barefoot stratum, in method 2 will need to execute 
`stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh`
The process will be fail due using `squash` parameter in docker build command.
The error message is following:
`Error response from daemon: squash is only supported with experimental mode`

**Environment**
OS: Debian 9
Client: Docker Engine - Community
 Version:           19.03.12
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        48a66213fe
 Built:             Mon Jun 22 15:45:52 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.12
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       48a66213fe
  Built:            Mon Jun 22 15:44:23 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683

If building from source:
Yes, build by script from repo.

Switch:
Build on other server

**To Reproduce**
1. Install latest Docker-CE
2. Execute `docker buid --squash`

**Expected behavior**
It will remind it is an experimental function

**Additional context**
It can be resolve by add config to docker.
1. Create /etc/docker/daemon.json, and add following content
```
{
        "experimental": true
}
```
2. Restart docker service

bocon13:
In release build mode:
```
-rw-------   1 bocon  staff    35M Oct 31 04:38 stratum-bf-9.2.0-docker.tar.gz  # down from 205M (without compression)
-r-xr-xr-x   1 bocon  staff    15M Oct 31 04:38 stratum-bf_0.0.1_amd64.deb      # down from 50M
-r-xr-xr-x   1 bocon  staff    15M Oct 31 04:38 stratum_bf_deb.deb

REPOSITORY                   TAG                                      IMAGE ID            CREATED             SIZE
stratumproject/stratum-bf    9.2.0                                    25ca75d8495d        24 minutes ago      131MB
```
fixes #382


stratum/hal/bin/barefoot/deb/postinst
```diff
@@ -1,14 +1,23 @@
...
 # SPDX-License-Identifier: Apache-2.0
-# Set up the huge page
-echo "vm.nr_hugepages = 128" >> /etc/sysctl.conf
-sysctl -p /etc/sysctl.conf
-mkdir -p /mnt/huge
-mount -t hugetlbfs nodev /mnt/huge
+
+# If not running in Docker,
+if ! cat /proc/1/cgroup | grep -q "/docker"; then
+  # Set up the huge page
```
bocon13, line 7:
```suggestion
  # Set up huge pages
```

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
```diff
@@ -96,36 +102,56 @@ else
...
-docker build --squash \
+docker build \
+  $DOCKER_OPTS \
   -t "$RUNTIME_IMAGE" \
   --build-arg STRATUM_TARGET="$STRATUM_TARGET" \
   -f "$DOCKERFILE_DIR/Dockerfile" \
   "$(pwd)"
 
 docker save $RUNTIME_IMAGE -o ${STRATUM_NAME}-${SDE_VERSION}-docker.tar
+gzip ${STRATUM_NAME}-${SDE_VERSION}-docker.tar
```
bocon13, line 154:
```suggestion
docker save $RUNTIME_IMAGE | gzip > ${STRATUM_NAME}-${SDE_VERSION}-docker.tar.gz
```

bocon13: This is slightly faster (5.4 sec vs. 5.6 sec) and a little cleaner.

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
```diff
@@ -43,6 +44,10 @@ if [ -n "$1" ]; then
   SDE_TAR_DIR=$( cd $(dirname "$SDE_TAR") >/dev/null 2>&1 && pwd )
   SDE_TAR_NAME=$( basename $SDE_TAR )
   DOCKER_OPTS+="-v $SDE_TAR_DIR:/bf-tar "
+  if [ -t 0 ]; then
+    # Running in a TTY
```
bocon13, line 48:
```suggestion
    # Running in a TTY, so make the shells interactive (so things like Ctrl-C work)
```

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
```diff
@@ -96,36 +102,56 @@ else
...
+fi
 
-# Build Stratum BF in Docker
+# Build Stratum BF in Docker (optimized and stripped)
+EXTRA_BUILD_OPTS=""
+if [ -n "$RELEASE_BUILD" ]; then
+  # Build with optimization enabled (-O2) and with assert() calls disabled (-DNDEBUG)
+  EXTRA_BUILD_OPTS+="--compilation_mode=opt "
+  # Strip all symbols
+  EXTRA_BUILD_OPTS+="--linkopt=-Wl,--strip-all "
```
pudelkoM, line 116: These should be in the bazelrc.

stratum/hal/bin/barefoot/BUILD
```diff
@@ -118,10 +118,8 @@ pkg_deb(
     name = "stratum_bf_deb",
     architecture = "amd64",
     data = ":stratum_bf_data",
-    depends = [
-        "systemd",
-        "kmod",
-    ],
+    depends = [ "kmod" ],
```
pudelkoM, line 121:
```suggestion
    depends = ["kmod"],
```

buildifier

stratum/hal/bin/barefoot/deb/postinst
```diff
@@ -1,14 +1,22 @@
...
-mount -t hugetlbfs nodev /mnt/huge
+
+# If not running in Docker,
+if ! cat /proc/1/cgroup | grep -q "/docker"; then
+  # Set up huge pages
+  echo "vm.nr_hugepages = 128" >> /etc/sysctl.conf
+  sysctl -p /etc/sysctl.conf
+  mkdir -p /mnt/huge
+  mount -t hugetlbfs nodev /mnt/huge
+fi
```
pudelkoM, line 12: The logic here would be that the page setup would always happen outside of docker?

pudelkoM: The problem is that we can't modify the host at deb install time, right?

stratum/hal/bin/barefoot/deb/postinst
```diff
@@ -1,14 +1,22 @@
...
 ldconfig
 
-# Enables Stratum service
-systemctl enable stratum_bf.service
+# If systemd is installed,
+if command -v systemctl &> /dev/null
+then
+  # Enable Stratum service
+  systemctl enable stratum_bf.service
+fi
```
pudelkoM, line 22: Did you know [`dh_systemd_enable`](https://manpages.debian.org/testing/debhelper/dh_systemd_enable.1.en.html)?

Never used it, but might be applicable to our use case.

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
```diff
@@ -96,36 +102,56 @@ else
...
+     cp -f /stratum/bazel-bin/stratum/hal/bin/barefoot/${STRATUM_TARGET}_deb.deb /output/ && \
+     cp -f \$(readlink -f /stratum/bazel-bin/stratum/hal/bin/barefoot/${STRATUM_TARGET}_deb.deb) /output/"
 set +x
 
+
+DOCKER_OPTS=""
+if [ "$(docker version | grep Experimental | grep true | wc -l)" -eq "2" ]; then
+  DOCKER_OPTS+="--squash "
+fi
+
```
pudelkoM, line 140: I'd place the all opts in a single place, else you might wonder why it changes during script execution.

stratum/hal/bin/barefoot/BUILD
```diff
@@ -118,10 +118,8 @@ pkg_deb(
...
+    recommends = [ "systemd" ],
```
bocon13, line 122:
```suggestion
    recommends = ["systemd"],
```

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
bocon13: Docker options are specific to each of the 3 docker invocations. `--squash` for example is only relevant to the 3rd: docker build. If this is confusing, perhaps theres another way that we can call each variable.

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
bocon13: This is a good idea. Will do

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
pudelkoM: I didn't see that the bind command is in there too. Fine as is then.

stratum/hal/bin/barefoot/deb/postinst
bocon13: `sysctl` can't be used in Docker unless running with `--sysctl` or `--privileged` because the container and the host share the same kernel. Even if it could be run, it doesn't make sense to enable huge pages on the host building the image. Similar issue with `mount`...

Anyway, we need to enable huge pages on the switch manually each time the OS is installed, or we could add a check for huge pages in the `start-stratum.sh` script and do it there if needed.

stratum/hal/bin/barefoot/deb/postinst
bocon13: The intention of this block is to skip installing the service file if the user doesn't have systemd (e.g. in the Docker container). I haven't heard of `dh_systemd_enable` but we could see if it does what we need.

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
```diff
@@ -96,36 +102,55 @@ else
...
        --jobs $JOBS && \
-     cp /stratum/bazel-bin/stratum/hal/bin/barefoot/${STRATUM_TARGET}_deb.deb /output/ && \
-     cp \$(readlink -f /stratum/bazel-bin/stratum/hal/bin/barefoot/${STRATUM_TARGET}_deb.deb) /output/"
+     cp -f /stratum/bazel-bin/stratum/hal/bin/barefoot/${STRATUM_TARGET}_deb.deb /output/ && \
+     cp -f \$(readlink -f /stratum/bazel-bin/stratum/hal/bin/barefoot/${STRATUM_TARGET}_deb.deb) /output/"
 set +x
 
+
+DOCKER_OPTS=""
+if [ "$(docker version | grep Experimental | grep true | wc -l)" -eq "2" ]; then
```
pudelkoM, line 137: The recommended way to check for that is:

```bash
> docker version -f '{{.Server.Experimental}}'
true
```

stratum/hal/bin/barefoot/deb/prerm
```diff
@@ -2,9 +2,13 @@
 # Copyright 2020-present Open Networking Foundation
 # SPDX-License-Identifier: Apache-2.0
 
-# Stop and disable Stratum service
-systemctl stop stratum_bf.service
-systemctl disable stratum_bf.service
+# If systemd is installed,
+if command -v systemctl &> /dev/null
+then
```
pudelkoM, line 7:
```suggestion
if command -v systemctl &> /dev/null; then
```

stratum/hal/bin/barefoot/deb/postinst
bocon13: I think dh_systemd_enable is just to help in the debian packaging process, which is managed by Bazel. It would then autogenerate code for the postinst, prerm, etc. scripts similar to what we have.

bocon13: It seems like there is a recommendation to use `deb-systemd-helper` in maintscripts, though
https://people.debian.org/~stapelberg/debconf13-making-your-package-work-with-systemd.pdf

stratum/hal/bin/barefoot/build-bf-sde.sh
```diff
@@ -132,9 +132,9 @@ popd
 echo "BF SDE build complete."
```
pudelkoM, line 133: Just found this one:

```
In stratum/hal/bin/barefoot/build-bf-sde.sh line 115:
$sudo pip install pyrsistent==0.14.0
^---^ SC2154: sudo is referenced but not assigned.
```

stratum/hal/bin/barefoot/deb/postinst
pudelkoM: Looks like a good idea, but should be applied to all platforms then. Out of scope for this PR.

stratum/hal/bin/barefoot/deb/postinst
```diff
@@ -1,14 +1,22 @@
...
 
 # Reload libraries
 ldconfig
 
-# Enables Stratum service
-systemctl enable stratum_bf.service
+# If systemd is installed,
+if command -v systemctl &> /dev/null
+then
```
bocon13, line 19:
```suggestion
if command -v systemctl &> /dev/null; then
```

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
bocon13, line 137:
```suggestion
if [ "$(docker version -f '{{.Server.Experimental}}')" -eq "true" ]; then
```

bocon13: That is definitely cleaner.

stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
```diff
@@ -96,36 +102,55 @@ else
   print_help
   exit 1
 fi
+if [ -t 0 ]; then
+  # Running in a TTY
+  DOCKER_OPTS+="-it "
```
bocon13, line 107: change to EXTRA_DOCKER_RUN_OPTS and move to top

abhilashendurthi:
verified building stratum-bf for `9.0.0` and `9.2.0`, both builds were successful.
9.0.0 Jenkins [build 751](https://jenkins.stratumproject.org/blue/organizations/jenkins/stratum-bf-build/detail/stratum-bf-build/751/pipeline/ )
9.2.0 Jenkins [build 752](https://jenkins.stratumproject.org/blue/organizations/jenkins/stratum-bf-build/detail/stratum-bf-build/752/pipeline/)

abhilashendurthi (approved):

PR #384, opened 2020-10-29
Fix missing bf_lib in pkg

breezestars:
- Add libbf_switchd_lib.so, libpiall.so


pudelkoM:
Makes me wonder how I compiled this before???

Yi-Tseng:
> Makes me wonder how I compiled this before???

You can still build everything (the deb). But you cannot run it since the library is missing in the package.

bocon13:
Thanks for the PR, we might be tweaking the way we build stratum_bf again and in the process revert some of the changes that require these libraries to be included.

See #385

PR #385, opened 2020-10-30
Remove libavago, libpython3.4m from direct bfsde link targets

pudelkoM:
They are needed as a shared lib at runtime by the SDE, but not when compiling Stratum code.
Linking libavago causes linker errors around missing symbols and mismatched libstdc++ versions.

### DIFF
--- a/bazel/external/bfsde.BUILD
+++ b/bazel/external/bfsde.BUILD
@@ -12,12 +12,10 @@ package(
 cc_library(
     name = "bfsde",
     srcs = glob([
-        "barefoot-bin/lib/libavago.so*",
         "barefoot-bin/lib/libbf_switchd_lib.so*",
         "barefoot-bin/lib/libbfsys.so*",
         "barefoot-bin/lib/libbfutils.so*",
         "barefoot-bin/lib/libdriver.so*",
-        "barefoot-bin/lib/libpython3.4m.so*",
     ]),
     hdrs = glob([
         "barefoot-bin/include/bf_switchd/*.h",


bazel/external/bfsde.BUILD
```diff
@@ -12,12 +12,10 @@ package(
 cc_library(
     name = "bfsde",
     srcs = glob([
-        "barefoot-bin/lib/libavago.so*",
         "barefoot-bin/lib/libbf_switchd_lib.so*",
```
bocon13, line 15: We should consider statically linking libbf_switchd_lib.

bazel/external/bfsde.BUILD
```diff
@@ -12,12 +12,10 @@ package(
...
         "barefoot-bin/lib/libbfsys.so*",
         "barefoot-bin/lib/libbfutils.so*",
         "barefoot-bin/lib/libdriver.so*",
-        "barefoot-bin/lib/libpython3.4m.so*",
```
bocon13, line 20: Needed by libdriver, which we plan to dynamically link, so this can be removed.

bazel/external/bfsde.BUILD
```diff
@@ -12,12 +12,10 @@ package(
 cc_library(
     name = "bfsde",
     srcs = glob([
-        "barefoot-bin/lib/libavago.so*",
         "barefoot-bin/lib/libbf_switchd_lib.so*",
         "barefoot-bin/lib/libbfsys.so*",
         "barefoot-bin/lib/libbfutils.so*",
         "barefoot-bin/lib/libdriver.so*",
```
bocon13, line 18: dynamically link. depends on:

```
libdriver.so.0
	libpython3.4m.so.1.0
	libavago.so.0
``

File: bazel/external/bfsde.BUILD [hunk 4]
```diff
@@ -12,12 +12,10 @@ package(
 cc_library(
     name = "bfsde",
     srcs = glob([
-        "barefoot-bin/lib/libavago.so*",
```
bocon13, line 15: Depended on by libdriver which is dynamically linked. Can remove

bazel/external/bfsde.BUILD
```diff
@@ -12,12 +12,10 @@ package(
...
         "barefoot-bin/lib/libbf_switchd_lib.so*",
         "barefoot-bin/lib/libbfsys.so*",
         "barefoot-bin/lib/libbfutils.so*",
```
bocon13, line 17: Can be static or dynamic (was static before), depends on libbfsys.so

pudelkoM:
Root cause of the issue was that the SDE package was compiled on gcc-6, while Stratum used gcc-5.

PR #386, opened 2020-10-30
Remove support for the tar pipeline format in Stratum-bf

pudelkoM:
This PR removes support for the tar pipeline format in Stratum-bf.

The deprecated flag `-incompatible_enable_p4_device_config_tar` is removed as well. Use the protobuf or PI format instead.


stratum/hal/lib/barefoot/bf_pipeline_utils.cc
```diff
@@ -72,76 +30,6 @@ std::string Uint32ToLeByteStream(uint32 val) {
 }
 }  // namespace
 
-::util::Status ExtractBfPipelineConfig(
```
pudelkoM, line 75: @bocon13 After removing format 2, this function collapsed to basically
```c++
  bf_config->Clear();
  if (bf_config->ParseFromString(config.p4_device_config())) {
    return util::OkStatus();
  } else {
   RETURN_ERROR(ERR_INVALID_PARAM) << "Unknown format for p4_device_config";
  }
```

and I removed it and its tests completely. But maybe there is value in having it as a standalone function?

stratum/hal/lib/barefoot/bf_switch.cc
```diff
@@ -111,14 +111,16 @@ namespace {
...
-  if (ExtractBfPipelineConfig(forwarding_config, &bf_config).ok()) {
+  if (bf_config.ParseFromString(forwarding_config.p4_device_config())) {
     std::string pi_p4_device_config;
     RETURN_IF_ERROR(
         BfPipelineConfigToPiConfig(bf_config, &pi_p4_device_config));
     legacy_config->set_p4_device_config(pi_p4_device_config);
+  } else {
+    *legacy_config = forwarding_config;
   }
+
```
pudelkoM, line 123: To me having the other code path in an explicit branch made the code easier to read, and it matches the comment (Otherwise, ...)

stratum/hal/lib/barefoot/bf_pipeline_utils.cc
bocon13: I agree that this method is perhaps trivial now, but if we decide to support other formats in the future (e.g. the message format directly from the SDE's bfrt.proto) then this might be useful.

stratum/hal/lib/barefoot/bf_pipeline_utils_test.cc
```diff
@@ -41,150 +41,6 @@ static constexpr char bf_config_2pipe_str[] = R"PROTO(
...
-    std::string bf_config_bytes;
-    ASSERT_TRUE(bf_config.SerializeToString(&bf_config_bytes));
-    p4_config.set_p4_device_config(bf_config_bytes);
-  }
-  BfPipelineConfig extracted_bf_config;
-  EXPECT_OK(ExtractBfPipelineConfig(p4_config, &extracted_bf_config));
-  EXPECT_TRUE(ProtoEqual(bf_config, extracted_bf_config))
-      << "Expected: " << bf_config.ShortDebugString()
-      << ", got: " << extracted_bf_config.ShortDebugString();
-}
```
bocon13, line 87: These tests seem like they are still valid.

stratum/hal/lib/barefoot/bf_switch.cc
```diff
@@ -111,14 +111,16 @@ namespace {
 ::util::Status ConvertToLegacyForwardingPipelineConfig(
     const ::p4::v1::ForwardingPipelineConfig& forwarding_config,
     ::p4::v1::ForwardingPipelineConfig* legacy_config) {
-  *legacy_config = forwarding_config;
   BfPipelineConfig bf_config;
   if (ExtractBfPipelineConfig(forwarding_config, &bf_config).ok()) {
     std::string pi_p4_device_config;
     RETURN_IF_ERROR(
         BfPipelineConfigToPiConfig(bf_config, &pi_p4_device_config));
     legacy_config->set_p4_device_config(pi_p4_device_config);
```
bocon13, line 119: I think the legacy_config will be missing the p4info

pudelkoM: Good catch!

bocon13 (approved):
lgtm

PR #394, opened 2020-11-03
Return gnmi poll/get errors back to the client

pudelkoM:
Currently gNMI get errors inside Stratum are not reported back to the client. Instead an empty response is delivered.
This seems like an oversight or not fully finished code, instead of on purpose. But we should check the call sites of `HandlePoll`. At first glance they all expect `HandlePoll` to sometimes fail.

New behavior:
```
REQUEST
path {
  elem {
    name: "interfaces"
  }
  elem {
    name: "interface"
    key {
      key: "name"
      value: "1/0"
    }
  }
  elem {
    name: "state"
  }
  elem {
    name: "ifindex"
  }
}
encoding: PROTO

E20201102 17:01:08.019973 316675520 gnmi_cli.cc:249] Return Error: stub->Get(&ctx, req, &resp) failed with generic::invalid_argument: Something
```


bocon13:
The spec is quite clear about what is expected:
https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#25-error-handling

Is this issue limited to one function, or do we need to take a look at the whole class?

stratum/hal/lib/common/gnmi_publisher.cc
```diff
@@ -114,6 +114,7 @@ ::util::Status GnmiPublisher::HandlePoll(const SubscriptionHandle& handle) {
   if ((status = (*handle)(PollEvent())) != ::util::OkStatus()) {
     // Something went wrong.
     LOG(ERROR) << "Handler returned non-OK status: " << status;
+    return status;
```
bocon13, line 117: RETURN_IF_ERROR

stratum/hal/lib/common/gnmi_publisher.cc
```diff
@@ -114,6 +114,7 @@ ::util::Status GnmiPublisher::HandlePoll(const SubscriptionHandle& handle) {
   if ((status = (*handle)(PollEvent())) != ::util::OkStatus()) {
```
bocon13, line 114: ^^^ HandleChange discards error as well

Check to see if there are other discarded statuses

bocon13 (approved):
lgtm

PR #395, opened 2020-11-03
Fix invalid uses of APPEND_ERROR ignoring the new status

pudelkoM:
`APPEND_ERROR` creates a new Status object, therefore the return value must be assigned to something. This PR fixes the existing instances where this is not the case.

<s>I don't think constructors can be marked with `MUST_USE_RESULT`, we should verify this and other options</s>. We can mark them, but the `MakeErrorStream` class is complicated and I did not find a place which does not cause false positives.

TODO
- [x] Fix all instances


stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -852,8 +852,8 @@ ::util::Status BFChassisManager::UnregisterEventWriters() {
   APPEND_STATUS_IF_ERROR(
```
Yi-Tseng, line 852: check this one, too

pudelkoM: Checked. This is correct usage.

bocon13 (approved):

PR #396, opened 2020-11-03
Add CI format check for modified C++ files

pudelkoM:
Since we can't enable `clang-format` globally yet, this PR enforces code format only on the modified files of a PR and against a list of known good files.

The check is not bullet-proof since it only applies against the `master` branch. Due to the temporary nature of this sub-feature (we aim for 100% compliance), the team decided to not implement a base branch lookup and instead override the check in PRs as needed. Should we need this feature, we have some pointers on how to get started:

> Could use `CIRCLE_PULL_REQUEST` and `GITHUB_TOKEN` env vars
> 
> https://api.github.com/repos/stratum/stratum/pulls/396
> data.base.ref contains the branch
>

### DIFF
--- a/.circleci/check-cpp-format.sh
+++ b/.circleci/check-cpp-format.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+set -ex
+
+# Files in this branch that are different from master.
+FILES=$(git diff-tree --no-commit-id --name-status -r HEAD origin/master \
+          | grep -v "^D" \
+          | cut -f 2 \
+          | grep -F -e ".cc" -e ".h" || :
+      )
+
+# List of files that are already formatted.
+read -r -d '\0' KNOWN_FILES << EOF
+stratum/tools/gnmi/gnmi_cli.cc
+\0
+EOF
+
+for file in "${FILES}" "${KNOWN_FILES}"; do
+    [ -f "$file" ] || continue  # handle empty match
+    clang-format --style=file -i "${file}"
+done
+
+# Report which files need to be fixed.
+git update-index --refresh
+git diff
--- a/.circleci/check-cpp-format.sh
+++ b/.circleci/check-cpp-format.sh
@@ -150,8 +150,7 @@ stratum/tools/gnmi/gnmi_cli.cc
 \0
 EOF
 
-echo "$CHANGED_FILES" | xargs -t -n1 clang-format --style=file -i
-echo "$KNOWN_FILES" | xargs -t -n1 clang-format --style=file -i
+echo -e "$KNOWN_FILES\n$CHANGED_FILES" | sort -u | xargs -t -n1 clang-format --style=file -i
 
 # Report which files need to be fixed.
 git update-index --refresh
--- a/.circleci/check-cpp-format.sh
+++ b/.circleci/check-cpp-format.sh
@@ -1,4 +1,8 @@
 #!/bin/bash
+#
+# Copyright 2020-present Open Networking Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
 set -ex
 
 # Files in this branch that are different from master.
--- a/.circleci/check-cpp-format.sh
+++ b/.circleci/check-cpp-format.sh
@@ -154,4 +154,3 @@ echo -e "$KNOWN_FILES\n$CHANGED_FILES" | sort -u | xargs -t -n1 clang-format --s
 
 # Report which files need to be fixed.
 git update-index --refresh
-git diff
--- a/.circleci/check-cpp-format.sh
+++ b/.circleci/check-cpp-format.sh
@@ -15,18 +15,11 @@ CHANGED_FILES=$(git diff-tree --no-commit-id --name-status -r HEAD origin/master
 # List of files that are already formatted.
 read -r -d '\0' KNOWN_FILES << EOF
 stratum/glue/gtl/cleanup_test.cc
-stratum/glue/gtl/cleanup_test.cc
-stratum/glue/gtl/cleanup.h
 stratum/glue/gtl/cleanup.h
 stratum/glue/gtl/map_util_test.cc
-stratum/glue/gtl/map_util_test.cc
-stratum/glue/gtl/map_util.h
 stratum/glue/gtl/map_util.h
 stratum/glue/gtl/source_location.h
-stratum/glue/gtl/source_location.h
 stratum/glue/gtl/stl_util.h
-stratum/glue/gtl/stl_util.h
-stratum/glue/init_google.h
 stratum/glue/init_google.h
 stratum/glue/status/posix_error_space.cc
 stratum/glue/status/status_test_util.cc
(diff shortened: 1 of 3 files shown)


Yi-Tseng:
Should we keep both cpplint and clang-format check?

pudelkoM:
> Should we keep both cpplint and clang-format check?

Yes, there is some minor overlap (line length), but cpplint mainly does semantic validation.

.circleci/config.yml
```diff
@@ -273,6 +273,17 @@ jobs:
...
 
+  cpp-format-check:
+    docker:
+      - image: stratumproject/build:build
+    steps:
+      - checkout
+      - run:
+          name: Run clang-format on files modified by commit
+          command: |
+            git diff-tree --no-commit-id --name-only -r HEAD origin/master | grep -F -e ".cc" -e ".h" | xargs -r clang-format --style=file -i
```
Yi-Tseng, line 284: check if a file removed

Yi-Tseng: git .... --name-status .... | grep -v D | awk '{print $2}' | ...

pudelkoM: very good idea

Yi-Tseng:
I think grep will return 1 if no cc or h file changed (e.g. fix a typo to README)
We can either ignore the error code from grep or add another check I think

.circleci/check-cpp-format.sh
```diff
@@ -0,0 +1,171 @@
+#!/bin/bash
+set -ex
+
+# Files in this branch that are different from master.
+CHANGED_FILES=$(git diff-tree --no-commit-id --name-status -r HEAD origin/master \
+                  | grep -v "^D" \
+                  | cut -f 2 \
+                  | grep -F -e ".cc" -e ".h" || :
```
pudelkoM, line 8: The `|| :` ensure grep will not return 1 on no match

.circleci/check-cpp-format.sh
```diff
@@ -0,0 +1,175 @@
...
+#
+# Copyright 2020-present Open Networking Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+set -ex
+
+# Files in this branch that are different from master.
+CHANGED_FILES=$(git diff-tree --no-commit-id --name-status -r HEAD origin/master \
+                  | grep -v "^D" \
+                  | cut -f 2 \
```
bocon13, line 11: git has a millions ways to do things, but this might be a little simpler:
```
git diff --name-only --diff-filter=d <BASE REF>
```

.circleci/config.yml
```diff
@@ -306,6 +315,7 @@ workflows:
       - unit_tests
       - cdlang_tests
       - coverage
+      - cpp-format-check
```
bocon13, line 318: We might want to set another environment variable to the job to determine the base branch. I think `pipeline.git.base_revision` or `pipeline.pull_request.target_branch` should do. Then this new variable can be used as the base ref in the diff command.

https://circleci-public.github.io/pipeline-preview-docs/pipeline_variables_overview.html
https://circleci-public.github.io/pipeline-preview-docs/pipeline_variables_reference.html

.circleci/check-cpp-format.sh
```diff
@@ -0,0 +1,175 @@
+#!/bin/bash
+#
+# Copyright 2020-present Open Networking Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+set -ex
+
+# Files in this branch that are different from master.
```
bocon13, line 8: The base ref is not always master.

.circleci/check-cpp-format.sh
bocon13: This will return a list of all files that are not deleted between <BASE REF> and HEAD

bocon13: Actually we can get this whole thing down to a one-liner :)

```
git diff --name-only --diff-filter=d <BASE REF> -- '*.h' '*.cc'
```

bocon13: And git diff returns 0 whether or not it finds anything, so we can drop the `|| :`

.circleci/check-cpp-format.sh
```diff
@@ -0,0 +1,175 @@
...
+CHANGED_FILES=$(git diff-tree --no-commit-id --name-status -r HEAD origin/master \
+                  | grep -v "^D" \
+                  | cut -f 2 \
+                  | grep -F -e ".cc" -e ".h" || :
+              )
+
+# List of files that are already formatted.
+read -r -d '\0' KNOWN_FILES << EOF
+stratum/glue/gtl/cleanup_test.cc
```
bocon13, line 17: Are there supposed to be so many duplicates?

.circleci/check-cpp-format.sh
```diff
@@ -0,0 +1,175 @@
...
+stratum/testing/tests/bcm_sim_test_fixture.cc
+stratum/testing/tests/bcm_sim_test_fixture.h
+stratum/testing/tests/bcm_sim_test.cc
+stratum/testing/tests/test_main.cc
+stratum/tools/gnmi/gnmi_cli.cc
+\0
+EOF
+
+echo "$CHANGED_FILES" | xargs -t -n1 clang-format --style=file -i
+echo "$KNOWN_FILES" | xargs -t -n1 clang-format --style=file -i
```
bocon13, line 171:
```suggestion
echo "$KNOWN_FILES\n$CHANGED_FILES" | sort -u | xargs -t -n1 clang-format --style=file -i
```

We can avoid potentially duplicate work.

.circleci/check-cpp-format.sh
```diff
@@ -0,0 +1,175 @@
...
+
+# Report which files need to be fixed.
+git update-index --refresh
+git diff
```
bocon13, line 175:
```suggestion
git diff --exit-code
```

Returns 1 if there are updates that are not committed (normal git diff always returns 0)

bocon13: Or replace both lines with:
```
git diff --name-only --exit-code
```

.circleci/config.yml
bocon13: If we don't want to add an environment variable, we could also pass the base ref as an arg to `check-cpp-format.sh`

.circleci/check-cpp-format.sh
pudelkoM: no, copy&paste fail

.circleci/check-cpp-format.sh
pudelkoM: I'm not sure about the exact diff anyway. It will tell you the exact changes required, but you will still just apply the tool. Just the affected files seems sufficient information.

.circleci/check-cpp-format.sh
pudelkoM: Amazing!

.circleci/config.yml
pudelkoM: `pipeline.git.base_revision` is the immediate previous commit.

`pipeline.pull_request.target_branch` is not yet available.

.circleci/check-cpp-format.sh
```diff
@@ -0,0 +1,157 @@
...
+stratum/p4c_backends/test/test_p4c_main.cc
+stratum/p4c_backends/test/test_target_info.cc
+stratum/p4c_backends/test/test_target_info.h
+stratum/testing/tests/bcm_sim_test_fixture.cc
+stratum/testing/tests/bcm_sim_test_fixture.h
+stratum/testing/tests/bcm_sim_test.cc
+stratum/testing/tests/test_main.cc
+stratum/tools/gnmi/gnmi_cli.cc
+\0
+EOF
```
Yi-Tseng, line 151: what is the reason we need to scan files that we already formatted?

.circleci/config.yml
bocon13: Could use `CIRCLE_PULL_REQUEST` and `GITHUB_TOKEN`

https://api.github.com/repos/stratum/stratum/pulls/396
`data.base.ref` contains the branch

.circleci/check-cpp-format.sh
pudelkoM: Removed `diff` but kept the `update-index`, it has the kinda neat message after the filename:

```
+ git update-index --refresh
.circleci/check-cpp-format.sh: needs update
stratum/hal/lib/barefoot/bf_switch.cc: needs update
```

Yi-Tseng (approved):

PR #403, opened 2020-11-06
Add SdnPortId to RetrieveValue and use it to populate ifindex in yang tree

bocon13:
Co-authored-by: Maximilian Pudelko pudelkoM@users.noreply.github.com


stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -287,12 +287,16 @@ ::util::Status BFChassisManager::PushChassisConfig(
         singleton_port_key;
 
     // Translate the logical SDN port to SDK port (device port ID)
-    uint32 sdk_port_id;
-    APPEND_STATUS_IF_ERROR(status,
-                           bf_pal_interface_->PortIdFromPortKeyGet(
-                               *unit, singleton_port_key, &sdk_port_id));
-    node_id_to_port_id_to_sdk_port_id[node_id][port_id] = sdk_port_id;
-    node_id_to_sdk_port_id_to_port_id[node_id][sdk_port_id] = port_id;
+
```
pudelkoM, line 290: newline

stratum/hal/lib/bcm/bcm_switch.cc
```diff
@@ -426,10 +426,15 @@ ::util::Status BcmSwitch::RetrieveValue(uint64 /*node_id*/,
             req.optical_transceiver_info().network_interface(),
             resp.mutable_optical_transceiver_info()));
         break;
+      case DataRequest::Request::kSdnPortId:
+        // Return the requested port ID because port translation is performed
```
pudelkoM, line 430:
```suggestion
        // Return the requested port ID because port translation is performed.
```

pudelkoM:
Since this touches all platforms now and adds a new message to SwitchInterface, I feel that this should be a PR to master.

stratum/hal/lib/barefoot/bf_switch.cc
```diff
@@ -277,9 +278,10 @@ ::util::Status BFSwitch::RetrieveValue(uint64 node_id,
         break;
       }
       default:
-        status =
-            MAKE_ERROR(ERR_UNIMPLEMENTED)
-            << "Request type "
+        status = MAKE_ERROR(ERR_UNIMPLEMENTED).without_logging()
+                 << "Not supported yet.";
+        VLOG(1)
+            << "DataRequest field "
```
pudelkoM, line 284: Since we decided that everyone must implement the full switch interface, is this change still valid?

If there is a unhandled type, should not be logged as ERROR and fixed (i.e with a dummy value like bcm_switch)?

stratum/hal/lib/common/common.proto
```diff
@@ -1171,6 +1171,16 @@ message FecStatus {
   FecMode mode = 1;
 }
 
+// Wrapper around the SDN port ID.
+// The target is expected to perform accept requests from the controller
```
pudelkoM, line 1175:
```suggestion
// The target is expected to accept requests from the controller
```

stratum/hal/lib/common/testdata/invalid_iface_component.pb.txt
```diff
@@ -38,8 +38,5 @@ interface {
     id {
       value: 1
     }
-    ifindex {
-      value: 1
-    }
```
pudelkoM, line 43: Can someone confirm that this removal is still needed after the latest design change?
My understanding of the test code is very minimal.

stratum/hal/lib/bcm/bcm_chassis_manager.cc
```diff
@@ -1375,7 +1375,7 @@ ::util::Status BcmChassisManager::SyncInternalState(
 
   // TODO(unknown): Update the LED of all the ports.
 
-  return ::util::OkStatus();
+  return error;
 }
```
bocon13, line 1380: missing kSdnPortId branch in GetPortData()

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -492,6 +492,10 @@ ::util::StatusOr<DataResponse> BFChassisManager::GetPortData(
         resp.mutable_loopback_status()->set_state(*config->loopback_mode);
       break;
     }
+    case DataRequest::Request::kSdnPortId: {
+      resp.mutable_sdn_port_id()->set_port_id(req.sdn_port_id().port_id());
```
bocon13, line 496:
```suggestion
      resp.mutable_sdn_port_id()->set_port_id(
          request.sdn_port_id().port_id());
```

stratum/hal/lib/barefoot/bf_switch.cc
bocon13: This is a good point. I'll revert this and make sure the error gets logged in all the switches

stratum/hal/lib/common/testdata/invalid_iface_component.pb.txt
bocon13: Yes, I believe it is still necessary. The openconfig_converter translates between oc.Device and ChassisConfig, and thus only parts of the tree that are present in both should be included. More practically, without a pointer to switch interface, we can't fill in the ifindex part anymore, and it is also part of the "state" tree, instead of the "config" tree. In this case, "id" corresponds to the SDN port number (or ChassisConfig SingletonPort ID), and we shouldn't be validating or including the ifindex as part of the config.

stratum/hal/lib/common/openconfig_converter.cc
```diff
@@ -246,9 +246,8 @@ SingletonPortToInterfaces(const SingletonPort &in) {
   interface_key.set_name(in.name());
   auto interface = interface_key.mutable_interface();
 
-  // SingletonPort.id -> /interfaces/interface/state/id(ifindex)
+  // SingletonPort.id -> /interfaces/interface/config/id
   interface->mutable_id()->set_value(in.id());
-  interface->mutable_ifindex()->set_value(in.id());
```
bocon13, line 251: @pudelkoM -- see we no longer translate include the ifindex in either direction of translation.

PR #405, opened 2020-11-06
Add APPEND_ERROR_WITH_CODE macro for error chaining

bocon13:
merge into #376

### DIFF
--- a/stratum/lib/macros.h
+++ b/stratum/lib/macros.h
@@ -2,7 +2,6 @@
 // Copyright 2018-present Open Networking Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-
 #ifndef STRATUM_LIB_MACROS_H_
 #define STRATUM_LIB_MACROS_H_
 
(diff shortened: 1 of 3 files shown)


stratum/lib/macros.h
```diff
@@ -66,6 +66,16 @@ inline const std::string FixMessage(const std::string& msg) {
                   : " ")                                                       \
           << FixMessage(__ret.status().error_message())
 
+// A macro for simplifying creation of a new error or appending new info to an
+// error based on the provided error code
+#define APPEND_ERROR_WITH_CODE(out, code)                                      \
+  if (!out.ok()) assert(out.error_code() == code);                             \
```
bocon13, line 72: assert() isn't really used on any non-test codepath, so we might want to do something else here (e.g. log an error)

bocon13:
Drop this...

bocon13:
Use this pattern instead

```
    ::util::Status error = MAKE_ERROR(ERR_REBOOT_REQUIRED)
                           << "Detected a change in BcmRxConfig for node_id: "
                           << node_id;
    APPEND_STATUS_IF_ERROR(status, error);
```

PR #406, opened 2020-11-06
Add configs for Wedge100BF-32QS

Yi-Tseng:


stratum/hal/config/x86-64-accton-wedge100bf-32qs-r0/chassis_config.pb.txt
```diff
@@ -0,0 +1,362 @@
+description: "Default Chassis Config for Edgecore Wedge100BF-32x"
+chassis {
+  platform: PLT_BAREFOOT_TOFINO
+  name: "Edgecore Wedge100BF-32x"
```
Yi-Tseng, line 4: fix

Yi-Tseng:
let's merge this to bfrt first and put it back to master

PR #407, opened 2020-11-06
[Do not merge] Add configs for Wedge100BF-32QS 

Yi-Tseng:


pudelkoM:
You want to merge this to `bfrt` first because we don't know if it works yet?

Yi-Tseng:
> You want to merge this to `bfrt` first because we don't know if it works yet?

Right, let me put this to draft. I think we can still merge this change to master first and merge master to the bfrt branch later

PR #415, opened 2020-11-09
Replace ChassisConfig platform Tofino enums with PLT_GENERIC_BAREFOOT_TOFINO

pudelkoM:
Replace the the existing `PLT_BAREFOOT_TOFINO` enums with `PLT_GENERIC_BAREFOOT_TOFINO` to match the BCM pattern.

The old enums are deprecated and will remain supported for one (1) more release.

TODO
- [x] let people (slack, TST, mailing list) know they need to change the config


stratum/hal/lib/common/common.proto
```diff
@@ -38,8 +38,8 @@ enum Platform {
   PLT_GENERIC_TOMAHAWK_PLUS = 4;
   PLT_MLNX_SN2700 = 9;
   PLT_P4_SOFT_SWITCH = 10;
-  PLT_BAREFOOT_TOFINO = 11;
-  PLT_BAREFOOT_TOFINO2 = 12;
```
Yi-Tseng, line 42: People will have an issue if they still use the old format. I would suggest using `[deprecated = true];` and add a new enum item.

pudelkoM: Makes sense, deprecate for the next release and remove in the next.

Yi-Tseng: right, and remember to let people (slack, TST, mailing list) know they need to change the config.
Not sure if the protobuf library will show any warning about using deprecated enum.

stratum/hal/lib/common/common.proto
```diff
@@ -38,8 +38,8 @@ enum Platform {
...
+  PLT_GENERIC_BAREFOOT_TOFINO = 11;
+  PLT_GENERIC_BAREFOOT_TOFINO2 = 12;
```
bocon13, line 42:
```suggestion
  PLT_BAREFOOT_TOFINO = 11;
  PLT_BAREFOOT_TOFINO2 = 12;
  PLT_GENERIC_BAREFOOT_TOFINO = 13;
  PLT_GENERIC_BAREFOOT_TOFINO2 = 14;
```

bocon13 (dismissed):
lgtm

PR #416, opened 2020-11-10
Replace PAL and PD wrappers with BfSdeWrapper

pudelkoM:
This PR replaces the PAL and PD wrappers with one BfSdeWrapper. This allows us to remove duplicate code and prepare for future changes to the underlying API (like BFRT).


stratum/hal/lib/barefoot/bf_sde_interface.h
```diff
@@ -0,0 +1,75 @@
...
+#include "stratum/lib/channel/channel.h"
+
+namespace stratum {
+namespace hal {
+namespace barefoot {
+
+// TODO: docs
+class BfSdeInterface {
+ public:
+  struct PortStatusChangeEvent {
```
pudelkoM, line 22: Rename to `LinkscanEvent`?

stratum/hal/lib/barefoot/bf_sde_interface.h
```diff
@@ -0,0 +1,75 @@
...
+    int device;
+    int port;
+    PortState state;
+    // constructor
```
pudelkoM, line 26: todo

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -28,23 +27,23 @@ namespace stratum {
 namespace hal {
 namespace barefoot {
 
-using PortStatusChangeEvent = BFPalInterface::PortStatusChangeEvent;
+using LinkscanEvent = BfSdeInterface::LinkscanEvent;
```
bocon13, line 30: I think "linkscan" is a term that Broadcom uses in their SDK, and isn't necessarily standardized.

I'm okay with the name if you want to have commonality between targets, but personally, I find PortStatusChangeEvent to be more descriptive.

stratum/hal/lib/barefoot/bf_sde_interface.h
```diff
@@ -0,0 +1,74 @@
...
+
+  virtual ::util::Status GetPortCounters(int device, int port,
+                                         PortCounters* counters) = 0;
+
+  virtual ::util::Status RegisterLinkscanEventWriter(
+      std::unique_ptr<ChannelWriter<LinkscanEvent> > writer) = 0;
+
+  virtual ::util::Status UnregisterLinkscanEventWriter() = 0;
+
+  virtual ::util::Status PortAdd(int device, int port, uint64 speed_bps,
```
pudelkoM, line 40:
```suggestion
  virtual ::util::Status AddPort(int device, int port, uint64 speed_bps,
```

Same for the others

stratum/hal/lib/barefoot/bf_chassis_manager.cc
pudelkoM: The SDE seems to use port status notification or change. Commonality in the wrappers would be partial at best anyway, so not something we should pursue.

I'd still drop the `Change` part of the event name and fix the `VerbNoun` naming of some of the functions.

pudelkoM: Looks good after the change.

We should keep the terminologies separate, e.g. `unit` vs. `device`.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -810,27 +808,27 @@ void BFChassisManager::ReadPortStatusChangeEvents() {
     // Handle received message.
     {
       absl::WriterMutexLock l(&chassis_lock);
-      const uint64* node_id = gtl::FindOrNull(unit_to_node_id_, event.unit);
+      const uint64* node_id = gtl::FindOrNull(unit_to_node_id_, event.device);
```
pudelkoM, line 811:
```suggestion
      const uint64* node_id = gtl::FindOrNull(device_to_node_id_, event.device);
```

stratum/hal/lib/barefoot/bf_sde_wrapper.h
```diff
@@ -0,0 +1,103 @@
...
+namespace barefoot {
+
+// The "BfSdeWrapper" is an implementation of BfSdeInterface which is used
+// on real hardware to talk to Tofino ASIC.
+class BfSdeWrapper : public BfSdeInterface {
+ public:
+  // Default MTU for for Tofino.
+  static constexpr int32 kBfDefaultMtu = 10 * 1024;  // 10K
+
+  // BcmSdkInterface public methods.
```
pudelkoM, line 28:
```suggestion
  // BfSdeInterface public methods.
```

stratum/hal/lib/barefoot/bf_sde_wrapper.h
```diff
@@ -0,0 +1,103 @@
...
+
+namespace stratum {
+namespace hal {
+namespace barefoot {
+
+// The "BfSdeWrapper" is an implementation of BfSdeInterface which is used
+// on real hardware to talk to Tofino ASIC.
+class BfSdeWrapper : public BfSdeInterface {
+ public:
+  // Default MTU for for Tofino.
```
pudelkoM, line 25:
```suggestion
  // Default MTU for ports on Tofino.
```

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -180,78 +119,150 @@ ::util::StatusOr<bf_loopback_mode_e> LoopbackModeToBf(
...
+  counters->set_in_unknown_protos(0);  // stat not meaningful
+  counters->set_in_errors(stats[bf_mac_stat_FrameswithanyError]);
+  counters->set_out_errors(stats[bf_mac_stat_FramesTransmittedwithError]);
+  counters->set_in_fcs_errors(stats[bf_mac_stat_FramesReceivedwithFCSError]);
+
+  return ::util::OkStatus();
+}
+
+::util::Status BfSdeWrapper::OnPortStatusEvent(int device, int port, bool up,
+                                               void* cookie) {
```
Yi-Tseng, line 165: remove the `cookie` parameter?

Yi-Tseng (changes requested):
Modify the `check-cpp-format.sh`

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -859,32 +857,32 @@ void BFChassisManager::ReadPortStatusChangeEvents() {
...
+        VLOG(1) << "Unknown SDK port " << event.port << " in node " << *node_id
+                << ".";
         continue;
       }
       auto* state = gtl::FindOrNull(node_id_to_port_id_to_port_state_[*node_id],
                                     *port_id);
       LOG(INFO) << "State of port " << *port_id << " in node " << *node_id
-                << " (SDK port " << event.port_id
-                << " ): " << PrintPortState(event.state) << ".";
-      if (state == nullptr) {
```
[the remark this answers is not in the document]
pudelkoM: This particular check got introduced very recently in #376.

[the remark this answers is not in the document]
bocon13: +1, sorry about that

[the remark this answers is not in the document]
Yi-Tseng: Ah, ok!

stratum/hal/lib/barefoot/bf_chassis_manager.cc
pudelkoM: Postponed to later PR due to size of change.

stratum/hal/lib/barefoot/bf_sde_wrapper.h
```diff
@@ -0,0 +1,104 @@
...
+  // RW mutex lock for protecting the singleton instance initialization and
+  // reading it back from other threads. Unlike other singleton classes, we
+  // use RW lock as we need the pointer to class to be returned.
+  static absl::Mutex init_lock_;
+
+  // The singleton instance.
+  static BfSdeWrapper* singleton_ GUARDED_BY(init_lock_);
+
+ private:
+  // Timeout for Write() operations on linkscan events.
```
pudelkoM, line 85:
```suggestion
  // Timeout for Write() operations on port status events.
```

Yi-Tseng (approved):

stratum/hal/lib/barefoot/bf_chassis_manager.h
```diff
@@ -141,26 +141,26 @@ class BFChassisManager {
   void TransceiverEventHandler(int slot, int port, HwState new_state)
       LOCKS_EXCLUDED(chassis_lock);
 
-  // helper to add / configure / enable a port with BFPalInterface
+  // helper to add / configure / enable a port with BfSdeInterface
   ::util::Status AddPortHelper(uint64 node_id, int unit, uint32 port_id,
```
bocon13, line 145: We might consider moving these into the sdk_wrapper.

stratum/hal/lib/barefoot/bf_sde_wrapper.h
```diff
@@ -0,0 +1,104 @@
...
+  // the instance.
+  static BfSdeWrapper* CreateSingleton() LOCKS_EXCLUDED(init_lock_);
+
+  // The following public functions are specific to this class. They are to be
+  // called by SDE callbacks only.
+
+  // Return the singleton instance to be used in the SDE callbacks.
+  static BfSdeWrapper* GetSingleton() LOCKS_EXCLUDED(init_lock_);
+
+  // Called whenever a linkscan event is received from SDK. It forwards the
```
bocon13, line 63: linkscan

stratum/hal/lib/barefoot/bf_sde_interface.h
```diff
@@ -0,0 +1,80 @@
...
+namespace hal {
+namespace barefoot {
+
+// TODO(max): docs
+class BfSdeInterface {
+ public:
+  // PortStatusEvent encapsulates the information received on a port status
+  // event. Port refers to the SDE internal device port ID.
+  struct PortStatusEvent {
+    int device;
```
bocon13, line 26: device refers to the sde device id. Should we consider translating these to be Stratum IDs?

pudelkoM: We kinda do already.

Check the calls to `SendPortOperStateGnmiEvent`, it uses the node ID: https://github.com/stratum/stratum/blob/12a01d535e8a0ace7e35e5ad215c9c172f8ab22d/stratum/hal/lib/barefoot/bf_chassis_manager.cc#L885

Unless you meant something else.

bocon13: I guess the problem is that the wrapper doesn't really know about Stratum IDs yet?

pudelkoM: `event.device` is used in one place only: `ReadPortStatusEvents` of the ChassisManger.

bocon13: What I was referring to is that we could maintain SDK port numbers entirely in the wrapper if we passed them through. This would keep these details hidden from callers.

pudelkoM: Not sure if the wrapper needs to know about P4rt node ids?

stratum/hal/lib/barefoot/bf_chassis_manager.h
bocon13: If we decide to do this, we can do it later

stratum/hal/lib/barefoot/bf_sde_interface.h
bocon13: We can do this later if we want/need to.

stratum/hal/lib/barefoot/bf_sde_wrapper.h
```diff
@@ -0,0 +1,104 @@
...
+  // linkscan event to the module who registered a callback by calling
+  // RegisterPortStatusEventWriter().
```
bocon13, line 65:
```suggestion
  // Called whenever a port status notification is received from SDK. It
  // forwards the linkscan event to the module who registered a callback by
  // calling RegisterPortStatusEventWriter().
```

stratum/hal/lib/barefoot/bf_sde_wrapper.h
pudelkoM: yes we do. Found no other instance of `linkscan`.

PR #417, opened 2020-11-10
Package gnmi caps file into debs

pudelkoM:
This PR packages the `gnmi_caps.pb.txt` file into the Debian packages, so it gets passed through to the Docker images and is available at runtime.

The caps file is currently hardcoded and resides here: https://github.com/stratum/stratum/blob/master/stratum/hal/lib/common/gnmi_caps.pb.txt
But unless `bazel run` is used, the data dependency is not picked up.

With the `exports_files()` command it is exposed to other Bazel BUILD files, where we can include it in the tar archives and finally the deb packages: `stratum/hal/bin/barefoot/BUILD`

The packages are installed and the file ends up in `/etc/stratum/`: https://github.com/stratum/stratum/blob/9cd027643d99be135f2c0376be96ab480748378d/stratum/hal/bin/barefoot/docker/Dockerfile#L10

At runtime the file is read into the `cap` response: https://github.com/stratum/stratum/blob/c14c8ae4f9de73d2036bba9a839925b2a62b71ff/stratum/hal/lib/common/config_monitoring_service.cc#L178

Thanks @eth0xFEED for reporting this issue!


stratum/hal/lib/common/config_monitoring_service_test.cc
```diff
@@ -52,6 +53,7 @@ class ConfigMonitoringServiceTest
  protected:
   void SetUp() override {
     FLAGS_chassis_config_file = FLAGS_test_tmpdir + "/config.pb.txt";
+    FLAGS_gnmi_capabilities_file = "stratum/hal/lib/common/gnmi_caps.pb.txt";
```
pudelkoM, line 56: This is why the path has to be a flag. For testing we use different paths.

Yi-Tseng (approved):
lgtm

bocon13 (dismissed):
lgtm

PR #425, opened 2020-11-13
EXPERIMENTAL: Adding local BFRuntime server/client for testing

bocon13:


stratum/hal/lib/barefoot/bfruntime.cc
```diff
@@ -0,0 +1,119 @@
...
+    auto table_entry = notification->mutable_table_entry();
+    table_entry->set_table_id(5);
+    auto table_key = table_entry->mutable_key();
+    auto key_field = table_key->add_fields();
+    key_field->mutable_exact()->set_value("6");
+    auto table_data = table_entry->mutable_data();
+    table_data->set_action_id(7);
+    auto data_field = table_data->add_fields();
+    data_field->set_field_id(8);
+    TODO(bocon) other fields...
```
bocon13, line 90:
```suggestion
    // TODO(bocon) other fields...
```

stratum/hal/lib/barefoot/bfruntime.cc
```diff
@@ -0,0 +1,119 @@
...
+    data_field->set_field_id(8);
+    // TODO(bocon) other fields...
+    // ::absl::Time t2 = ::absl::Now();
+
+    // uint8_t data[10000000];
+    // ::google::protobuf::io::ZeroCopyOutputStream* out =
+    //     new ::google::protobuf::io::ArrayOutputStream(data, sizeof(data));
+    // if(!resp.SerializeToZeroCopyStream(out)) break;
+    // ::absl::Time t3 = ::absl::Now();
+    if(!stream->Write(resp)) break;
```
pudelkoM, line 98: Does it make any difference when the `resp` is created inside the loop and not reused?

pudelkoM: I does!

```
Read 3659 in 5.00019293s
100k in 2m16.65463050025s
```

vs

```
Read 214822 in 5.000197231s
100k in 2.327600167s
```

My money is on some cow happening in the background due to the modifications done after `Write`.

PR #431, opened 2020-11-17
Remove older BF SDEs per Stratum deprecation policy

bocon13:


Yi-Tseng (commented):
Merge this after TST this week?

pudelkoM (approved):

Issue #371, opened 2020-10-23
Tag containers with vcs-ref label


bocon13:
We can generate the tag fairly easily:

```bash
GIT_URL=${GIT_URL:-"https://github.com/stratum/stratum"}
GIT_REF=$(git describe --tags --no-match --always --abbrev=40 --dirty | sed -E 's/^.*-g([0-9a-f]{40}-?.*)$/\1/')
# 2020-09-30 (if we match a tag) or cbe426de52bb5db99aa7a147dc5177c86eb83392 or cbe426de52bb5db99aa7a147dc5177c86eb83392-dirty
GIT_SHA=$(git describe --tags --match XXXXXXX --always --abbrev=40 --dirty)
# cbe426de52bb5db99aa7a147dc5177c86eb83392 or cbe426de52bb5db99aa7a147dc5177c86eb83392-dirty
```

Should we use Label Schema or OCI Image Spec? Or both?
Apparently, OCI Image Spec is more current.

http://label-schema.org/rc1/
```
--label vcs-url=$GIT_URL
--label version=$GIT_REF
--label vcs-ref=$GIT_SHA
```

https://github.com/opencontainers/image-spec/blob/master/annotations.md
```bash
--label org.opencontainers.image.source=$GIT_URL
--label org.opencontainers.image.version=$GIT_REF
--label org.opencontainers.image.revision=$GIT_SHA
```

This applies to runtime containers:

- [x] stratum/hal/bin/bcm/standalone/docker/Dockerfile
- [x] stratum/hal/bin/barefoot/docker/Dockerfile
- [ ] stratum/hal/bin/np4intel/docker/Dockerfile.runtime
- [x] tools/mininet/Dockerfile
- [x] Dockerfile.build

But we could also consider it for the builder container, which could help someone debug something?

Labels on a container can be printed using the following:
```bash
docker inspect --format='{{json .ContainerConfig.Labels}}' stratumproject/stratum-bf:9.3.0 | python -m json.tool
```

cc: @pudelkoM @Yi-Tseng @ccascone

bocon13:
```bash
$ docker inspect --format='{{json .ContainerConfig.Labels}}' stratumproject/stratum-bf:9.3.0 | python -m json.tool
{
    "description": "This Docker image includes runtime library for Barefoot Tofino switches",
    "maintainer": "Stratum dev <stratum-dev@lists.stratumproject.org>",
[...]

PR #432, opened 2020-11-17
Add git version tags to Broadcom and Barefoot Docker containers

bocon13:
Adds 3 new labels to Bcm and BF Docker images:
```
    "org.opencontainers.image.revision": "5c0f2bcdd04822c68b7aeda61f8dbedbf3a61eab",
    "org.opencontainers.image.source": "git@github.com:stratum/stratum",
    "org.opencontainers.image.version": "5c0f2bcdd04822c68b7aeda61f8dbedbf3a61eab",
```

Note: Version can be a commit hash or git tag. Revision is always a commit hash.

Partially addresses #371

### DIFF
--- a/stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
+++ b/stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
@@ -133,6 +133,21 @@ if [ "$(docker version -f '{{.Server.Experimental}}')" = "true" ]; then
   DOCKER_BUILD_OPTS+="--squash "
 fi
 
+DOCKER_BUILD_OPTS+="--label stratum-target=$STRATUM_TARGET "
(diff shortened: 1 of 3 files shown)


pudelkoM (approved):
Given today's <s>shitshow</s> adventure around finding the origin of a mystery docker image, I'm willing to compromise on the final tag format or content and focus on getting anything out there first.

Additional thoughts:
Should we include the build timestamp and hostname, too?
A git commit hash answers "what is inside?", but not "when?" or "where from?".

bocon13:
@pudelkoM we could add two more:
```
--label build-timestamp=$(date)
--label build-machine=$(hostname)
```

charlesmcchan (approved):
"Revision" is sufficient to give us a definitive answer of what version it is. "Version" doesn't seem very useful at this moment since we don't really use tags today. But it won't hurt leaving it there for future use.

PR #437, opened 2020-11-19
Extend SdeWrapper

pudelkoM:
First stab at porting the fixed-function components to use SdeWrapper.

- [x] PacketIoManger
- [x] PreManager 
- [x] ActionProfileManager
- [x] IdMapper
- [x] TableManager
- [ ] Unit tests

Closes #338


stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -309,6 +319,295 @@ ::util::Status BfSdeWrapper::SetTmCpuPort(int device, int port) {
...
+
+  return ::util::OkStatus();
+}
+
+// Create and start an new session.
+::util::StatusOr<std::shared_ptr<BfSdeInterface::SessionInterface>>
+BfSdeWrapper::CreateSession() {
+  auto bfrt_session = bfrt::BfRtSession::sessionCreate();
+  CHECK_RETURN_IF_FALSE(bfrt_session) << "Failed to create new session.";
+
```
pudelkoM, line 414: remove

stratum/hal/lib/barefoot/bf_sde_interface.h
```diff
@@ -71,6 +91,86 @@ class BfSdeInterface {
...
+  virtual ::util::Status DeleteMulticastGroup(
+      int device, std::shared_ptr<BfSdeInterface::SessionInterface> session,
+      uint32 group_id) = 0;
+
+  // Returns the multicast group with the given ID ($pre.mgid table), or all
+  // groups if ID is 0.
+  virtual ::util::Status GetMulticastGroups(
+      int device, std::shared_ptr<BfSdeInterface::SessionInterface> session,
+      uint32 group_id, std::vector<uint32>* group_ids,
+      std::vector<std::vector<uint32>>* mc_node_ids) = 0;
```
pudelkoM, line 149: This function interface is not the greatest, but least bad alternative I could find.

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -309,6 +319,785 @@ ::util::Status BfSdeWrapper::SetTmCpuPort(int device, int port) {
...
+  RETURN_IF_BFRT_ERROR(table->dataAllocate(&table_data));
+  uint32 id = usage;
+  for (size_t _ = 0; _ < table_size; ++_) {
+    // Key: $MULTICAST_NODE_ID
+    RETURN_IF_ERROR(SetField(table_key.get(), kMcNodeId, id));
+    bf_status_t status = table->tableEntryGet(
+        *real_session->bfrt_session_, bf_dev_tgt, *table_key,
+        bfrt::BfRtTable::BfRtTableGetFlag::GET_FROM_SW, table_data.get());
+    if (status == BF_OBJECT_NOT_FOUND) {
+      return id;
```
Yi-Tseng, line 657: Is it possible that multiple threads calling this method and get the same ID?
Since we are using the reader lock

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -55,7 +54,20 @@ ::util::Status BfrtPacketioManager::PushForwardingPipelineConfig(
     absl::WriterMutexLock l(&data_lock_);
     RETURN_IF_ERROR(BuildMetadataMapping(program.p4info()));
     // PushForwardingPipelineConfig resets the bf_pkt driver.
-    RETURN_IF_ERROR(StartIo());
+    RETURN_IF_ERROR(bf_sde_interface_->StartPacketIo(device_));
+    if (!initialized_) {
+      packet_receive_channel_ = Channel<std::string>::Create(128);
```
Yi-Tseng, line 59: Use constant, maybe `kMaxPktRxChnl` or something else?

stratum/hal/lib/barefoot/bfrt_packetio_manager.h
```diff
@@ -135,12 +127,19 @@ class BfrtPacketioManager {
...
+  //
+  pthread_t sde_rx_thread_id_ GUARDED_BY(data_lock_);
+
+  // Pointer to a BfSdeInterface implementation that wraps all the SDE calls.
+  BfSdeInterface* bf_sde_interface_ = nullptr;  // not owned by this class.
 
   // Fixed zero-based Tofino device number corresponding to the node/ASIC
   // managed by this class instance. Assigned in the class constructor.
-  const int device_id_;
+  const int device_;
```
Yi-Tseng, line 142: Any reason you want to remove the `_id` from the variable name? (also the parameter of constructor)

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -271,36 +250,38 @@ ::util::Status BfrtPacketioManager::TransmitPacket(
...
+    if (code == ERR_CANCELLED) break;
+    if (code == ERR_ENTRY_NOT_FOUND) {
+      LOG(ERROR) << "Read with infinite timeout failed with ENTRY_NOT_FOUND.";
+      continue;
+    }
+
+    ::p4::v1::PacketIn packet_in;
+    RETURN_IF_ERROR(ParsePacketIn(buffer, &packet_in));
+    {
+      LOG(WARNING) << "Getting rx_writer_lock_ " << &rx_writer_lock_;
```
Yi-Tseng, line 279: Note: use VLOG for this logging before merging the PR

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -271,36 +250,38 @@ ::util::Status BfrtPacketioManager::TransmitPacket(
...
+      absl::WriterMutexLock l(&rx_writer_lock_);
+      rx_writer_->Write(packet_in);
```
Yi-Tseng, line 281: check if `rx_writer_` is nullptr or not first.

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
pudelkoM: yes.

I initially though about changing this function to `AllocateFree...` and reserve the ID right here. Later users would have to call modify, though. But even that would not guarantee uniqueness, just make it less likely.

Alternatives I see:

- BfRt Transactions, would require changing the batching logic. Can't have a transaction inside a batch.
- Exclusive locking, like you noted
- Busy trying. Basically retry the insert with a new ID when a EXISTS error is returned. At some point there should be no other threads.

Happy to hear other ideas.

stratum/hal/lib/barefoot/bfrt_packetio_manager.h
pudelkoM: More concise. For some types the `_id_` makes a difference: `node` (from common.proto) / `node_id`. Port is borderline. But for device I did not find a conflict, so I went for the same approach as with `unit`: https://github.com/stratum/stratum/search?q=unit_

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -115,6 +92,8 @@ ::util::Status BfrtPacketioManager::Shutdown() {
     packetout_header_.clear();
     packetin_header_size_ = 0;
     packetout_header_size_ = 0;
+    packet_receive_channel_->Close();
+    pthread_join(sde_rx_thread_id_, nullptr);
```
Yi-Tseng, line 96: There might be a potential deadlock here.
If you call this Shutdown right after the pipeline config gets pushed, the method will first hold `data_lock_` and the `HandleSdePacketRx` thread will wait for the lock at line 262. And `pthread_join` will never return and stuck.

Yi-Tseng: I think one way to solve this issue is to remove `pthread_join` from this block and use another with ReaderMutexLock since we are not modifying the thread id variable
for example:

```cpp
{
  absl::WriterMutexLock l(&data_lock_);
  packetin_header_.clear();
  packetout_header_.clear();
  packetin_header_size_ = 0;
  packetout_header_size_ = 0;
  packet_receive_channel_->Close();
  initialized_ = false;
}
{
  absl::ReaderMutexLock l(&data_lock_);
  pthread_join(sde_rx_thread_id_, nullptr);
}
```

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -92,9 +92,12 @@ ::util::Status BfrtPacketioManager::Shutdown() {
     packetin_header_size_ = 0;
     packetout_header_size_ = 0;
     packet_receive_channel_->Close();
-    pthread_join(sde_rx_thread_id_, nullptr);
     initialized_ = false;
   }
+  {
+    absl::ReaderMutexLock l(&data_lock_);
+    pthread_join(sde_rx_thread_id_, nullptr);
+  }
```
Yi-Tseng, line 100: To be discussed

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -271,36 +251,38 @@ ::util::Status BfrtPacketioManager::TransmitPacket(
...
+    std::string buffer;
+    int code = reader->Read(&buffer, absl::InfiniteDuration()).error_code();
+    if (code == ERR_CANCELLED) break;
+    if (code == ERR_ENTRY_NOT_FOUND) {
+      LOG(ERROR) << "Read with infinite timeout failed with ENTRY_NOT_FOUND.";
+      continue;
+    }
+
+    ::p4::v1::PacketIn packet_in;
+    RETURN_IF_ERROR(ParsePacketIn(buffer, &packet_in));
```
Yi-Tseng, line 278: I think we should just log the error and continue if we receive a bad packet (e.g., packet too small) since we still need to receive other packets.

pudelkoM: https://s3-us-west-2.amazonaws.com/p4runtime/docs/master/P4Runtime-Spec.html#sec-stream-error-reporting

stratum/hal/lib/barefoot/bf.proto
```diff
@@ -43,3 +43,12 @@ message BfrtDeviceConfig {
...
 }
+
+// Internal message for action data passed from and to the SdeWrapper.
+message BfActionData {
+  message Param {
+    int32 id = 1;
+    bytes value = 2;
+  }
+  repeated Param params = 1;
+}
```
pudelkoM, line 54: Decide if/how to bundle internal data structure for communication from/to the SdeWrapper and managers.

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -115,9 +95,17 @@ ::util::Status BfrtPacketioManager::Shutdown() {
     packetout_header_.clear();
     packetin_header_size_ = 0;
     packetout_header_size_ = 0;
+    if (packet_receive_channel_) {
+      packet_receive_channel_->Close();
```
Yi-Tseng, line 99: The reason I add this check is that the `packet_receive_channel_` will be null if we push an invalid p4info

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -115,9 +95,17 @@ ::util::Status BfrtPacketioManager::Shutdown() {
...
+    }
     initialized_ = false;
   }
-
+  {
+    absl::ReaderMutexLock l(&data_lock_);
+    if (rx_thread_initialized_) {
```
Yi-Tseng, line 105: I add a new variable called `rx_thread_initialized_`, the sde_rx_thread_id_` will be invalid if we push an invalid p4info/pipeline config. And `pthread_join` will cause the program to crash.
Not sure if we can solve this without adding this variable.

pudelkoM: Init `sde_rx_thread_id_` with 0 and check `if != 0` ?

pudelkoM: Also noticed that `sde_rx_thread_id_` is missing from the constructor init list.

pudelkoM: Another approach would be to spawn the thread at class creation, then `Shutdown` will never be called on failures. But that might not be logically possible.

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
pudelkoM: https://github.com/stratum/stratum/blob/70d1a0c869634eac0bd3722101ce25c5e82d924c/stratum/hal/lib/bcm/bcm_chassis_manager.cc#L1481-L1486

stratum/hal/lib/barefoot/bfrt_packetio_manager.h
```diff
@@ -133,6 +133,9 @@ class BfrtPacketioManager {
   //
   pthread_t sde_rx_thread_id_ GUARDED_BY(data_lock_);
```
pudelkoM, line 134: Add to constructor init list

stratum/hal/lib/barefoot/bf_sde_mock.h
```diff
@@ -140,12 +140,72 @@ class BfSdeMock : public BfSdeInterface {
                      absl::optional<uint64>* byte_count,
                      absl::optional<uint64>* packet_count,
                      absl::Duration timeout));
+
```
pudelkoM, line 143: No spaces here

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
Yi-Tseng: That would work, will do

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
Yi-Tseng: this is better, will do

stratum/hal/lib/barefoot/bf.proto
pudelkoM: Comments from the sync:

- Having new data types is fine, if it significantly simplifies API usage

- There is some precedence for this in the BCM code with `BcmFlowEntry`

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
pudelkoM: I think this was for debugging and can be removed now.

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -115,9 +94,24 @@ ::util::Status BfrtPacketioManager::Shutdown() {
...
-
+  {
+    absl::ReaderMutexLock l(&data_lock_);
+    if (sde_rx_thread_id_ != 0 &&
+        pthread_join(sde_rx_thread_id_, nullptr) != 0) {
+      ::util::Status error = MAKE_ERROR(ERR_INTERNAL)
+                             << "Failed to join thread " << sde_rx_thread_id_;
+      APPEND_STATUS_IF_ERROR(status, error);
+    }
+    sde_rx_thread_id_ = 0;
```
pudelkoM, line 113:
```
stratum/hal/lib/barefoot/bfrt_packetio_manager.cc:113:5: warning: writing variable 'sde_rx_thread_id_' requires holding mutex 'data_lock_' exclusively [-Wthread-safety-analysis]
    sde_rx_thread_id_ = 0;
    ^
```

pudelkoM:
@Yi-Tseng voiced some concerns about the current level of abstraction the wrapper provides. Rather than having specialized "fixed-function" style APIs (i.e. `InsertActionProfileMember`), we could expose a generic `InsertBfrtTableEntry()` function and leave the details to the caller.

PR #440, opened 2020-11-19
Use empty build context for setup_dev_env.sh

bocon13:
We don't use anything from the context, so this will save some time.


pudelkoM:
Does this prevent shipping the entire stratum directory to the docker daemon?

bocon13:
> Does this prevent shipping the entire stratum directory to the docker daemon?

Yes. We don't need to ship the Stratum directory at build time because it is volume mounted at runtime. :)

https://docs.docker.com/engine/reference/commandline/build/#text-files

pudelkoM (approved):

PR #441, opened 2020-11-19
Fix style in hal and credential manager

bocon13:
This PR fixes the code style in Hal and the CredentialsManager.

The flags `-grpc_max_recv_msg_size` and `-grpc_max_send_msg_size` are in bytes now.

### DIFF
--- a/stratum/hal/lib/common/hal.cc
+++ b/stratum/hal/lib/common/hal.cc
@@ -20,10 +20,9 @@
 #include "stratum/procmon/procmon.grpc.pb.h"
 
 // TODO(unknown): Use FLAG_DEFINE for all flags.
-DEFINE_string(
-    external_stratum_urls, stratum::kExternalStratumUrls,
-    "Comma-separated list of URLs for server to listen to for external"
-    " calls from SDN controller, etc.");
+DEFINE_string(external_stratum_urls, stratum::kExternalStratumUrls,
+              "Comma-separated list of URLs for server to listen to for "
+              "external calls from SDN controller, etc.");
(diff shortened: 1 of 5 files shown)


stratum/hal/lib/common/hal.cc
```diff
@@ -2,28 +2,28 @@
...
+#include "stratum/lib/utils.h"
+#include "stratum/procmon/procmon.grpc.pb.h"
 
 // TODO(unknown): Use FLAG_DEFINE for all flags.
-DEFINE_string(external_stratum_urls, stratum::kExternalStratumUrls,
-            "Comma-separated list of URLs for server to listen to for external"
-            " calls from SDN controller, etc.");
+DEFINE_string(
+    external_stratum_urls, stratum::kExternalStratumUrls,
+    "Comma-separated list of URLs for server to listen to for external"
```
pudelkoM, line 25: Break the string at `for` for a nicer line layout.

pudelkoM (dismissed):

stratum/hal/lib/common/hal.cc
```diff
@@ -198,15 +198,14 @@ ::util::Status Hal::Run() {
     for (const auto& url : external_stratum_urls) {
       builder.AddListeningPort(url, server_credentials);
     }
+    constexpr uint32 MB = 1024 * 1024;
```
pudelkoM, line 201: This change kinda indicates that the flag definition is bad.

I think it should be `grpc_max_recv_msg_size` and in bytes or `grpc_max_recv_msg_size_mb`.

pudelkoM (approved):

PR #445, opened 2020-11-19
Fix thread safety of bf, bmv2 and np4 ChassisManager

pudelkoM:
Uncovered by #306

### DIFF
--- a/stratum/hal/lib/bmv2/bmv2_chassis_manager.cc
+++ b/stratum/hal/lib/bmv2/bmv2_chassis_manager.cc
@@ -433,10 +433,12 @@ void Bmv2ChassisManager::ReadPortStatusChangeEvents() {
 
   port_status_change_event_channel_ =
       Channel<PortStatusChangeEvent>::Create(kMaxPortStatusChangeEventDepth);
-
-  port_status_change_event_writer_ =
-      ChannelWriter<PortStatusChangeEvent>::Create(
-          port_status_change_event_channel_);
+  {
+    absl::WriterMutexLock l(&port_status_change_event_writer_lock_);
+    port_status_change_event_writer_ =
+        ChannelWriter<PortStatusChangeEvent>::Create(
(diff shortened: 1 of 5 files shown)


stratum/hal/lib/bmv2/bmv2_chassis_manager.cc
```diff
@@ -521,9 +523,9 @@ ::util::Status Bmv2ChassisManager::Shutdown() {
   // It is fine to release the chassis lock here (it is actually needed to call
   // UnregisterEventWriters or there would be a deadlock). Because initialized_
   // is set to true, RegisterEventWriters cannot be called.
-  APPEND_STATUS_IF_ERROR(status, UnregisterEventWriters());
```
bocon13, line 524: Should we look at the comment above?

Also, if we do decide to keep the lock to unregister the event writer, then we should probably just merge this with the block above.

stratum/hal/lib/np4intel/np4_chassis_manager.cc
```diff
@@ -408,9 +407,9 @@ ::util::Status NP4ChassisManager::Shutdown() {
   // It is fine to release the chassis lock here (it is actually needed to call
   // UnregisterEventWriters or there would be a deadlock). Because initialized_
   // is set to true, RegisterEventWriters cannot be called.
-  APPEND_STATUS_IF_ERROR(status, UnregisterEventWriters());
```
bocon13, line 411: Same comment as bmv2

stratum/hal/lib/bmv2/bmv2_chassis_manager.cc
pudelkoM: mhh, good catch.

I'll have to the verify the lock usage again.

pudelkoM:
PR #306 added exclusion for these affected classes. Currently we do not observe any synchronization issues and postpone this PR to a later date.

PR #449, opened 2020-11-21
Add SignalWaiter utility class

bocon13:
This class is initialized statically on program start and registers handlers to listen for SIGINT, SIGTERM, and SIGUSR2.

Calls to WaitForSignal()  block until one of the signals is received.

Users can also call HandleSignal() to deliver a signal to waiters instead of kill() or pthread_kill().

TODO
- [ ] Consider unlocking sem_ in destructor
- [ ] Add signal callback register function that automatically spins up in a thread
- [ ] Should we detach threads when spinning up a waiter? Cleaning them up would happen automatically when they return.

### DIFF
--- a/stratum/lib/signal_waiter.cc
+++ b/stratum/lib/signal_waiter.cc
@@ -1,0 +1,72 @@
+// Copyright 2020-present Open Networking Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+#include "stratum/lib/signal_waiter.h"
+
+#include "stratum/glue/logging.h"
+#include "stratum/glue/status/status_macros.h"
+#include "stratum/public/lib/error.h"
+
+namespace stratum {
+namespace hal {
+
+namespace {
+
+// Signal received callback which is registered as the handler for signals using
+// signal() system call.
+void SignalRcvCallback(int value) { SignalWaiter::HandleSignal(value); }
+
+}  // namespace
+
+// Initialize static member.
+SignalWaiter SignalWaiter::instance_;
+
+SignalWaiter::SignalWaiter(const std::vector<int> signals) : signals_(signals) {
+  CHECK_ERR(sem_init(&sem_, 0, 0));
+  // Register the signal handlers and save the old handlers as well.
+  for (const int s : signals) {
+    sighandler_t h = signal(s, SignalRcvCallback);
+    if (h == SIG_ERR) {
+      LOG(FATAL) << "Failed to register signal: " << strsignal(s) << " (" << s
+                 << ").";
+    }
+    old_signal_handlers_[s] = h;
+  }
+}
+
+SignalWaiter::~SignalWaiter() {
+  // Register the old handlers for all the signals.
+  for (const auto& e : old_signal_handlers_) {
+    signal(e.first, e.second);
+  }
+  old_signal_handlers_.clear();
+
+  CHECK_ERR(sem_destroy(&sem_));
+}
+
+// Reminder: async-signal-safe function calls only!
+::util::Status SignalWaiter::HandleSignal(int value) {
+  const auto& s = instance_.signals_;
+  if (std::find(s.begin(), s.end(), value) == s.end()) {
+    RETURN_ERROR(ERR_INTERNAL)
+        << "Tried to handle unregistered signal: " << strsignal(value) << " ("
+        << value << ").";
+  }
+
+  // Wake up one thread waiting for a signal.
+  CHECK_ERR(sem_post(&instance_.sem_));
+  return ::util::OkStatus();
+}
+
+void SignalWaiter::WaitForSignal() {
+  while (sem_wait(&instance_.sem_) != 0) {
+    // Fail if not EINTR; otherwise continue waiting.
+    PCHECK(errno == EINTR);
+  }
+
+  // Wake up another thread that is waiting for a signal.
+  CHECK_ERR(sem_post(&instance_.sem_));
+}
+
+}  // namespace hal
+}  // namespace stratum
--- a/stratum/lib/signal_waiter.cc
+++ b/stratum/lib/signal_waiter.cc
@@ -39,7 +39,6 @@ SignalWaiter::~SignalWaiter() {
   for (const auto& e : old_signal_handlers_) {
     signal(e.first, e.second);
   }
-  old_signal_handlers_.clear();
 
   CHECK_ERR(sem_destroy(&sem_));
 }
--- a/stratum/lib/signal_waiter.h
+++ b/stratum/lib/signal_waiter.h
@@ -1,0 +1,67 @@
+// Copyright 2020-present Open Networking Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef STRATUM_LIB_SIGNAL_WAITER_H_
+#define STRATUM_LIB_SIGNAL_WAITER_H_
+
+#include <semaphore.h>
+#include <signal.h>
+
+#include <atomic>
+#include <vector>
+
+#include "absl/container/flat_hash_map.h"
+#include "stratum/glue/status/status.h"
+
+namespace stratum {
+namespace hal {
+
+// Class 'SignalWaiter' is a utility class that waits allows calling threads to
+// wait on a set of signals. This class is initialized statically on program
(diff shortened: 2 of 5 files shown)


stratum/lib/signal_waiter.cc
```diff
@@ -0,0 +1,72 @@
...
+    old_signal_handlers_[s] = h;
+  }
+}
+
+SignalWaiter::~SignalWaiter() {
+  // Register the old handlers for all the signals.
+  for (const auto& e : old_signal_handlers_) {
+    signal(e.first, e.second);
+  }
+  old_signal_handlers_.clear();
```
bocon13, line 42: About to be destroyed, so this is superfluous.
```suggestion
```

stratum/lib/signal_waiter_test.cc
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  void TearDown() override {
+    int sem_value = GetSemaphoreValue();
+    EXPECT_LE(1, sem_value);
+    LOG(INFO) << "Ended test with semaphore value: " << sem_value;
+    Reset();
+  }
+
+  void Reset() {
```
bocon13, line 35:
```suggestion
    ResetSem();
  }

  void ResetSem() {
```

bocon13:
Coverage should be 100%, but it looks like we don't get coverage data inside ASSERT_DEATH calls.

```
TEST_F(SignalWaiterTest, ConstructorError) {
  ::std::string error_msg("Failed to register signal: ");
  error_msg += strsignal(SIGKILL);
  ASSERT_DEATH(ConstructWaiter({SIGKILL}), error_msg);

  // Send a registered signal because the test framework expects one.
  EXPECT_OK(SignalWaiter::HandleSignal(SIGINT));
}
```

stratum/lib/signal_waiter_test.cc
```diff
@@ -0,0 +1,245 @@
...
+  EXPECT_EQ(sa1.sa_handler, sa2.sa_handler);
+
+  // Send a registered signal because the test framework expects one.
+  EXPECT_OK(SignalWaiter::HandleSignal(SIGINT));
+}
+
+TEST_F(SignalWaiterTest, ConstructorError) {
+  ::std::string error_msg("Failed to register signal: ");
+  error_msg += strsignal(SIGKILL);
+  ASSERT_DEATH(ConstructWaiter({SIGKILL}), error_msg);
```
bocon13, line 238:
```suggestion
  EXPECT_DEATH(ConstructWaiter({SIGKILL}), error_msg);
```

stratum/lib/signal_waiter_test.cc
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  void TearDown() override {
+    int sem_value = GetSemaphoreValue();
+    EXPECT_LE(1, sem_value);
+    LOG(INFO) << "Ended test with semaphore value: " << sem_value;
+    ResetSem();
+  }
+
+  void ResetSem() {
```
bocon13, line 35:
```suggestion
    ResetSemaphore();
  }

  void ResetSemaphore() {
```

stratum/lib/signal_waiter.h
```diff
@@ -0,0 +1,67 @@
...
+  // Blocking call to wait for one of the signals. Returns when one of the
+  // signals is received.
+  static void WaitForSignal();
+
+  // SignalWaiter is neither copyable nor movable.
+  SignalWaiter(const SignalWaiter&) = delete;
+  SignalWaiter& operator=(const SignalWaiter&) = delete;
+
+  // Static instance that gets created at program start.
+  static SignalWaiter instance_;
```
bocon13, line 40: Make this static const

stratum/lib/signal_waiter.h
```diff
@@ -0,0 +1,67 @@
...
+  ~SignalWaiter();
+
+  // Semaphore that is used to block threads that call WaitForSignal(). The
+  // semaphore is initialized locked (value 0). Waiting threads call sem_wait(),
+  // and the signal handler calls sem_post(). The semaphone remains unlocked
+  // (value >= 1) after a signal has been handled.
+  sem_t sem_;
+
+  // Vector of signals for which we registered handlers.
+  std::vector<int> signals_;
```
bocon13, line 54: const?

stratum/lib/signal_waiter.cc
```diff
@@ -0,0 +1,71 @@
...
+// Signal received callback which is registered as the handler for signals using
+// signal() system call.
+void SignalRcvCallback(int value) { SignalWaiter::HandleSignal(value); }
+
+}  // namespace
+
+// Initialize static member.
+SignalWaiter SignalWaiter::instance_;
+
+SignalWaiter::SignalWaiter(const std::vector<int> signals) : signals_(signals) {
```
bocon13, line 24: Make pattern, so that we can return a StatusOr

pudelkoM, line 24: Consider set to prevent duplicate signals and weird behavior after deconstruction.

pudelkoM:
After looking at this in more detail, some other points:

- This class is inherently not thread-safe and should say so in the docs

- Other classes must not use this class in their static initialization (ordering issue)

- If possible, the constructor should be as simple as possible; never throw or abort. 
   - Counter point: guaranteeing `sem_init` makes `WaitForSignal` safe to call independent of manual init.


Maybes:

- Explicit start and shutdown (`Activate()`, `Deactivate()`), instead of de/constructor, to signal clear usage. Could also make it more testable

- Decide if we want a barrier (unlock once) or a notification

- Is there a use case for having multiple threads waiting on `WaitForSignal`? Should we register functions instead?

stratum/lib/signal_waiter.cc
```diff
@@ -0,0 +1,71 @@
...
+  for (const auto& e : old_signal_handlers_) {
+    signal(e.first, e.second);
+  }
+
+  CHECK_ERR(sem_destroy(&sem_));
+}
+
+// Reminder: async-signal-safe function calls only!
+::util::Status SignalWaiter::HandleSignal(int value) {
+  const auto& s = instance_.signals_;
```
bocon13, line 48: If we want to log, use RAW_LOG

https://github.com/google/glog/blob/master/src/glog/raw_logging.h.in

stratum/lib/signal_waiter.cc
```diff
@@ -0,0 +1,71 @@
...
+  if (std::find(s.begin(), s.end(), value) == s.end()) {
+    RETURN_ERROR(ERR_INTERNAL)
```
bocon13, line 50: Check to make sure that this does not call new

pudelkoM: I do not have concrete evidence, as that depends on the implementation, but there is this note on the docs:
> If the algorithm fails to allocate memory, std::bad_alloc is thrown. 

https://en.cppreference.com/w/cpp/algorithm/find


Maybe this is another plus of using a set: https://en.cppreference.com/w/cpp/container/unordered_set/count

pudelkoM:
Check if we can use a function returning a pointer to static variable for the lock-free singleton requirement: https://youtu.be/xu7q8dGvuwk?t=2390

bocon13:
That’s actually pretty clever. I’ll try to see if I can refactor to use
that pattern.

stratum/lib/signal_waiter_test.cc
```diff
@@ -0,0 +1,245 @@
...
+  EXPECT_OK(SignalWaiter::HandleSignal(SIGINT));
+  SignalWaiter::WaitForSignal();
+  SignalWaiter::WaitForSignal();
+}
+
+TEST_F(SignalWaiterTest, DoubleSignalBeforeWait) {
+  EXPECT_OK(SignalWaiter::HandleSignal(SIGINT));
+  EXPECT_OK(SignalWaiter::HandleSignal(SIGINT));
+  SignalWaiter::WaitForSignal();
+  EXPECT_EQ(2, GetSemaphoreValue());
```
pudelkoM, line 66: We should not test the internal state, only the public API.

bocon13:
superseded by #652

pudelkoM:
We can revisit this if the async shutdown problem re-surfaces in other binaries.

PR #450, opened 2020-11-21
Fix gnoi dep for Mac OS

bocon13:


pudelkoM:
I had to make a similar sed fix before: https://github.com/stratum/stratum/commit/8d0bc6ce2d6d821b8bfddf40f9779264324f9657#diff-efd0b863ccd8180f488c2a50ccb629224973c43eeed0fffd7ccb9200d43bde86R15-R16

Not sure why I did not use `-i''`, but if that works we should apply it there too.

pudelkoM:
I'm actually still getting the error:

```
ERROR: /Users/max/projects/stratum/stratum/hal/lib/common/BUILD:23:19: //stratum/hal/lib/common:admin_service depends on @com_github_openconfig_gnoi//:system_cc_grpc in repository @com_github_openconfig_gnoi which failed to fetch. no such package '@com_github_openconfig_gnoi//': Error applying patch command find . -name *.proto | xargs sed -i 's#github.com/openconfig/##g':
sed: 1: "./interface/interface.p ...": invalid command code .
```

@Brian can you revisit or confirm?

pudelkoM (approved):

pudelkoM (approved):
Fixed other instances, too.

pudelkoM (approved):

bocon13:
@pudelkoM thanks for taking care of this

PR #456, opened 2020-12-04
Fix UNAUTHENTICATED not being treated as the highest error code

pudelkoM:
Also enable the test that will ensure correct translation in the future.


bocon13 (approved):
Thanks for the fix! Looks like UNAUTHENTICATED is now part of gRPC

PR #458, opened 2020-12-04
Fix bf.proto link in README.pipeline.md

pudelkoM:
Use absolute paths (but not http urls) for files in other directories.


bocon13 (approved):
lgtm

PR #459, opened 2020-12-05
Harmonize using path of test_utils in tests

pudelkoM:
Prefer using the relative path when already in the same namespace.

### DIFF
--- a/stratum/hal/lib/barefoot/bf_chassis_manager_test.cc
+++ b/stratum/hal/lib/barefoot/bf_chassis_manager_test.cc
@@ -17,7 +17,11 @@
 #include "stratum/lib/constants.h"
 #include "stratum/lib/test_utils/matchers.h"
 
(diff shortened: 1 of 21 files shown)


pudelkoM:
We will have to override the style check, because this PR touches all lot of unformatted test code and we don't want to address that issue here.

stratum/hal/lib/barefoot/bf_chassis_manager_test.cc
```diff
@@ -17,7 +17,11 @@
 #include "stratum/lib/constants.h"
 #include "stratum/lib/test_utils/matchers.h"
 
-using ::stratum::test_utils::EqualsProto;
+namespace stratum {
+namespace hal {
+namespace barefoot {
```
pudelkoM, line 22: Don't `using` include other namespaces into the global namespace: https://google.github.io/styleguide/cppguide.html#Namespaces

PR #469, opened 2020-12-18
Add fast path for stratum_bf container build

bocon13:


bocon13:
No longer needed. Just use the Bazel cache and the build is free. :)

PR #470, opened 2020-12-18
Deprecate bf_sim flag

bocon13:
TODO
- [ ] Remove includes and bazel deps
- [ ] Update docs
- [ ] Consider removing PhalSim altogether


bocon13:
Done in #743

PR #479, opened 2020-12-18
Fix priority and action param handling in BfrtTableManager::BuildP4TableEntry

pudelkoM:
When building the P4 table entry from BfRt keys/data, we must only read fields that are valid. This PR fixes this for the priority field and action parameters on indirect tables.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager.cc
@@ -380,6 +380,7 @@ struct RegisterClearThreadData {
                    p4_info_manager_->FindTableByID(request.table_id()));
   result.set_table_id(request.table_id());
 
+  bool has_priority_field = false;
   // Match keys
   for (const auto& expected_match_field : table.match_fields()) {
     ::p4::v1::FieldMatch match;  // Added to the entry later.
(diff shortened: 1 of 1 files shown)


stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -447,17 +451,20 @@ ::util::StatusOr<::p4::v1::TableEntry> BfrtTableManager::BuildP4TableEntry(
...
-  ASSIGN_OR_RETURN(auto action, p4_info_manager_->FindActionByID(action_id));
-  for (const auto& expected_param : action.params()) {
-    std::string value;
-    RETURN_IF_ERROR(table_data->GetParam(expected_param.id(), &value));
-    auto* param = result.mutable_action()->mutable_action()->add_params();
-    param->set_param_id(expected_param.id());
-    param->set_value(value);
+  if (action_id) {
+    ASSIGN_OR_RETURN(auto action, p4_info_manager_->FindActionByID(action_id));
+    result.mutable_action()->mutable_action()->set_action_id(action_id);
```
bocon13, line 457: Get action ID from the action object

bocon13 (approved):

PR #486, opened 2021-01-04
Demangle gNMI event names

### LINKED ISSUE #453: Use glog demangle to improve output of C++ mangled types in gNMI log
Demangle() definition: https://github.com/google/glog/blob/master/src/demangle.h

https://github.com/stratum/stratum/blob/d81b708113c54229245956762036d78af844ba9c/stratum/hal/lib/common/gnmi_events.h#L685

pudelkoM:
Closes #453


stratum/lib/utils.cc
```diff
@@ -236,4 +237,16 @@ size_t ProtoHash(const google::protobuf::Message& m) {
   return string_hasher(s);
 }
 
+std::string CxxDemangle(const char* mangled) {
```
pudelkoM, line 240:
```suggestion
std::string CxaDemangle(const char* mangled) {
```

We might call it just `Demangle`.

Yi-Tseng (approved):

bocon13 (approved):

PR #489, opened 2021-01-05
WIP of migration to bazel docker rules

pudelkoM:
This PR migrates await from `DOCKERFILE`s and bash scripts, to [rules_docker](https://github.com/bazelbuild/rules_docker).

These rules require access to a running docker daemon, because we have to run installation scripts (debs).

TODOs:

- [ ] Figure out how to make Docker-in-Docker work (`setup_dev_env.sh`)

Issue #475, opened 2020-12-18
Consider adding commit hash in the Stratum binary


bocon13:
Also maybe the most recent tag

PR #490, opened 2021-01-06
Add workspace status script

pudelkoM:
The workspace script collects information that can be used to stamp binaries and other Bazel artifacts in the future.
It extends the existing files `./bazel-out/stable-status.txt` and `./bazel-out/volatile-status.txt`.

```
BUILD_TIMESTAMP 1610129464
GIT_REF b56022e7692ecaca94c2f6cc69761fd9145fa6e0-dirty
GIT_SHA b56022e7692ecaca94c2f6cc69761fd9145fa6e0-dirty
GIT_URL https://github.com/stratum/stratum.git
```

Use cases could be PR #489 or issue #475

### DIFF
--- a/.circleci/workspace_status.sh
+++ b/.circleci/workspace_status.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+#
+# Copyright 2020-present Open Networking Foundation
+# SPDX-License-Identifier: Apache-2.0
+
+# This script will be run bazel when building process starts to
+# generate key-value information that represents the status of the
+# workspace. The output should be like
+#
+# KEY1 VALUE1
+# KEY2 VALUE2
+#
+# If the script exits with non-zero code, it's considered as a failure
+# and the output will be discarded.
+
+# The code below presents an implementation that works for git repository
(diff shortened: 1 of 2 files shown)


.circleci/workspace_status.sh
```diff
@@ -0,0 +1,56 @@
...
+# If the script exits with non-zero code, it's considered as a failure
+# and the output will be discarded.
+
+# The code below presents an implementation that works for git repository
+git_rev=$(git rev-parse HEAD)
+if [[ $? != 0 ]];
+then
+    exit 1
+fi
+echo "BUILD_SCM_REVISION ${git_rev}"
```
pudelkoM, line 22: Exact values tbd

Yi-Tseng:
can we use `set -e -o pipefail` so we don't have to check the result from every output?

pudelkoM:
> can we use `set -e -o pipefail` so we don't have to check the result from every output?

Definitely worth trying. The script is still a little WIP.

bocon13 (approved):
looks reasonable

PR #491, opened 2021-01-06
Pass error message from PI status to ::util::Status

pudelkoM:
Currently errors from PI look mostly like this:

```
E20210106 08:49:35.132390   122 bf_switch.cc:135] Return Error: pi_node->PushForwardingPipelineConfig(config) failed with generic::unknown:
E20210106 08:49:35.133188   122 p4_service.cc:411] generic::unknown: Error without message at stratum/hal/lib/common/p4_service.cc:411
```

This is because PI uses UNKNOWN error codes instead of the more descriptive ones: https://github.com/p4lang/PI/blob/e0d7e1f43cf1593abbd2b6e6c46285309d4f3870/proto/frontend/src/device_mgr.cpp#L80-L82
The returned errors still contain a message string, but we drop that and replace it with an empty one in the conversion (`toUtilStatus`). This PR addresses the message.

### DIFF
--- a/stratum/hal/lib/pi/pi_node.cc
+++ b/stratum/hal/lib/pi/pi_node.cc
@@ -35,7 +35,7 @@ namespace {
     if (results->size() != 0)
       return MAKE_ERROR(ERR_INTERNAL) << "Expected empty results vector.";
     results->resize(updates_size);
-    return ::util::Status();
+    return ::util::OkStatus();
   }
   ::util::Status status(::util::Status::canonical_space(), from.code(),
                         from.message());
@@ -50,10 +50,11 @@ namespace {
 
 ::util::Status toUtilStatus(const DeviceMgr::Status& from) {
   if (from.code() == Code::OK) {
-    return ::util::Status();
+    return ::util::OkStatus();
+  } else {
+    return ::util::Status(::util::Status::canonical_space(), from.code(),
+                          from.message());
   }
-  return ::util::Status(::util::Status::canonical_space(), from.code(),
-                        from.message());
 }
 
 }  // namespace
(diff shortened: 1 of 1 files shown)


stratum/hal/lib/pi/pi_node.cc
```diff
@@ -35,9 +35,10 @@ ::util::Status toUtilStatus(const DeviceMgr::Status& from,
     if (results->size() != 0)
       return MAKE_ERROR(ERR_INTERNAL) << "Expected empty results vector.";
     results->resize(updates_size);
-    return ::util::Status();
+    return ::util::OkStatus();
   }
-  ::util::Status status(::util::Status::canonical_space(), from.code(), "");
+  ::util::Status status(::util::Status::canonical_space(), from.code(),
+                        from.message());
```
Yi-Tseng, line 41:
```suggestion
  ::util::Status status(from.code(), from.message());
```

pudelkoM: Good idea, but the compiler disagrees:

```
stratum/hal/lib/pi/pi_node.cc:40:18: error: no matching constructor for initialization of '::util::Status'
  ::util::Status status(from.code(), from.message());
                 ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~
./stratum/glue/status/status.h:71:3: note: candidate constructor not viable: no known conversion from '::google::protobuf::int32' (aka 'int') to '::util::error::Code' for 1st argument
  Status(::util::error::Code code, const std::string& error_message);
  ^
```

Yi-Tseng: ah, I forgot that is the enum. the current code is good enough I think

Yi-Tseng (approved):

PR #496, opened 2021-01-07
Only read action profile group / member fields when active and on indirect table entries

pudelkoM:
Currently we opportunistically read both the `$SELECTOR_GROUP_ID` and `$ACTION_MEMBER_ID` field in returned action data on P4 table reads. The SDE will happily return values for both (i.e. 0), even when one field was never set. This means only action profile groups worked.
This PR addresses this issue by checking the `isActive` flag for these oneof fields, returning `NOT_FOUND` when appropriate.

This change also gets rid of the log spam when reading regular table entries:
`E20210107 00:20:27.048781   106 bf_sde_wrapper.cc:96] StratumErrorSpace::ERR_ENTRY_NOT_FOUND: 'table->dataFieldIdGet(field_name, action_id, &field_id)' failed with error message: Object not found.`

Example:

```python
# create member and group
am = action_profile_member["FabricIngress.next.hashed_profile"]
am.member_id = 1
am.action = Action("FabricIngress.next.output_hashed")
am.action["port_num"] = "1"
am.insert()
am.read(lambda e: print(e))
ag = action_profile_group["FabricIngress.next.hashed_profile"](group_id=2)
ag.add(1)
ag.max_size = 16
ag.insert()
# table entry to member
te = table_entry["FabricIngress.next.hashed"](action_profile_member=1)
te.match["next_id"] = "1"
te.member_id = 1
te.insert()
te.read(lambda e: print(e))
# table entry to group
te = table_entry["FabricIngress.next.hashed"](action_profile_group=2)
te.match["next_id"] = "2"
te.group_id = 2
te.insert()
te.read(lambda e: print(e))
# regular direct table entry
te = table_entry["FabricIngress.acl.acl"](action="FabricIngress.acl.drop")
te.match['eth_type'] = "0x800"
te.priority = 10
te.counter_data.byte_count = 100
te.counter_data.packet_count = 2
te.insert()
te.read(lambda e: print(e))

param_id: 1
value: "\000\001"

action_profile_id: 285274751 ("FabricIngress.next.hashed_profile")
member_id: 1
action {
  action_id: 16815357 ("FabricIngress.next.output_hashed")
  params {
    param_id: 1 ("port_num")
    value: "\\x00\\x01"
  }
}

field_id: 1
exact {
  value: "\000\000\000\001"
}

table_id: 33577058 ("FabricIngress.next.hashed")
match {
  field_id: 1 ("next_id")
  exact {
    value: "\\x00\\x00\\x00\\x01"
  }
}
action {
  action_profile_member_id: 1
}
```


stratum/hal/lib/barefoot/macros.h
```diff
@@ -79,6 +79,14 @@ class BooleanBfStatus {
...
            << FixMessage(bf_err_str(__ret.status()))
 
+// Same as RETURN_IF_BFRT_ERROR but does not write to LOG(ERROR).
+#define RETURN_IF_BFRT_ERROR_WITHOUT_LOGGING(expr)            \
+  if (const BooleanBfStatus __ret = BooleanBfStatus(expr)) {  \
+  } else /* NOLINT */                                         \
+    return MAKE_ERROR(__ret.error_code()).without_logging()   \
+           << "'" << #expr << "' failed with error message: " \
+           << FixMessage(bf_err_str(__ret.status()))
+
```
pudelkoM, line 89: The duplication here is a little unfortunate, maybe you guys have a better idea.

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -562,15 +567,68 @@ ::util::Status TableData::SetActionMemberId(uint64 action_member_id) {
...
+  bfrt::DataType data_type;
+  if (table->actionIdApplicable()) {
+    bf_rt_id_t action_id;
+    RETURN_IF_BFRT_ERROR(table_data_->actionIdGet(&action_id));
+    RETURN_IF_BFRT_ERROR_WITHOUT_LOGGING(
+        table->dataFieldIdGet("$ACTION_MEMBER_ID", action_id, &field_id));
+    RETURN_IF_BFRT_ERROR(
+        table->dataFieldDataTypeGet(field_id, action_id, &data_type));
+  } else {
+    RETURN_IF_BFRT_ERROR(table->dataFieldIdGet("$ACTION_MEMBER_ID", &field_id));
```
pudelkoM, line 582: Should be `RETURN_IF_BFRT_ERROR_WITHOUT_LOGGING`?

stratum/hal/lib/barefoot/macros.h
Yi-Tseng: Since we will print the error message in the P4 service, can we just ignore logging by default?

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
pudelkoM: This is fine because if the table as no action IDs, it's usually an indirect one.

stratum/hal/lib/barefoot/macros.h
pudelkoM: I think sometimes the full stack trace is nice to have in the logs.

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -562,15 +567,68 @@ ::util::Status TableData::SetActionMemberId(uint64 action_member_id) {
...
-  return GetField(*table_data_, "$ACTION_MEMBER_ID", action_member_id);
+  const bfrt::BfRtTable* table;
+  RETURN_IF_BFRT_ERROR(table_data_->getParent(&table));
+  bf_rt_id_t field_id;
+  bfrt::DataType data_type;
+  if (table->actionIdApplicable()) {
+    bf_rt_id_t action_id;
+    RETURN_IF_BFRT_ERROR(table_data_->actionIdGet(&action_id));
+    RETURN_IF_BFRT_ERROR_WITHOUT_LOGGING(
+        table->dataFieldIdGet("$ACTION_MEMBER_ID", action_id, &field_id));
```
Yi-Tseng, line 578: Maybe I understand incorrectly
Is it possible to combine an action + a member?
In P4Runtime spec, we cannot combine action + action member in a table entry

pudelkoM: I don't think so. Maybe we can use `actionIdApplicable()` has the first check.

Yi-Tseng: I think the `is_active` below will be false if the table uses direct action(normal action) instead of indirect action(action member)

Yi-Tseng: But we can still check by using `actionIdApplicable` and return earlier

stratum/hal/lib/barefoot/macros.h
Yi-Tseng: So far I don't have any other idea to reduce the code duplication here

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
pudelkoM: > But we can still check by using `actionIdApplicable` and return earlier

See latest change. This seems to work just fine.

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -567,15 +567,63 @@ ::util::Status TableData::SetActionMemberId(uint64 action_member_id) {
...
 ::util::Status TableData::GetActionMemberId(uint64* action_member_id) const {
-  return GetField(*table_data_, "$ACTION_MEMBER_ID", action_member_id);
+  const bfrt::BfRtTable* table;
+  RETURN_IF_BFRT_ERROR(table_data_->getParent(&table));
+  // Here we assume that table entries with action IDs (direct match-action) can
+  // never hold action member or group IDs (indirect match-action). Since this
+  // function is regularly called on both, we do not log this error here.
+  if (table->actionIdApplicable()) {
+    return MAKE_ERROR(ERR_ENTRY_NOT_FOUND).without_logging()
+           << "Table data does not contain an action member ID.";
```
Yi-Tseng, line 577: Since we are checking the "table" in the `if` condition, maybe it can be something like "Table does not contain an action member ID field."? same comment for the action group one

Yi-Tseng (approved):
One comment about the message, but LGTM overall

Yi-Tseng (approved):

PR #498, opened 2021-01-08
Add tool to canonicalize Tofino port ids and apply it

### LINKED ISSUE #379: Update default port maps to use friendly port ids

pudelkoM:
This PR canonicalizes the Tofino port IDs in our chassis configs to match the port name and port field (`"10/0"` -> `10`).
It also adds a tool to do this in an automated fashion:

```bash
ls -1 stratum/hal/config/*/chassis_config.pb.txt | \
  xargs -n 1 bazel run //stratum/hal/config:chassis_config_migrator -- \
    -chassis_config_file
```

TODOs:

- [x] Try loading the configs once
- [x] ~~Move code to tools~~ We don't want to make this too official
- [x] Publish docker container

Closes #379

### DIFF
--- a/stratum/hal/config/tofino_chassis_config_migrator.cc
+++ b/stratum/hal/config/tofino_chassis_config_migrator.cc
@@ -1,0 +1,73 @@
+// Copyright 2021-present Open Networking Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+#include <string>
+
+#include "absl/strings/numbers.h"
+#include "absl/strings/str_split.h"
+#include "gflags/gflags.h"
+#include "stratum/glue/init_google.h"
+#include "stratum/glue/logging.h"
+#include "stratum/hal/lib/common/common.pb.h"
+#include "stratum/lib/macros.h"
+#include "stratum/lib/utils.h"
+
+DEFINE_string(chassis_config_file, "",
+              "Path to chassis configuration to migrate.");
+
+namespace stratum {
+namespace hal {
+namespace config {
+
+const char kUsage[] =
+    R"USAGE(usage: --chassis_config_file=<path>
+
+Chassis configuration migration tool.
+
+Combine with xargs for mass migration:
+ls -1 stratum/hal/config/*/chassis_config.pb.txt | \
+  xargs -n 1 bazel run //stratum/hal/config:tofino_chassis_config_migrator -- \
+    -chassis_config_file
+)USAGE";
+
+::util::Status Main(int argc, char** argv) {
+  ::gflags::SetUsageMessage(kUsage);
+  InitGoogle(argv[0], &argc, &argv, true);
+  stratum::InitStratumLogging();
+
+  CHECK_RETURN_IF_FALSE(!FLAGS_chassis_config_file.empty())
+      << "No chassis config given.";
+  ChassisConfig config;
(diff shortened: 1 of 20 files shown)


stratum/hal/config/x86-64-accton-wedge100bf-32qs-r0/chassis_config.pb.txt
```diff
@@ -373,7 +373,7 @@ singleton_ports {
   node: 1
 }
 singleton_ports {
-  id: 66
+  id: 33
```
Yi-Tseng, line 376: duplicated

pudelkoM: Good catch.

Should we remove the whole port? Everything but the name (`33/2`) is the same.

Yi-Tseng: the channel is different, these two ports are ports connected to the CPU, you can remove them(or one of them) if you like.

stratum/hal/config/tofino_chassis_config_migrator.cc
```diff
@@ -0,0 +1,91 @@
...
+      << "No chassis config given.";
+  ChassisConfig config;
+  RETURN_IF_ERROR(ReadProtoFromTextFile(FLAGS_chassis_config_file, &config));
+  if (config.chassis().platform() != PLT_GENERIC_BAREFOOT_TOFINO &&
+      config.chassis().platform() != PLT_GENERIC_BAREFOOT_TOFINO2) {
+    RETURN_ERROR(ERR_INVALID_PARAM)
+        << "Chassis config is not for Tofino platform";
+  }
+  for (auto& singleton_port : *config.mutable_singleton_ports()) {
+    std::vector<std::string> parts = absl::StrSplit(singleton_port.name(), '/');
```
pudelkoM, line 49: This takes the port name as source of truth. We could also use the slot/port/channel combo, as it kinda must be correct.

Yi-Tseng: Can we use the value from the `port` and `channel` fields in the `singleton_ports` message instead of parsing the `name` field?

pudelkoM: Changed to code to do that.

stratum/hal/config/x86-64-delta-ag9064v1-r0/chassis_config.pb.txt
```diff
@@ -9,714 +9,714 @@ nodes {
   index: 1
 }
 singleton_ports {
-  id: 48
+  id: 1
   name: "1/0"
   slot: 1
-  port: 48
+  port: 1
```
pudelkoM, line 15: @Yi-Tseng I think this port value is incorrect and does not work with the current Stratum logic. Can you take a look at this config?

Yi-Tseng: you are right, the `port` should be the front panel port number

pudelkoM: Thanks for confirming!

PR #500, opened 2021-01-09
Stamp binaries with version information

pudelkoM:
Findings so far:

Variables can be added to `bazel-out/stable-status.txt` by naming them `STABLE_FOO`.

Changes to `bazel-out/volatile-status.txt` do NOT trigger rebuilds by design. We can add a timestamp to `bazel-out/stable-status.txt`, but that seems to be against the intended purpose.

Linkstamping on `cc_library` only works with `BUILD_SCM_REVISION` and `BUILD_TIMESTAMP`.

Linkstamping can be controlled on the `cc_binary` with `stamp = 1` or on the CLI (`--stamp`), but only affects linkstamping variables:
```
> bazel run //stratum/lib:stamping_example_main
kBuildTimestamp: 1610156461
kBuildTimestamp2: 0
kBuildTimestampHuman: 2021-01-08T17:41:01-0800
kGitRef: 403c08bd1191737f8696b8a939a0cc5567828ba9-dirty
kBuildScmRevision: 0

> bazel run --stamp //stratum/lib:stamping_example_main
kBuildTimestamp: 1610156456
kBuildTimestamp2: 1610156456
kBuildTimestampHuman: 2021-01-08T17:40:56-0800
kGitRef: 403c08bd1191737f8696b8a939a0cc5567828ba9-dirty
kBuildScmRevision: 403c08bd1191737f8696b8a939a0cc5567828ba9
```

Other, unrelated targets seem to unaffected by stamping (and are thus cached), but we need to confirm this!

Links:
https://groups.google.com/g/bazel-discuss/c/9kqLmOq_m_U
https://stackoverflow.com/a/47630130
https://github.com/bazelbuild/bazel/issues/1758
https://github.com/envoyproxy/envoy/blob/master/source/common/version/BUILD#L13-L47
https://github.com/bazelbuild/bazel/blob/082d58de852ebaa640bcf13cf419cbb94eec2b26/src/main/java/com/google/devtools/build/lib/bazel/BazelWorkspaceStatusModule.java#L316
https://github.com/bazelbuild/bazel/issues/4863

### DIFF
--- a/.circleci/workspace_status.sh
+++ b/.circleci/workspace_status.sh
@@ -10,6 +10,9 @@
 # KEY1 VALUE1
 # KEY2 VALUE2
 #
+# Keys starting with "STABLE_" will go into the stable status file. Others in
+# the volatile.
+#
 # If the script exits with non-zero code, it's considered as a failure
 # and the output will be discarded.
 set -e -o pipefail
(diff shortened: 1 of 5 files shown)


pudelkoM:
Call summary:

- proceed with linkstamping as it seems to do the right thing in regards to caching
- restrict ourselves to Bazel defined variables (`BUILD_*`) which are good enough
- integrate version string with `gflags::SetVersionString` for `--version`
- print version on startup

bocon13:
see #516

PR #506, opened 2021-01-14
Create codeql-analysis.yml

bocon13:


bocon13:
Need to run the Bazel build: https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/troubleshooting-the-codeql-workflow

PR #509, opened 2021-01-14
Pass given initial pipeline in Mininet StratumBmv2Switch

davidgengenbach:
The default `STRATUM_INIT_PIPELINE` was used, even when passing a custom initial pipeline in the constructor.


Yi-Tseng (approved):
LGTM

PR #511, opened 2021-01-15
Add TAI as Debian package dependency, if enabled

pudelkoM:
I think this commit removed it accidentally: https://github.com/stratum/stratum/commit/ff762d8cec73ac6f779b8f230088348030c653be#diff-4a60e1cf5aa2f74e522a0b1b161ba8ae7d4296a92751ddc63ffe74dc34eeff9d

### DIFF
--- a/stratum/hal/bin/bcm/standalone/BUILD
+++ b/stratum/hal/bin/bcm/standalone/BUILD
@@ -166,7 +166,10 @@ pkg_deb(
         "libedit2",
         "libjudydebian1",
         "libssl1.1",
-    ],
+    ] + select({
+        "//stratum/hal/lib/phal:with_tai": ["cassini-tai"],
+        "//conditions:default": [],
+    }),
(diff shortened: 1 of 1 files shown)


Yi-Tseng:
Can we enable it by default and let the phal decides to use it or not? (like ONLP Phal)

Yi-Tseng (approved):

pudelkoM:
> Can we enable it by default and let the phal decides to use it or not? (like ONLP Phal)
> We can do that in a different patch

I would like that, unified build and everything. But I think the grpc connect will not fail on setup, but on first call. We'll have to think about this more.

Yi-Tseng:
> > Can we enable it by default and let the phal decides to use it or not? (like ONLP Phal)
> > We can do that in a different patch
> 
> I would like that, unified build and everything. But I think the grpc connect will not fail on setup, but on first call. We'll have to think about this more.

Right, we may want to send a dummy message to check the connection

PR #512, opened 2021-01-15
Compress the Debian packages

pudelkoM:
We should compress the files when packing them into Debian packages to reduce file and container sizes.

Bz2 seems to sit in the sweet spot of speed and compression (19%). Keep in mind that this workflow is also used when doing rapid development, thus should be fast. Therefore xz (13%) is not chosen.

Anecdata:

```
0.36 ms    74M Jan 15 04:22 bazel-bin/stratum/hal/bin/barefoot/stratum_bf_data.tar
5.727 s    14M Jan 15 04:23 bazel-bin/stratum/hal/bin/barefoot/stratum_bf_data.tar.bz2 
9.993 s    15M Jan 15 04:23 bazel-bin/stratum/hal/bin/barefoot/stratum_bf_data.tar.gz
23.546 s  9.4M Jan 15 04:21 bazel-bin/stratum/hal/bin/barefoot/stratum_bf_data.tar.xz
```

### DIFF
--- a/stratum/p4c_backends/fpm/BUILD
+++ b/stratum/p4c_backends/fpm/BUILD
@@ -1926,6 +1926,7 @@ pkg_tar(
 
 pkg_tar(
     name = "fpm_p4c_data",
+    extension = "tar.xz",
     deps = [
         ":fpm_p4c_bin",
         ":fpm_p4c_includes",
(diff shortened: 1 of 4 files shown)


pudelkoM:
@bocon13 you may remember the file size increase we observed between two releases. I certain we had similar options enabled before, but they got removed by accident.

stratum/p4c_backends/fpm/BUILD
```diff
@@ -1926,6 +1926,7 @@ pkg_tar(
 
 pkg_tar(
     name = "fpm_p4c_data",
+    extension = "tar.xz",
```
pudelkoM, line 1929: We "never" build this by hand, therefore I'm willing to spend more time here.

PR #516, opened 2021-01-19
Stamp Stratum release binaries

pudelkoM:
Print a INFO log message on startup with the stamping information like commit, user, build host and timestamp. Also sets the version for gflags `--version`.

Stamping is only enabled for release builds, but disabled for tests to allow better caching.

This change automatically applies to all binaries calling `InitStratumLogging`, which includes all of our tools as well.

### Examples
```
# start-stratum.sh -v=1 ...

bf_kdrv_mod found! Unloading first...
loading bf_kdrv_mod...
I20210119 20:36:00.767335 13548 logging.cc:62] Stratum version 55d522f799c379fe4409a4813a9f0ecf5f63f12d built at 2021-07-15T15:31:20-07:00 on host menlo-pdp-lotta-nics by user max.
bf_sysfs_fname /sys/class/bf/bf0/device/dev_add
...
```

```
# start-stratum.sh --version
bf_kdrv_mod found! Unloading first...
loading bf_kdrv_mod...
stratum_bf version 51a48a1a24d76e2ebe6ea543c664289f314260e4
```

### Known issues

Our CI process modifies the git tree by patching the bazel config. This means all published build artifacts will report `(Modified)` in their version. Since having any version information right now is more valuable than 100% correct info, this is an acceptable tradeoff and can be addressed in a later issue.

Closes #500

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -5,6 +5,7 @@
 build --conlyopt='-std=gnu11'
 build --cxxopt='-std=gnu++11'
 build --workspace_status_command=.circleci/workspace_status.sh
+build --stamp
 
 # General test options
 test --test_output=errors
@@ -47,4 +48,3 @@ build:tsan --strip=never
 # Release build (build with -O2, no asserts, and strip all symbols)
 build:release --compilation_mode=opt
 build:release --linkopt=-Wl,--strip-all
-build:release --stamp
--- a/.bazelrc
+++ b/.bazelrc
@@ -47,3 +47,4 @@ build:tsan --strip=never
 # Release build (build with -O2, no asserts, and strip all symbols)
 build:release --compilation_mode=opt
 build:release --linkopt=-Wl,--strip-all
+build:release --stamp
--- a/.bazelrc
+++ b/.bazelrc
@@ -5,8 +5,6 @@
 build --conlyopt='-std=gnu11'
 build --cxxopt='-std=gnu++11'
 build --workspace_status_command=.circleci/workspace_status.sh
-build --stamp
-test --nostamp
 
 # General test options
 test --test_output=errors
@@ -46,6 +44,7 @@ build:tsan --copt -g
 build:tsan --copt -fno-omit-frame-pointer
 build:tsan --strip=never
 
-# Release build (build with -O2, no asserts, and strip all symbols)
+# Release build (build with -O2, no asserts, strip all symbols, and stamp)
 build:release --compilation_mode=opt
 build:release --linkopt=-Wl,--strip-all
+build:release --stamp
--- a/.bazelrc
+++ b/.bazelrc
@@ -6,6 +6,7 @@ build --conlyopt='-std=gnu11'
 build --cxxopt='-std=gnu++11'
 build --workspace_status_command=.circleci/workspace_status.sh
 build --stamp
+test --nostamp
 
 # General test options
 test --test_output=errors
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,6 +95,8 @@ jobs:
       - run:
           name: Test Stratum targets
           command: xargs -a .circleci/test-targets.txt bazel test
+      - run: git status --porcelain
+      # - run: bazel run //stratum/hal/bin/bcm/standalone:stratum_bcm_sdklt
       - *clean_bazel_cache
       - *save_bazel_cache
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,8 +93,6 @@ jobs:
       - run:
           name: Test Stratum targets
           command: xargs -a .circleci/test-targets.txt bazel test
-      - run: git status --porcelain
-      # - run: bazel run //stratum/hal/bin/bcm/standalone:stratum_bcm_sdklt
       - *clean_bazel_cache
       - *save_bazel_cache
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,9 +45,7 @@ store_bazel_profile: &store_bazel_profile
 set_bazelrc: &set_bazelrc
   run:
     name: Set bazelrc
-    command: |
-      cat .circleci/bazelrc >> .bazelrc
-      echo .bazelrc >> .gitignore
+    command: cat .circleci/bazelrc >> .bazelrc
 
 # Convenience anchors to update published Docker images. Images are first pulled
 # to allow for layer cache hits and reduce build times.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,9 @@ store_bazel_profile: &store_bazel_profile
 set_bazelrc: &set_bazelrc
   run:
     name: Set bazelrc
-    command: cat .circleci/bazelrc >> .bazelrc
+    command: |
+      cat .circleci/bazelrc >> .bazelrc
+      echo .bazelrc >> .gitignore
 
 # Convenience anchors to update published Docker images. Images are first pulled
 # to allow for layer cache hits and reduce build times.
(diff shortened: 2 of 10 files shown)


stratum/glue/logging.cc
```diff
@@ -57,6 +58,11 @@ void InitStratumLogging() {
   if (FLAGS_logtostderr) {
     LogToStderr();
   }
+
+  LOG(INFO) << "Stratum version " << kBuildScmRevision << " ("
+            << kBuildScmStatus << ")"
+            << " built at " << kBuildTimestamp << " on host " << kBuildHost
```
bocon13, line 64: Consider a more human-readable timestamp; for example:
```suggestion
            << " built at " << absl::FromTimeT(kBuildTimestamp) << " on host " << kBuildHost
```

bocon13: And perhaps make this part of the log message conditional if `kBuildTimestamp > 0`

.circleci/workspace_status.sh
```diff
@@ -25,3 +28,15 @@ echo "GIT_SHA ${git_sha}"
 # Tag name, or GIT_SHA if not on a tag
 git_ref=$(git describe --tags --no-match --always --abbrev=40 --dirty | sed -E 's/^.*-g([0-9a-f]{40}-?.*)$/\1/')
 echo "GIT_REF ${git_ref}"
```
bocon13, line 30: `GIT_*` variable seem like they are no longer necessary. Doesn't hurt to keep them for potential future use, but up to you.

stratum/lib/BUILD
```diff
@@ -7,6 +7,7 @@ licenses(["notice"])  # Apache v2
 load(
     "//bazel:rules.bzl",
     "stratum_cc_library",
+    "stratum_cc_binary",
```
bocon13, line 10: This may not be needed
```suggestion
```

stratum/glue/BUILD
```diff
@@ -57,3 +59,9 @@ stratum_cc_library(
         "integral_types.h",
     ],
 )
+
+cc_library(
+    name = "stamping",
```
bocon13, line 64: Nit: I think `version_stamp.h` and `version_stamp.cc` might be a bit more descriptive.
Or, `version.h` and `version_linkstamp.cc`

What do you think?

stratum/glue/logging.cc
pudelkoM: Good idea, I had this in mind when writing the code but must have eluded me at some point.
For the conditional part a ternary operator expression could do it.

stratum/glue/BUILD
pudelkoM: I had `version.h` in the draft, but though that's too limited because we also do time, host and user besides git ref (what I consider our version).

No strong feelings either way, though.

.circleci/workspace_status.sh
```diff
@@ -25,3 +28,15 @@ echo "GIT_SHA ${git_sha}"
...
+
+# Plain git revision for linkstamping.
+echo "BUILD_SCM_REVISION" $(git rev-parse HEAD)
```
pudelkoM, line 33: Since this is just a string, should we consider the more advanced form from `GIT_REF`?

This is just the Bazel default.

.circleci/workspace_status.sh
pudelkoM: Agree, but see next comment.

Should we follow through with Bazel docker, we could use them again.

.circleci/workspace_status.sh
```diff
@@ -25,3 +28,15 @@ echo "GIT_SHA ${git_sha}"
...
+
+# Git tree status for linkstamping.
+if [[ -n $(git status --porcelain) ]];
+then
+    tree_status="Modified"
```
Yi-Tseng, line 38: CI will apply the bazelrc first, this will always be modified when we build the container image form CI

Yi-Tseng: maybe check `$CI` or `$CIRCLE_CI` ?

stratum/glue/logging.cc
pudelkoM: Here is how a non-stamped build looks like:

`I20210120 23:16:39.726083 12592 logging.cc:63] Stratum version 0 (redacted) built at 1970-01-01T00:00:00+00:00 on host redacted by user redacted.`

Just ommiting the ts does not change much. Do we just omit the whole message? I'd actually like to keep it, as it's very obviously not an official release.

stratum/glue/logging.cc
```diff
@@ -57,6 +59,11 @@ void InitStratumLogging() {
   if (FLAGS_logtostderr) {
     LogToStderr();
   }
+
+  LOG(INFO) << "Stratum version " << kBuildScmRevision << " ("
+            << kBuildScmStatus << ")"
+            << " built at " << absl::FromTimeT(kBuildTimestamp) << " on host "
+            << kBuildHost << " by user " << kBuildUser << ".";
```
pudelkoM, line 66: Recent version of glog added `SetApplicationFingerprint`, check if it can replace this.

.circleci/workspace_status.sh
bocon13: pass `.bazelrc` as a flag instead of relying on default?

bocon13: Could add composite file to .gitignore or put it in /tmp

stratum/glue/logging.cc
bocon13: `Stratum version - development (unstamped)`

bocon13: "0" is better than "1970-01-01T00:00:00+00:00"

bocon13: or turn on stamping by default :)

.circleci/config.yml
```diff
@@ -45,7 +45,9 @@ store_bazel_profile: &store_bazel_profile
 set_bazelrc: &set_bazelrc
   run:
     name: Set bazelrc
-    command: cat .circleci/bazelrc >> .bazelrc
+    command: |
+      cat .circleci/bazelrc >> .bazelrc
+      echo .bazelrc >> .gitignore
```
pudelkoM, line 50: If this work, add a comment about stamping

.circleci/workspace_status.sh
```diff
@@ -10,6 +10,9 @@
 # KEY1 VALUE1
 # KEY2 VALUE2
 #
+# Keys starting with "STABLE_" will go into the stable status file. Others in
+# the volatile.
```
pudelkoM, line 14:
```suggestion
# the volatile one.
```

.bazelrc
```diff
@@ -5,6 +5,8 @@
 build --conlyopt='-std=gnu11'
 build --cxxopt='-std=gnu++11'
 build --workspace_status_command=.circleci/workspace_status.sh
+build --stamp
+test --nostamp
```
pudelkoM, line 9: If this does not allow test caching, we could try setting `stamp` on targets of interest instead: https://docs.bazel.build/versions/master/user-manual.html#flag--stamp

.circleci/config.yml
Yi-Tseng: Maybe put a full bazelrc file in `.circleci` dir and use that for CI(as Brian mentioned) so we don't have to modify the root one every time we run the CI job?

bocon13:
supersedes #500

.circleci/workspace_status.sh
pudelkoM: We decided to ignore the scm status for now.

pudelkoM:
CI caching seems to be happy with the stamp config. I'll merge this change and monitor the situation over the next couple of PRs.

PR #518, opened 2021-01-19
Fix unreachable code errors

pudelkoM:
This PR enables the unreachable code detection in the compiler and fixes all current instances.

There might be a concern about the removed `break`s, but we have a warning for implicit switch fallthroughs in place.


stratum/hal/lib/common/openconfig_converter.cc
```diff
@@ -761,9 +761,8 @@ ::util::StatusOr<SingletonPort> InterfaceToSingletonPort(
   }
 
   // FIXME(Yi Tseng): Should we use other field to store interface channel?
-  for (auto& channel : interface.physical_channel()) {
-    to.set_channel(channel.value());
-    break;
+  if (interface.physical_channel_size() > 0) {
+    to.set_channel(interface.physical_channel(0).value());
   }
```
pudelkoM, line 766: This is technically not incorrect, just a bit misleading at glance.

stratum/glue/net_util/ipaddress.cc
```diff
@@ -736,7 +736,6 @@ std::string SocketAddress::ToString() const {
 
 std::string SocketAddress::ToPackedString() const {
   LOG(FATAL) << __FUNCTION__ << " not yet implemented in depot3";
```
Yi-Tseng, line 738: some idea to implement this function

according to the doc:
```
Returns the address as a sequence of bytes in network-byte-order, IPAddress first, then port.
```

```c++
std::string SocketAddress::ToPackedString() const {
  std::string ip_addr_str = host_.ToPackedString();
  char tmp[ip_addr_str.size() + 2]; // IP address len + L4 port len
  strncpy(tmp, ip_addr_str.c_str(), ip_addr_str.size());
  tmp[ip_addr_str.size()] = (char)(port_ >> 8);
  tmp[ip_addr_str.size()+1] = (char)(port_ & 0xff);
  return std::string(tmp, ip_addr_str.size()+2);
}
```

pudelkoM: Thanks for picking this up and trying to fix the root cause!

If you want, we should open another PR for this.

Yi-Tseng (dismissed):
overall looks good

stratum/glue/net_util/ipaddress.cc
pudelkoM: We even have the tests for it https://github.com/stratum/stratum/blob/9f5bd2b285badbef11e81eca6c31d4a3c4342843/stratum/glue/net_util/ipaddress_test.cc#L1186-L1276

PR #519, opened 2021-01-20
Enable compiler check for shadowed variables and fix

pudelkoM:
This PR enables two new compiler flags: `-Werror=shadow` and `-Werror=uninitialized` to improve code quality. It also fixes all current instances in the code base.


pudelkoM:
I'm a little unsure about this flag. It requires more fixes to existing code and makes variable naming a tiny bit harder sometimes, but it also found a real but in the credentials manager.

bocon13 (approved):

Yi-Tseng (approved):

bocon13 (commented):
LGTM. Given the chance for real bugs (e.g. shadowing class members with locals) and the fact that we've seen some of these in the past, I'm okay with the downsides.

PR #524, opened 2021-01-21
Implement simple port shaping on Tofino

pudelkoM:
This PR adds support for configuring port shaping on Tofino switches. Initially we expose this feature via the ChassisConfig only, but might extend this to gNMI in the future.

Example config:

```
nodes {
...
}
vendor_config {
  tofino_config {
    node_id_to_port_shaping_config {
      key: 1
      value {
        per_port_shaping_configs {
          key: 9
          value {
            byte_shaping {
              max_rate_bps: 10000000000 # 10G
              max_burst_bytes: 16384 # 2x jumbo frame
            }
          }
        }
      }
    }
  }
}
singleton_ports {
...
```

TODOs:

- [x] Data plane verification test
- [x] Save shaping config into internal port config map
- [x] Clean up shaping config on updates
- [x] Example in readme

### DIFF
--- a/stratum/hal/lib/barefoot/bf_chassis_manager.cc
+++ b/stratum/hal/lib/barefoot/bf_chassis_manager.cc
@@ -119,6 +119,36 @@ BFChassisManager::~BFChassisManager() = default;
     config->admin_state = ADMIN_STATE_ENABLED;
   }
 
+  if (config_params.has_shaping_config()) {
+    LOG(INFO) << "Configuring port shaping on port " << port_id << " in node "
+              << node_id << " (SDK Port " << sdk_port_id
+              << "): " << config_params.shaping_config().ShortDebugString()
+              << ".";
+    switch (config_params.shaping_config().shaping_case()) {
+      case PortConfigParams::ShapingConfig::kPacketShaping:
+        RETURN_IF_ERROR(bf_sde_interface_->SetPortShapingRate(
+            unit, sdk_port_id, true,
+            config_params.shaping_config()
+                .packet_shaping()
+                .burst_size_packets(),
+            config_params.shaping_config().packet_shaping().rate_pps()));
+        break;
+      case PortConfigParams::ShapingConfig::kByteShaping:
+        RETURN_IF_ERROR(bf_sde_interface_->SetPortShapingRate(
+            unit, sdk_port_id, false,
+            config_params.shaping_config().byte_shaping().burst_size_bytes(),
+            config_params.shaping_config().byte_shaping().rate_kbits()));
+        break;
+      default:
+        RETURN_ERROR(ERR_INVALID_PARAM)
+            << "Invalid port shaping config "
+            << config_params.shaping_config().ShortDebugString() << ".";
+    }
+    RETURN_IF_ERROR(bf_sde_interface_->EnablePortShaping(unit, sdk_port_id,
+                                                         TRI_STATE_TRUE));
+    // TODO(max): update config?
+  }
+
   return ::util::OkStatus();
 }
 
@@ -278,7 +308,7 @@ BFChassisManager::~BFChassisManager() = default;
     }
   }
 
-  for (auto singleton_port : config.singleton_ports()) {
+  for (const auto& singleton_port : config.singleton_ports()) {
     uint32 port_id = singleton_port.id();
     uint64 node_id = singleton_port.node();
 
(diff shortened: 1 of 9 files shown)


stratum/hal/lib/common/common.proto
```diff
@@ -276,6 +291,8 @@ message PortConfigParams {
   uint64 mac_address = 7;
   // The configured loopback state for this port.
   LoopbackState loopback_mode = 8;
+  // The port shaping configuration for this port.
```
pudelkoM, line 294:
```suggestion
  // The traffic shaping configuration for this port.
```

pudelkoM:
TST comments:

- A comprehensive solution should consider more than just shaping on Tofino, i.e. H-QoS
- Repeating the same config on every port is error prone, modeling with references is prefered

stratum/hal/lib/common/common.proto
```diff
@@ -658,8 +658,35 @@ message GoogleConfig {
...
+        uint64 max_rate_bps = 1;
+        uint32 max_burst_bytes = 2;
+      }
+      oneof shaping {
+        PacketShape packet_shaping = 1;
+        ByteShape byte_shaping = 2;
+      }
+    }
+    // Map from port id to its shaping config given by BfPerPortShapingConfig.
+    map<uint32, BfPerPortShapingConfig> per_port_shaping_configs = 1;
```
ccascone, line 679: Just making sure, is port ID here the SDN port ID or the SDK one? From the code, it seems the former, which is good. It would be good to add a comprehensive example in the Barefoot README that shows both how to use `singleton_ports` and `vendor_config` in the same file.

pudelkoM: SDN port ID.

+1 on the example

stratum/hal/lib/barefoot/bf_sde_interface.h
```diff
@@ -177,6 +177,17 @@ class BfSdeInterface {
   // Disable a port.
   virtual ::util::Status DisablePort(int device, int port) = 0;
 
+  // Set the port shaping properties on a port.
+  // If in_pps is true, the burst size and rate is measured in packets and pps.
+  // Else, it's in bytes and kbps.
+  virtual ::util::Status SetPortShapingRate(int device, int port, bool in_pps,
```
bocon13, line 183: Suggestion: `in_pps` -> `is_in_pps` or `is_pps`

stratum/hal/lib/barefoot/bf_sde_interface.h
```diff
@@ -177,6 +177,17 @@ class BfSdeInterface {
...
+                                            uint32 burst_size,
+                                            uint32 rate_per_second) = 0;
```
bocon13, line 185: We might want to keep the rate in bps to be consistent with other uses throughout.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -392,6 +428,40 @@ ::util::Status BFChassisManager::PushChassisConfig(
...
+          shaping_config.packet_shaping().max_burst_packets(),
+          shaping_config.packet_shaping().max_rate_pps()));
+      break;
+    case TofinoConfig::BfPortShapingConfig::BfPerPortShapingConfig::
+        kByteShaping:
+      // The SDE expects the rate in kbps.
+      RETURN_IF_ERROR(bf_sde_interface_->SetPortShapingRate(
+          unit, sdk_port_id, false,
+          shaping_config.byte_shaping().max_burst_bytes(),
+          shaping_config.byte_shaping().max_rate_bps() / 1000));
```
bocon13, line 449: Might be better to do this conversion in the wrapper

pudelkoM: I guess it comes down to this: Minimal wrapper around SDE or sane interface with QoL improvements. Keep in mind that bps will require `uint64`.

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -924,6 +924,31 @@ ::util::Status BfSdeWrapper::DisablePort(int device, int port) {
   return ::util::OkStatus();
 }
 
+::util::Status BfSdeWrapper::SetPortShapingRate(int device, int port,
+                                                bool is_in_pps,
+                                                uint32 burst_size,
+                                                uint64 rate_per_second) {
+  if (!is_in_pps) {
+    rate_per_second /= 1000;  // The SDE expects the bitrate in kpps.
```
pudelkoM, line 932:
```suggestion
    rate_per_second /= 1000;  // The SDE expects the bitrate in kbps.
```

PR #535, opened 2021-02-01
Add deflect-on-drop API to Stratum-bfrt

pudelkoM:
This PR adds allows setting the deflect-on-drop config on Stratum-bfrt via the ChassisConfig.

Example config:
```
vendor_config {
  tofino_config {
    node_id_to_deflect_on_drop_configs {
      key: 1  # node id
      value {
        drop_targets {
          port: 16  # reference to singleton port
          # or 
          # sdk_port: 324
          queue: 0
        }
      }
    }
  }
}
```

TODO:

- [x] test on HW

### DIFF
--- a/stratum/hal/lib/barefoot/bf_chassis_manager.cc
+++ b/stratum/hal/lib/barefoot/bf_chassis_manager.cc
@@ -407,18 +407,30 @@ BFChassisManager::~BFChassisManager() = default;
       const uint64 node_id = key.first;
       const auto& deflect_config = key.second;
       for (const auto& drop_target : deflect_config.drop_targets()) {
-        const uint32 port_id = drop_target.port();
         CHECK_RETURN_IF_FALSE(node_id_to_port_id_to_sdk_port_id.count(node_id));
         CHECK_RETURN_IF_FALSE(node_id_to_unit.count(node_id));
-        int unit = node_id_to_unit[node_id];
-        CHECK_RETURN_IF_FALSE(
-            node_id_to_port_id_to_sdk_port_id[node_id].count(port_id));
-        const uint32 sdk_port_id =
-            node_id_to_port_id_to_sdk_port_id[node_id][port_id];
+        const int unit = node_id_to_unit[node_id];
+        uint32 sdk_port_id;
+        switch (drop_target.port_type_case()) {
+          case TofinoConfig::DeflectOnPacketDropConfig::DropTarget::kPort: {
+            const uint32 port_id = drop_target.port();
+            CHECK_RETURN_IF_FALSE(
+                node_id_to_port_id_to_sdk_port_id[node_id].count(port_id));
+            sdk_port_id = node_id_to_port_id_to_sdk_port_id[node_id][port_id];
+            break;
+          }
+          case TofinoConfig::DeflectOnPacketDropConfig::DropTarget::kSdkPort: {
+            sdk_port_id = drop_target.sdk_port();
+            break;
+          }
+          default:
+            RETURN_ERROR(ERR_INVALID_PARAM)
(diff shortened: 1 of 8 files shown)


stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -393,6 +397,32 @@ ::util::Status BFChassisManager::PushChassisConfig(
             shaping_config;
       }
     }
+
+    // Handle deflect-on-drop config.
+    const auto& node_id_to_deflect_on_drop_configs =
+        config.vendor_config()
+            .tofino_config()
+            .node_id_to_deflect_on_drop_configs();
+    for (const auto& key : node_id_to_deflect_on_drop_configs) {
```
pudelkoM, line 406: Add validation against multiple destinations within the same pipe.

stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -393,6 +397,32 @@ ::util::Status BFChassisManager::PushChassisConfig(
...
+        CHECK_RETURN_IF_FALSE(node_id_to_port_id_to_sdk_port_id.count(node_id));
+        CHECK_RETURN_IF_FALSE(node_id_to_unit.count(node_id));
+        int unit = node_id_to_unit[node_id];
+        CHECK_RETURN_IF_FALSE(
+            node_id_to_port_id_to_sdk_port_id[node_id].count(port_id));
+        const uint32 sdk_port_id =
+            node_id_to_port_id_to_sdk_port_id[node_id][port_id];
+        RETURN_IF_ERROR(bf_sde_interface_->SetDeflectOnDropDestination(
+            unit, sdk_port_id, drop_target.queue()));
+        LOG(INFO) << "Configured deflect on drop target port " << sdk_port_id
```
pudelkoM, line 420:
```suggestion
        LOG(INFO) << "Configured deflect on drop target SDK port " << sdk_port_id
```

stratum/hal/lib/common/common.proto
```diff
@@ -679,9 +679,20 @@ message TofinoConfig {
     map<uint32, BfPerPortShapingConfig> per_port_shaping_configs = 1;
   }
 
+  // DeflectOnPacketDropConfig specifies the port and queue where dropped
+  // packets will be deflected to.
+  message DeflectOnPacketDropConfig {
+    message DropTarget {
+      uint32 port = 1;
```
pudelkoM, line 686: We might need to change that to `oneof { port, sdk_port}`, else we have to add the recirc ports to the singleton port list.

Yi-Tseng: I think we can use `oneof`. I don't think we cannot add recirculate port to the singleton port list since we cannot find the SDK port number for the recirculate port by using port/channel

Yi-Tseng: should we also add `pipe` to this config for specific pipes?

ccascone: I see some inconsistency with port shaping config where the port id is passed as a map key. Not a big deal, since this is an internal API, but I was wondering if you also see this as a potential source of confusion and if you have given some thought to how to make the two configs similar. For example, should we use the `oneof` approach for both port shaping and dod?

pudelkoM: Notes from sync:

- `oneof` requires a flat config structure with repeated fields instead of nested maps
- might want to change port shaping from nested maps to flat (future PR)

Yi-Tseng (approved):
lgtm

pudelkoM:
Functionality has been empirically confirmed by @Yi-Tseng on a hardware switch.

PR #536, opened 2021-02-02
Refactor SwitchInterface to use generic StreamMessageRequest/Response types

### LINKED ISSUE #482: Support for digests and idle timeout notifications on tofino?
Hi, is it possible to get digests and idle timeout notifications through Stratum when running on tofino?
From what i can tell it seems like this is not currently supported, but i could be wrong.
Thanks!

### LINKED ISSUE #528: stratum_bmv2 does not send P4Runtime DigestList via StreamChannel
Even if bmv2 and PI support P4Runtime digests, `StreamMessageResponse.DigestList` messages do not come out of `stratum_bmv2` when they should. I believe this is happening because we don't have a "digest channel" as we do for packet-ins in the p4_service/switch_interface.

pudelkoM:
This PR modifies the SwitchInterface to use the generic `StreamMessageRequest` and `StreamMessageResponse` data types instead of the embedded messages like `PacketIn` or `PacketOut`. Using the encapsulating message reduces the number of nearly identical functions in the interface. Switch implementations can still decide which sub-messages to support, but extending support in one implementation does not require changes to the interface anymore.

This change also includes code changes to allow passing digests lists and acks from and to the controller for Stratum-bf, as PI does most of the work and we only connect the parts.

We do not address the topic of caching and resending digests / acks in this PR. For now this problem is pushed into the switch implementations (by passing down the acks transparently).

### Design discussion

Passing down the more generic `StreamMessageResponse` channel (writer) instead of the specialized `PacketIn` weakens the type-safety of the interface: a packetio manger could write digest lists to the channel. Here are all options:

- Pass down generic `StreamMessageResponse`: weakens type-safety, most generic, no extra resources
- Split sub-messages into separate functions and channels: enlarges SwitchInterface with nearly duplicate functions, needs additional threads and channels per sub-message type
- `ConstraintChannelWriter`: no extra resources, keeps type-safety, needs somewhat specialized one-off code

TODOs:

- [x] Test this on HW
- [x] Update all implementations with new SwitchInterface function signatures

Closes #528
Closes #482

### DIFF
--- a/stratum/hal/lib/bcm/bcm_node.cc
+++ b/stratum/hal/lib/bcm/bcm_node.cc
@@ -354,7 +354,6 @@ BcmNode::~BcmNode() {}
     case ::p4::v1::StreamMessageRequest::kPacket: {
       return bcm_packetio_manager_->TransmitPacket(
           GoogleConfig::BCM_KNET_INTF_PURPOSE_CONTROLLER, req.packet());
-      break;
     }
     default:
       RETURN_ERROR(ERR_UNIMPLEMENTED) << "Unsupported StreamMessageRequest "
--- a/stratum/hal/lib/bcm/bcm_node.cc
+++ b/stratum/hal/lib/bcm/bcm_node.cc
@@ -344,7 +344,7 @@ BcmNode::~BcmNode() {}
       GoogleConfig::BCM_KNET_INTF_PURPOSE_CONTROLLER);
 }
 
-::util::Status BcmNode::SendStreamMessageRequest(
+::util::Status BcmNode::HandleStreamMessageRequest(
     const ::p4::v1::StreamMessageRequest& req) {
   absl::ReaderMutexLock l(&lock_);
   if (!initialized_) {
--- a/stratum/hal/lib/bcm/bcm_node.cc
+++ b/stratum/hal/lib/bcm/bcm_node.cc
@@ -10,6 +10,7 @@
 #include "absl/memory/memory.h"
 #include "absl/synchronization/mutex.h"
 #include "gflags/gflags.h"
+#include "stratum/hal/lib/common/channel_writer_wrapper.h"
 #include "stratum/lib/macros.h"
 
 // TODO(unknown): This flag is currently false to skip static entry writes
@@ -318,17 +319,28 @@ BcmNode::~BcmNode() {}
   return ::util::OkStatus();
 }
 
-::util::Status BcmNode::RegisterPacketReceiveWriter(
-    const std::shared_ptr<WriterInterface<::p4::v1::PacketIn>>& writer) {
+::util::Status BcmNode::RegisterStreamMessageResponseWriter(
+    const std::shared_ptr<WriterInterface<::p4::v1::StreamMessageResponse>>&
+        writer) {
   absl::WriterMutexLock l(&lock_);
   if (!initialized_) {
     return MAKE_ERROR(ERR_NOT_INITIALIZED) << "Not initialized!";
   }
+  auto w =
+      std::make_shared<ConstraintChannelWriter<::p4::v1::StreamMessageResponse,
+                                               ::p4::v1::PacketIn>>(
+          writer, &::p4::v1::StreamMessageResponse::mutable_packet);
+  ::p4::v1::PacketIn pkt;
+  LOG(WARNING) << "RegisterStreamMessageResponseWriter: "
+               << Demangle(typeid(w).name());
+
+  bcm_packetio_manager_->RegisterPacketReceiveWriterOld(
+      GoogleConfig::BCM_KNET_INTF_PURPOSE_CONTROLLER, w);
   return bcm_packetio_manager_->RegisterPacketReceiveWriter(
       GoogleConfig::BCM_KNET_INTF_PURPOSE_CONTROLLER, writer);
 }
 
-::util::Status BcmNode::UnregisterPacketReceiveWriter() {
+::util::Status BcmNode::UnregisterStreamMessageResponseWriter() {
   absl::WriterMutexLock l(&lock_);
   if (!initialized_) {
     return MAKE_ERROR(ERR_NOT_INITIALIZED) << "Not initialized!";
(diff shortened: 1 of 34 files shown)


stratum/hal/lib/bcm/bcm_packetio_manager.cc
```diff
@@ -343,7 +343,8 @@ ::util::Status BcmPacketioManager::Shutdown() {
 
 ::util::Status BcmPacketioManager::RegisterPacketReceiveWriter(
     GoogleConfig::BcmKnetIntfPurpose purpose,
-    const std::shared_ptr<WriterInterface<::p4::v1::PacketIn>>& writer) {
+    const std::shared_ptr<WriterInterface<::p4::v1::StreamMessageResponse>>&
```
pudelkoM, line 346: Here is the main problem of using the generic parent message type: the packetio manager should only be able to create `PacketIn` messages, but now we have pass in the `StreamMessageResponse` type.

stratum/hal/lib/bcm/bcm_node.cc
```diff
@@ -318,17 +319,28 @@ ::util::Status BcmNode::ReadForwardingEntries(
...
+    const std::shared_ptr<WriterInterface<::p4::v1::StreamMessageResponse>>&
+        writer) {
   absl::WriterMutexLock l(&lock_);
   if (!initialized_) {
     return MAKE_ERROR(ERR_NOT_INITIALIZED) << "Not initialized!";
   }
+  auto w =
+      std::make_shared<ConstraintChannelWriter<::p4::v1::StreamMessageResponse,
+                                               ::p4::v1::PacketIn>>(
+          writer, &::p4::v1::StreamMessageResponse::mutable_packet);
```
pudelkoM, line 332: This creates a wrapper around the generic `StreamMessageResponse` writer that only accepts `PacketIn` messages. Since it satisfies the `WriterInterface<::p4::v1::PacketIn>` interface, we can pass it to the unmodified managers and maintain type safety.

stratum/hal/lib/bcm/bcm_packetio_manager.cc
Yi-Tseng: We could create the StreamMessageResponse in the SwitchInterface, but that means we need another reader thread in the SwitchInterface to read the PacketIn message from the packet-IO manager (and maybe the digest manager) 🤔

pudelkoM: yes, that's another alternative. Creating a new channel and thread for each sub-type. Not my favorite, tbh.

stratum/hal/lib/bcm/bcm_node_test.cc
```diff
@@ -1619,26 +1620,27 @@ TEST_F(BcmNodeTest, WriteForwardingEntriesSuccess_DeleteMulticastGroupEntry) {
...
 
-  auto writer = std::make_shared<WriterMock<::p4::v1::PacketIn>>();
-  EXPECT_CALL(*bcm_packetio_manager_mock_,
-              RegisterPacketReceiveWriter(
-                  GoogleConfig::BCM_KNET_INTF_PURPOSE_CONTROLLER, Eq(writer)))
+  auto writer = std::make_shared<WriterMock<::p4::v1::StreamMessageResponse>>();
+  EXPECT_CALL(
+      *bcm_packetio_manager_mock_,
+      RegisterPacketReceiveWriter(
+          GoogleConfig::BCM_KNET_INTF_PURPOSE_CONTROLLER, _ /*Eq(writer)*/))
```
pudelkoM, line 1632: Check if a match is possible.

stratum/hal/lib/common/p4_service.cc
```diff
@@ -833,42 +850,48 @@ bool P4Service::IsMasterController(uint64 node_id, uint64 connection_id) const {
...
 }
 
-void P4Service::PacketReceiveHandler(uint64 node_id,
-                                     const ::p4::v1::PacketIn& packet) {
-  // We send the packets only to the master controller stream for this node.
+void P4Service::StreamResponseReceiveHandler(
+    uint64 node_id, const ::p4::v1::StreamMessageResponse& resp) {
+  // We don't expect arbitration updates from the switch.
+  if (resp.has_arbitration()) {
+    LOG(ERROR) << "Received MasterArbitrationUpdate from switch. This should "
```
pudelkoM, line 887: `FATAL`?

stratum/hal/lib/common/switch_interface.h
```diff
@@ -150,25 +149,33 @@ class SwitchInterface {
...
+  // Registers a writer to be invoked when we receive a StreamMessageResponse on
+  // the specified node which are destined for the controller. A
+  // StreamMessageResponse can carry many different types of sub-messages, such
+  // as PacketIns, digests or idle timeout notifications.
+  // PacketIns: The sent ::p4::PacketIn instance includes all the info on
+  //      where the packet was received on this node as well as its payload.
+  virtual ::util::Status RegisterStreamMessageResponseWriter(
       uint64 node_id,
-      std::shared_ptr<WriterInterface<::p4::v1::PacketIn>> writer) = 0;
+      std::shared_ptr<WriterInterface<::p4::v1::StreamMessageResponse>>
```
bocon13, line 160: Can we make this a const reference instead of a copy?

stratum/hal/lib/common/p4_service.cc
```diff
@@ -578,30 +578,45 @@ ::grpc::Status P4Service::StreamChannel(
...
+          status = MAKE_ERROR(ERR_PERMISSION_DENIED)
+                   << "Controller with connection ID " << connection_id
+                   << "is not a master";
+        } else {
+          // If master, try to ack the digest.
+          status = switch_interface_->SendStreamMessageRequest(node_id, req);
+        }
         if (!status.ok()) {
-          LOG_EVERY_N(INFO, 500) << "Failed to transmit packet: " << status;
+          LOG_EVERY_N(INFO, 500) << "Failed to ack digest: " << status;
```
bocon13, line 611: Maybe make this more frequent?

stratum/hal/lib/common/switch_interface.h
pudelkoM: Here is the literature I found on this topic:

https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs
> Avoid defining functions that require a const reference parameter to outlive the call, because const reference parameters bind to temporaries.

https://abseil.io/tips/116
> It is still OK to pass an argument by const reference to a constructor if the argument is copied, or just used in the constructor and no reference to it is kept in the constructed object.

https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/
> Guideline: Express that a function will store and share ownership of a heap object using a by-value shared_ptr parameter.

https://stackoverflow.com/questions/3310737
> Shortly, there is no reason to pass by value, unless the goal is to share ownership of an object

stratum/hal/lib/common/p4_service.cc
bocon13: Fatal is probably warranted here, but it's also not harmful to just drop the arb update and continue...

stratum/hal/lib/common/p4_service.cc
pudelkoM: yes, just `LOG`

stratum/hal/lib/common/p4_service.cc
```diff
@@ -578,30 +578,45 @@ ::grpc::Status P4Service::StreamChannel(
...
           auto resp = ToStreamMessageResponse(status);
-          *resp.mutable_error()->mutable_packet_out()->mutable_packet_out() =
-              req.packet();
+          *resp.mutable_error()
+               ->mutable_digest_list_ack()
+               ->mutable_digest_list_ack() = req.digest_ack();
           stream->Write(resp);  // Best effort.
```
pudelkoM, line 616: Add comment about not clogging the stream channel.

stratum/hal/lib/common/p4_service.cc
pudelkoM: We agreed on `FATAL` on the basis that this is a serious bug in our code and we want to increase the chances of getting a user report in that case.

stratum/hal/lib/bcm/bcm_node_test.cc
pudelkoM: sadly not.

PR #538, opened 2021-02-04
Update setup_dev_env.sh for automated testing

naradahess:
- Added the --program option to setup_dev_env.sh to allow invocation with a program other than bash. When this option is given, the docker container is run without -ti (though the user could add those options back in manually if needed).
- Added the run_unit_tests.sh script to the tools directory. This is intended for use with the --program option of setup_dev_env.sh
- Changed setup_dev_env.sh to use the user's group ID in the docker container. Otherwise, it defaults to using the user's UID as the GID. This can cause issues on shared systems where the user does not have sudo privileges, as the bogus GID maps to a random group, and the user cannot delete files in the .cache directory that were created with that GID. This has been tested in various environments at Extreme. Please test in your environment to verify it causes no issues.

### DIFF
--- a/setup_dev_env.sh
+++ b/setup_dev_env.sh
@@ -30,6 +30,7 @@ Usage: $0
     [--git-name <name>]             use the provided name for git commits
     [--git-email <email>]           use the provided email for git commits
     [--git-editor <editor command>] use the provided editor for git
+    [--program <path>]              run the provided program instead of bash
     [--np4-intel]                   create NP4 Intel build environment
     [-- [Docker options]]           additional Docker options for running the container
 EOF
@@ -68,6 +69,11 @@ do
         ;;
     --git-editor)
         GIT_EDITOR="$2"
+        shift
+        shift
+        ;;
+    --program)
+        DOCKER_PROGRAM="$2"
         shift
         shift
         ;;
@@ -131,5 +137,15 @@ fi
 if [ "$NP4_INTEL" == YES ]; then
     DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS -v /dev/intel-fpga-fme.0:/dev/intel-fpga-fme.0"
 fi
+
+DOCKER_GID=$(id -g $USER)
+if [ -z "$DOCKER_GID" ]; then
+    DOCKER_GID=$(id -u $USER)
+fi
+if [ -z "$DOCKER_PROGRAM" ]; then
+    DOCKER_PROGRAM="bash"
+    DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS -it"
+fi
+
 DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS $@"
-docker run $DOCKER_RUN_OPTIONS -w /stratum --user $USER -ti $IMAGE_NAME bash
+docker run $DOCKER_RUN_OPTIONS -w /stratum --user $USER:$DOCKER_GID $IMAGE_NAME $DOCKER_PROGRAM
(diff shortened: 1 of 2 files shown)


tools/run_unit_tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Copyright 2021-present Extreme Networks, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# Run stratum unit tests inside the stratum dev container.
+# Run from /stratum.
+
+TEST_CMD="xargs -a .circleci/test-targets.txt bazel test"
+# --flaky_test_attempts to retry transient admin_service_test failures
+TEST_CMD="$TEST_CMD --flaky_test_attempts 5"
```
pudelkoM, line 10: This should be part of the bazel rc.

tools/run_unit_tests.sh
```diff
@@ -0,0 +1,11 @@
...
+eval $TEST_CMD
```
pudelkoM, line 11: I like the idea of putting "complicated" commands into scripts, but then we should also use them in CI here: https://github.com/stratum/stratum/blob/68c8bf999baf256af23f530e36cec060b9919d49/.circleci/config.yml#L95

tools/run_unit_tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
pudelkoM, line 1: There is no formal rule in place, but it seems that we're putting CI (and testing is CI) related scripts in `.circleci/`. `tools/` is more user oriented.

This file literally depends on `.circleci/test-targets.txt`.

tools/run_unit_tests.sh
pudelkoM, line 11:
```
> shellcheck tools/run_unit_tests.sh

In tools/run_unit_tests.sh line 11:
eval $TEST_CMD
     ^-------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
eval "$TEST_CMD"
```

setup_dev_env.sh
```diff
@@ -30,6 +30,8 @@ Usage: $0
     [--git-name <name>]             use the provided name for git commits
     [--git-email <email>]           use the provided email for git commits
     [--git-editor <editor command>] use the provided editor for git
+    [--program <path>]              run the provided program instead of bash
+                                    for example: /stratum/tools/run_unit_tests.sh
```
pudelkoM, line 34: I see why you want to add this command, but the `setup_dev_env.sh` script is not for CI.

You can just use the image normally, i.e. `docker run` and set a different [entrypoint](https://docs.docker.com/engine/reference/commandline/run/).
The details depend on your CI system, but most should support something like this: https://github.com/stratum/stratum/blob/68c8bf999baf256af23f530e36cec060b9919d49/.circleci/config.yml#L79-L80

Considering that the Stratum CI will not use this flag, I'm against adding this flag.

tools/run_unit_tests.sh
```diff
@@ -0,0 +1,11 @@
...
+# Copyright 2021-present Extreme Networks, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# Run stratum unit tests inside the stratum dev container.
+# Run from /stratum.
+
```
pudelkoM, line 7:
```suggestion
set -e
```
for error propagation?

tools/run_unit_tests.sh
naradahess: Actually, we do not want TEST_CMD in quotes because we want the words split and globbing to happen. (shellcheck can be a little overreaching.) All your other comments make sense and I will reevaluate my approach.

tools/run_unit_tests.sh
naradahess: Probably not really needed as there is only one command in this script and it is at the end. But good general practice, yes.

tools/run_unit_tests.sh
pudelkoM: I see. You can leave this as is then, we'll re-check this once we add shellcheck to CI.

PR #544, opened 2021-02-10
Update APS Networks switches in the README

bocon13:
Renaming Stordis to APS Networks

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ to program fixed-pipeline switches using the Tomahawk chip from Broadcom.
 
 ## Tofino
 
+- APS Networks BF2556X-1T
+- APS Networks BF6064X-T
 - Delta AG9064v1
 - Edgecore Wedge100BF-32QS
 - Edgecore Wedge100BF-32X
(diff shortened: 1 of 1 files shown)


pudelkoM:
Do we also want to rename files and folders like `stratum/hal/config/x86-64-stordis-bf6064x-t-r0`? There are also some strings in ChassisConfigs.

bocon13:
> Do we also want to rename files and folders like `stratum/hal/config/x86-64-stordis-bf6064x-t-r0`? There are also some strings in ChassisConfigs.

That requires some changes outside of Stratum as well (e.g. in ONL) that I don't really want to worry about. Maybe we can ask them to comment on their plans here.

pudelkoM:
> That requires some changes outside of Stratum as well (e.g. in ONL)

Good point. Even if there will be a new platform string, it's probably better to duplicate than to rename to keep compat with older devices.

pudelkoM (approved):
LGTM.

Change git commit message to "Update APS ..."?

PR #549, opened 2021-02-13
Do not return priority for default table entries

pudelkoM:
The the SDE returns 0 (unset, highest), which we would incorrectly translate to UINT32_MAX.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager.cc
@@ -446,7 +446,7 @@ struct RegisterClearThreadData {
   }
 
   // Priority.
-  // Default action do not have a priority, even when the table ususally
+  // Default actions do not have a priority, even when the table usually
   // requires one. The SDE would return 0 (highest).
   if (has_priority_field && !request.is_default_action()) {
     uint32 bf_priority;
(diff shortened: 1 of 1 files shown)


ccascone (dismissed):
LGTM

stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -445,8 +445,10 @@ ::util::StatusOr<::p4::v1::TableEntry> BfrtTableManager::BuildP4TableEntry(
     }
   }
 
-  // Priority
-  if (has_priority_field) {
+  // Priority.
+  // Default action do not have a priority, even when the table ususally
```
pudelkoM, line 449:
```suggestion
  // Default actions do not have a priority, even when the table usually
```

Yi-Tseng (dismissed):

stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -444,8 +444,14 @@ ::util::StatusOr<::p4::v1::TableEntry> BfrtTableManager::BuildP4TableEntry(
             << ".";
     }
   }
+  
+  // Default actions do not have a priority, even when the table usually
+  // requires one. The SDE would return 0 (highest) which we must not translate.
+  if (request.is_default_action()) {
+    has_priority_field = false;
+  }
```
Yi-Tseng, line 452: I was thinking we can simplify to this

```c++
has_priority_field = !request.is_default_action();
```

However, the current one more clear to developers.
feel free to merge it when you think it Is ready

PR #557, opened 2021-02-20
The codecov upload script should return 1 on failures

pudelkoM:
Previously the job did not fail, even when the coverage upload job failed.

https://docs.codecov.io/docs/about-the-codecov-bash-uploader#uploading-process


pudelkoM:
Not sure what the problem is, but none of the flags work as documented. `-Z` just always fails.

pudelkoM:
Since there currently is no way to detect errors from the codecov script, we'll close this issue and monitor CI.

PR #561, opened 2021-02-22
Initial support for SDE 9.4.0 and 9.3.1

pudelkoM:
This PR add support for SDE version 9.3.1 (now the recommended release) and 9.4.0 (experimental).

- [x] Update READMEs


bazel/deps.bzl
```diff
@@ -23,6 +23,8 @@ BF_SDE_PI_VER = {
     "9_1_0": "41358da0ff32c94fa13179b9cee0ab597c9ccbcc",
     "9_2_0": "4546038f5770e84dc0d2bba90f1ee7811c9955df",
     "9_3_0": "4546038f5770e84dc0d2bba90f1ee7811c9955df",
+    "9_3_1": "4546038f5770e84dc0d2bba90f1ee7811c9955df",
+    "9_4_0": "4546038f5770e84dc0d2bba90f1ee7811c9955df",
```
pudelkoM, line 27: @Yi-Tseng to confirm the PI versions.

bazel/deps.bzl
```diff
@@ -23,6 +23,8 @@ BF_SDE_PI_VER = {
     "9_1_0": "41358da0ff32c94fa13179b9cee0ab597c9ccbcc",
```
pudelkoM, line 23: Maybe remove?

stratum/hal/lib/barefoot/BUILD
```diff
@@ -145,6 +145,10 @@ stratum_cc_library(
     name = "bf_sde_wrapper",
     srcs = ["bf_sde_wrapper.cc"],
     hdrs = ["bf_sde_wrapper.h"],
+    defines = select({
+        "//stratum/hal/lib/pi:sde_9_4_0": ["SDE_9_4_0"],
+        "//conditions:default": [],
```
pudelkoM, line 150: We should probably list all versions here and hard fail if the `--define` flags was not set.

bazel/deps.bzl
Yi-Tseng: yes, `4546038f5770e84dc0d2bba90f1ee7811c9955df` is the correct one.

Yi-Tseng:
Is it possible to use the `VERSION` file from the `$SDE_INSTALL/share/VERSION` to find the version?

bazel/deps.bzl
pudelkoM: Thanks for confirming.

pudelkoM:
> 
> 
> Is it possible to use the `VERSION` file from the `$SDE_INSTALL/share/VERSION` to find the version?

Good idea, I'll look into that.

pudelkoM:
I have found no way to `--define` or otherwise influence the build config based on the contents of a file. For now we can enforce the usage of `--define sde_ver=x` when building stratum-bf{rt}.

bazel/deps.bzl
pudelkoM: Will do later.

PR #564, opened 2021-02-25
Auto-detect and expose BF SDE version as config_setting

pudelkoM:
This PR implements BF SDE version auto-detection and exposes the result as a `config_setting` so `selects()` can match on it. This removes the need to manually specify the `--define sde_ver=x` flag on the command line. Further, we can detect and reject unsupported versions.

Detection is implemented by reading the `$SDE_INSTALL/share/VERSION` file and then templating the value into a [`string_setting`](https://github.com/bazelbuild/bazel-skylib/blob/528e4241345536c487cca8b11db138104bb3bd68/docs/common_settings_doc.md#string_setting), a user-defined [`build_setting`](https://docs.bazel.build/versions/master/skylark/config.html). This is the preferred way over `--define`s.

TODOs:

- [x] Are the generated `config_settings` in `bf_pi.bzl` still needed?
- [x] Remove SDE 9.3.1 and 9.4.0 references, they will be added in a later PR


stratum/hal/lib/barefoot/BUILD
```diff
@@ -17,6 +17,17 @@ package(
...
+    {
+        "@local_barefoot_bin//:sde_version_9.1.0": ["SDE_9_1_0"],
+        "@local_barefoot_bin//:sde_version_9.2.0": ["SDE_9_2_0"],
+        "@local_barefoot_bin//:sde_version_9.3.0": ["SDE_9_3_0"],
+        "@local_barefoot_bin//:sde_version_9.3.1": ["SDE_9_3_1"],
+        # "@local_barefoot_bin//:sde_version_9.4.0": ["SDE_9_4_0"],
+    },
+    no_match_error = "Unsupported SDE version",
+)
+
```
pudelkoM, line 30: This defines are not needed yet (9.4.0 will), but demonstrate how to use the new `config_setting`s.

bazel/external/bfsde.BUILD
```diff
@@ -84,3 +85,46 @@ pkg_tar(
...
     strip_prefix = "barefoot-bin",
 )
+
+# This string setting is templated with the correct version string by reading
+# the $SDE_INSTALL/share/VERSION file. Then one of the config settings below
+# will match and can be used with select().
+string_setting(
+    name = "sde_version_setting",
+    build_setting_default = "{SDE_VERSION}",
+)
```
pudelkoM, line 95: I actually don't know how to set a non default value. `string_flags` are exposed to the CLI, this one is **NOT**.

stratum/hal/lib/barefoot/barefoot.bzl
```diff
@@ -19,7 +19,14 @@ def _impl(repository_ctx):
         print("SDE_INSTALL is deprecated, please use SDE_INSTALL_TAR")
         bf_sde_install_path = repository_ctx.os.environ["SDE_INSTALL"]
         repository_ctx.symlink(bf_sde_install_path, local_install_path)
-    repository_ctx.symlink(Label("@//bazel:external/bfsde.BUILD"), "BUILD")
+    ver = repository_ctx.read(local_install_path + "/share/VERSION").strip("\n")
```
pudelkoM, line 22: If the SDE structure ever changes, we can handle this here and still expose a unified interface.

PR #565, opened 2021-02-26
Log the SDE version on SdeWrapper creation

pudelkoM:
Currently the runtime logs do not contain the SDE version. This change logs it once on startup with level `INFO`.

There are two possible places to implement the log:

- `BfSdeWrapper::CreateSingleton()`
  - It's always logged, users can't forget it
  - Not configurable (level, time) 
  - Single place to update/fix/keep track of
  - No precedence: we don't log non-errors in `CreateSingleton`
- `main.cc/main_bfrt.cc`: Log in main by calling the public `GetSdeVersion()` function
  - Logging is callers responsibility, more flexible
  - Makes SDE version publicly available, other modules might want to implement dependent behavior in the future
  - Precedence: `GetBfChipType`, `IsSoftwareModel`

### DIFF
--- a/stratum/hal/lib/barefoot/bf_sde_mock.h
+++ b/stratum/hal/lib/barefoot/bf_sde_mock.h
@@ -97,6 +97,7 @@ class BfSdeMock : public BfSdeInterface {
   MOCK_METHOD2(SetTmCpuPort, ::util::Status(int device, int port));
   MOCK_METHOD1(IsSoftwareModel, ::util::StatusOr<bool>(int device));
   MOCK_CONST_METHOD1(GetBfChipType, std::string(int device));
+  MOCK_CONST_METHOD0(GetSdeVersion, std::string());
   MOCK_METHOD2(TxPacket, ::util::Status(int device, const std::string& packet));
   MOCK_METHOD1(StartPacketIo, ::util::Status(int device));
   MOCK_METHOD1(StopPacketIo, ::util::Status(int device));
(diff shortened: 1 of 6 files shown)


stratum/hal/lib/barefoot/bf_sde_mock.h
```diff
@@ -97,6 +97,7 @@ class BfSdeMock : public BfSdeInterface {
   MOCK_METHOD2(SetTmCpuPort, ::util::Status(int device, int port));
   MOCK_METHOD1(IsSoftwareModel, ::util::StatusOr<bool>(int device));
   MOCK_CONST_METHOD1(GetBfChipType, std::string(int device));
+  MOCK_CONST_METHOD0(GetSdeVersion, std::string ());
```
Yi-Tseng, line 100:
```suggestion
  MOCK_CONST_METHOD0(GetSdeVersion, std::string());
```

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -2967,11 +2971,24 @@ ::util::Status BfSdeWrapper::SynchronizeRegisters(
...
 
+std::string BfSdeWrapper::GetSdeVersionPrivate() const {
+#if defined(SDE_9_1_0)
+  return "9.1.0";
+#elif defined(SDE_9_2_0)
+  return "9.2.0";
+#elif defined(SDE_9_3_0)
+  return "9.3.0";
+#elif defined(SDE_9_3_1)
+  return "9.3.1";
```
Yi-Tseng, line 2982: can we add `#error` for other cases that don't match the cases below

pudelkoM: I actually had an `#error` here before, but removed it because compilation will fail anyway (missing return on a non-void function). But an `#error` can be more clear with a message that gets printed.

Yi-Tseng (dismissed):

pudelkoM:
@bocon13 and I prefer the second form.

PR #566, opened 2021-02-26
Remove SDE 9.1.0

pudelkoM:
It does not compile anymore because of missing BfRt APIs and version 9.4.0 is released.


bocon13 (approved):
lgtm

PR #571, opened 2021-03-05
Reject P4RT reads or writes before a pipeline has been pushed in Stratum-bfrt

pudelkoM:
Stratum-bfrt did not check for a pipeline and thus segfaulted on a null pointer. This PR adds the appropriate check to all public functions.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_node.cc
+++ b/stratum/hal/lib/barefoot/bfrt_node.cc
@@ -123,10 +123,19 @@ BfrtNode::~BfrtNode() = default;
 
 ::util::Status BfrtNode::Shutdown() {
   absl::WriterMutexLock l(&lock_);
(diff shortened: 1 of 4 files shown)


stratum/hal/lib/barefoot/bfrt_node.cc
```diff
@@ -120,7 +122,10 @@ ::util::Status BfrtNode::VerifyForwardingPipelineConfig(
 }
 
 ::util::Status BfrtNode::Shutdown() {
-  // RETURN_IF_BFRT_ERROR(bf_device_remove(device_id_));
+  absl::WriterMutexLock l(&lock_);
+  // TODO(max): do we need to de-init the ASIC or SDE?
+  pipeline_initialized_ = false;
```
pudelkoM, line 127: At least the packetio manager can be shutdown here. Maybe more.

PR #572, opened 2021-03-07
Update gflags dependency to 2.2.2

pudelkoM:
Import gflags as recommended by https://gflags.github.io/gflags/#bazel


pudelkoM:
Actually, tag `2.2.2` is even older than what we have right now. Let's move to a newer commit instead?

Yi-Tseng:
I think the upgrade to the latest one should be ok, I don't see a huge change to the gflag codebase

PR #577, opened 2021-03-16
disable ipv6 solicitations in mininet

eth0xFEED:
while debugging an issue noticed that kernel originated packets filling the logs , this turns off that behavior

### DIFF
--- a/tools/mininet/stratum.py
+++ b/tools/mininet/stratum.py
@@ -294,4 +294,4 @@ switches = {'stratum-bmv2': StratumBmv2Switch}
 hosts = {
     'no-offload-host': NoOffloadHost,
     'no-ipv6-host': NoIpv6OffloadHost
-    }
+}
--- a/tools/mininet/stratum.py
+++ b/tools/mininet/stratum.py
@@ -278,8 +278,20 @@ class NoOffloadHost(Host):
             self.cmd(cmd)
         return r
 
+class NoIpv6OffloadHost(NoOffloadHost):
+    def __init__(self, name, inNamespace=True, **params):
+        NoOffloadHost.__init__(self, name, inNamespace=inNamespace, **params)
+
+    def config(self, **params):
+        r = super(NoOffloadHost, self).config(**params)
+        self.cmd("sysctl net.ipv6.conf.%s.disable_ipv6=1" % (self.defaultIntf()))
+        return r
+
 
 # Exports for bin/mn
 switches = {'stratum-bmv2': StratumBmv2Switch}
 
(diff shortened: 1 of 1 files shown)


tools/mininet/stratum.py
```diff
@@ -276,6 +276,8 @@ def config(self, **params):
             cmd = "/sbin/ethtool --offload %s %s off" \
                   % (self.defaultIntf(), off)
             self.cmd(cmd)
+        cmd = "sysctl net.ipv6.conf.%s.disable_ipv6=1" % (self.defaultIntf())
+        self.cmd(cmd)
```
pudelkoM, line 280: I'm not sure how the `hosts` export for mininet works, but shouldn't this command be it's own class? Like `NoIpv6Host`?

eth0xFEED: Good point will get that fixed up 👍

tools/mininet/stratum.py
```diff
@@ -278,8 +278,20 @@ def config(self, **params):
...
+
 
 # Exports for bin/mn
 switches = {'stratum-bmv2': StratumBmv2Switch}
 
-hosts = {'no-offload-host': NoOffloadHost}
+hosts = {
+    'no-offload-host': NoOffloadHost,
+    'no-ipv6-host': NoIpv6OffloadHost
+    }
```
pudelkoM, line 297:
```suggestion
}
```

pudelkoM (dismissed):
LGTM, but I'll let @Yi-Tseng decide. He has more experience with mininet.

Yi-Tseng (dismissed):
LGTM

Yi-Tseng (approved):

tools/mininet/stratum.py
```diff
@@ -278,8 +278,20 @@ def config(self, **params):
...
+
 
 # Exports for bin/mn
 switches = {'stratum-bmv2': StratumBmv2Switch}
 
-hosts = {'no-offload-host': NoOffloadHost}
+hosts = {
+    'no-offload-host': NoOffloadHost,
+    'no-ipv6-host': NoIpv6OffloadHost
+}
```
pudelkoM, line 297: @Yi-Tseng do you know how this export interacts with Mininet? Does this create two hosts?

Yi-Tseng: This is for the command line parameter, for example:

`mn --custom stratum.py --switch stratum-bmv2 --host no-ipv6-host linear,3`

Ref:
https://github.com/stratum/stratum/blob/main/tools/mininet/Dockerfile#L118

PR #581, opened 2021-03-18
Fix zero byte string bug in IsDontCareMatch for ranges

pudelkoM:
`std::string("\x00")` will surprisingly create an empty string `""`, not a one byte string containing a single zero byte character. As a result, `absl::StartsWith(..., "\x00")` will always return true and trim the whole lower range match down to a single byte. If this last byte is a zero, the match will incorrectly classified as "don't care".
Also see the Notes section of [basic_string](https://en.cppreference.com/w/cpp/string/basic_string/basic_string).

### DIFF
--- a/stratum/hal/lib/barefoot/utils_test.cc
+++ b/stratum/hal/lib/barefoot/utils_test.cc
@@ -160,6 +160,12 @@ TEST(IsDontCareMatchTest, ClassifyRangeMatch) {
     m.set_high("\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff", 10);
     EXPECT_FALSE(IsDontCareMatch(m, 81)) << m.DebugString();
   }
+  {
+    ::p4::v1::FieldMatch::Range m;
+    m.set_low("\x00\x40\x00", 3);
+    m.set_high("\x03\xFF\xFF", 3);
+    EXPECT_FALSE(IsDontCareMatch(m, 18)) << m.DebugString();
+  }
 }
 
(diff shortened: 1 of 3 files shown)


stratum/hal/lib/barefoot/utils_test.cc
```diff
@@ -160,6 +160,12 @@ TEST(IsDontCareMatchTest, ClassifyRangeMatch) {
     m.set_high("\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff", 10);
     EXPECT_FALSE(IsDontCareMatch(m, 81)) << m.DebugString();
   }
+  {
+    ::p4::v1::FieldMatch::Range m;
+    m.set_low("\x00\x40\x00", 3);
+    m.set_high("\x03\xFF\xFF", 3);
+    EXPECT_FALSE(IsDontCareMatch(m, 24)) << m.DebugString();
```
pudelkoM, line 167:
```suggestion
    EXPECT_FALSE(IsDontCareMatch(m, 18)) << m.DebugString();
```

Yi-Tseng:
How about `std::string("\x00", 1)` ?

pudelkoM:
> ~How about `std::string("\x00", 1)` ? ~
> NVM, I saw your change

I search and replaced all other instances I could find, just the one also in this PR. Then I just fixed the algorithm to be more concise. As a side effect, there's no `"\x00"` string anymore, just a `'\x00'` character.

PR #585, opened 2021-03-19
Add remote bfshell support

Yi-Tseng:
Adds missing files to support remote bfshell (via TCP)


pudelkoM:
Nice!

Is the full bfshell or just ucli available over telnet?

Yi-Tseng:
> Nice!
> 
> Is the full bfshell or just ucli available over telnet?

we actually expose full bfshell (ucli, pipemgr, bfrt_python, ...) but some features require additional libraries

pudelkoM:
> > Nice!
> > Is the full bfshell or just ucli available over telnet?
> 
> we actually expose full bfshell (ucli, pipemgr, bfrt_python, ...) but some features require additional libraries

So this is even better than the normally exposed shell. I wonder if these libraries allow flipping this to false: https://github.com/stratum/stratum/blob/662cc6a42ff3e82858be2c95ccd4ba1ceee1ce2f/stratum/hal/bin/barefoot/main_bfrt.cc#L59

pudelkoM:
To answer my own question: no-ish

Stratum starts but accessing `bfrt_python` will result in a segfault:

```
bfshell> bfrt_python
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
Fatal Python error: Py_Initialize: Unable to get the locale encoding
ImportError: No module named 'encodings'

Current thread 0x00007fd6253bc700 (most recent call first):
*** Aborted at 1616199084 (unix time) try "date -d @1616199084" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGABRT (@0x5978) received by PID 22904 (TID 0x7fd6253bc700) from PID 22904; stack trace: ***
    @     0x7fd63875b0e0 (unknown)
    @     0x7fd637fc2fff gsignal
    @     0x7fd637fc442a abort
    @     0x7fd638ad585f Py_FatalError
    @     0x7fd638ad5b3a (unknown)
    @     0x7fd638ad7b89 _Py_InitializeEx_Private
    @     0x7fd63a44513b bfrt_cli_cmd
    @     0x7fd66bf96438 clish_shell_exec_action
    @     0x7fd66bf96a32 clish_shell_execute
    @     0x7fd66bf99a8c (unknown)
    @     0x7fd66bf982a2 clish_shell_loop
    @     0x7fd66bf9db9f (unknown)
    @     0x7fd6387514a4 start_thread
    @     0x7fd638078d0f clone
    @                0x0 (unknown)
fish: 'start-stratum.sh -v=1 -bf_switc…' terminated by signal SIGABRT (Abort)
```

**Same over telnet, running `bfrt_python` will crash the Stratum process!**

pudelkoM (dismissed):
In light of the brittleness of the bf-shell I'm ok with this, but would not replace the `ucli` with it yet.

Maybe add some docs: "This is debug only and has a high chance to crash Stratum!"

ccascone (approved):
LGTM. @Yi-Tseng it would be great if you could update the README with an example command of how to run the bfshell when using the Stratum Docker container.

stratum/hal/bin/barefoot/README.run.md
```diff
@@ -453,6 +453,22 @@ The CLI will start after Stratum and the SDE finish initialization. You may
 need to press "Enter" a few times to see the prompt. Please refer to BF SDE
 documentation for details on how to use the CLI.
 
+### Access BF Shell via telnet (experimental)
+
+When Stratum started, the BF Shell will be enbaled by default.
```
pudelkoM, line 458:
```suggestion
When Stratum started, the BF Shell will be enabled by default.
```

stratum/hal/bin/barefoot/README.run.md
```diff
@@ -453,6 +453,22 @@ The CLI will start after Stratum and the SDE finish initialization. You may
...
+
+To access the BF Shell, use the `telnet` command:
+
+```bash
+telnet [Switch IP] 9999
+```
+
+Once enter the BF Shell, type `ucli` to access the BF CLI.
+
+> Since we are not support all features from the BF SHell, using commands other than `ucli` may cause Stratum crash.
```
pudelkoM, line 468:
```suggestion
> Since we do not support all features from the BF Shell, using commands other than `ucli` may cause Stratum crash.
```

stratum/hal/bin/barefoot/README.run.md
```diff
@@ -453,6 +453,22 @@ The CLI will start after Stratum and the SDE finish initialization. You may
...
+
+To exit the telnet, press `Ctrl` and `]` to close the connection and type `quit` to leave the telnet.
```
pudelkoM, line 470:
```suggestion
To exit the telnet session, press `Ctrl` and `]` to close the connection and type `quit` to leave the telnet.
```

stratum/hal/bin/barefoot/README.run.md
```diff
@@ -453,6 +453,22 @@ The CLI will start after Stratum and the SDE finish initialization. You may
...
+
+When Stratum started, the BF Shell will be enbaled by default.
+
+To access the BF Shell, use the `telnet` command:
+
+```bash
+telnet [Switch IP] 9999
+```
+
+Once enter the BF Shell, type `ucli` to access the BF CLI.
```
pudelkoM, line 466:
```suggestion
Once you have entered the BF Shell, type `ucli` to access the BF CLI.
```

stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -23,6 +23,7 @@ else
     DOCKER_NET_OPTS="-p 28000:28000 "
     DOCKER_NET_OPTS+="-p 9339:9339 "
     DOCKER_NET_OPTS+="-p 9559:9559 "
+    DOCKER_NET_OPTS+="-p 9999:9999 "
```
ccascone, line 26: In the PDP sync we discussed about not exposing the port, but instead providing instructions to run telnet inside the container. The rationale was something around security, but I don't think I can provide a strong argument.

Still, I think it would be more convenient to allow running telnet inside the container.

For example, in Aether, we could access the BF shell in one command using kubectl exec, instead of forwarding ports and making sure telnet is installed.

If you agree, can you please update the README with an example of how to run telnet inside the container? We might also need to update the Dockerfile to make sure telnet is installed.

PR #587, opened 2021-03-23
Implement a dummy ForwardingViability handler in all switches

pudelkoM:
Not handling the `ForwardingViability` request type in the switch interface leads to undesirable errors in the log. This PR normalizes the behavior across all `SwitchInterface` implementations with stratum-bcm, where we now always return `TRUNK_MEMBER_BLOCK_STATE_UNKNOWN`.

This behavior is in line with our decision that all `SwitchInterface` implementations must implement all message types.

### DIFF
--- a/stratum/hal/lib/barefoot/bf_chassis_manager.cc
+++ b/stratum/hal/lib/barefoot/bf_chassis_manager.cc
@@ -706,6 +706,15 @@ BFChassisManager::GetPortConfig(uint64 node_id, uint32 port_id) const {
       resp.mutable_sdn_port_id()->set_port_id(sdk_port_id);
       break;
     }
+    case Request::kForwardingViability: {
+      // Find current port forwarding viable state for port located at:
+      // - node_id: req.forwarding_viable().node_id()
+      // - port_id: req.forwarding_viable().port_id()
+      // and then write it into the response.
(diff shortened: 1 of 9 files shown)


Yi-Tseng (approved):
consider using the `TRUNK_MEMBER_BLOCK_STATE_UNKNOWN` as the default state? since the document/comment in the common.proto says it is the default/invalid value.

pudelkoM:
> consider using the `TRUNK_MEMBER_BLOCK_STATE_UNKNOWN` as the default state? since the document/comment in the common.proto says it is the default/invalid value.

Good point! For this particular path it's not that relevant since `UNKNOWN` will be converted to false, but other paths with enums or string values can express this.

I updated the return values in all dummy implementations to `UNKNOWN`.

pudelkoM:
Example response from a dummy:

```
notification {
  timestamp: 1616626572898947016
  update {
    path {
      elem {
        name: "interfaces"
      }
      elem {
        name: "interface"
        key {
          key: "name"
          value: "1/0"
        }
      }
      elem {
        name: "ethernet"
      }
      elem {
        name: "state"
      }
      elem {
        name: "forwarding-viable"
      }
    }
    val {
      bool_val: false
    }
  }
}
```

PR #588, opened 2021-03-23
Implement a dummy MacAddress handler in all switches

pudelkoM:
Not handling the MacAddress request type in the switch interface leads to undesirable errors in the log. This normalizes the behavior across all switches, which now return `00:00:00:00:00:00`.

This behavior is in line with our decision that all SwitchInterface implementations must implement all message types.

### DIFF
--- a/stratum/hal/lib/barefoot/bf_chassis_manager.cc
+++ b/stratum/hal/lib/barefoot/bf_chassis_manager.cc
@@ -641,6 +641,15 @@ BFChassisManager::GetPortConfig(uint64 node_id, uint32 port_id) const {
       resp.mutable_admin_status()->set_state(config->admin_state);
       break;
     }
+    case Request::kMacAddress: {
+      // TODO(unknown) Find out why the controller needs it.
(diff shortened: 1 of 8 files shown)


stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -641,6 +641,15 @@ ::util::StatusOr<DataResponse> BFChassisManager::GetPortData(
       resp.mutable_admin_status()->set_state(config->admin_state);
       break;
     }
+    case Request::kMacAddress: {
+      // TODO(unknown) Find out why the controller needs it.
+      // Find MAC address of port located at:
+      // - node_id: req.mac_address().node_id()
+      // - port_id: req.mac_address().port_id()
+      // and then write it into the response.
+      resp.mutable_mac_address()->set_mac_address(0x112233445566ull);
```
Yi-Tseng, line 650: I would suggest using `0` as the default mac address if possible.

PR #589, opened 2021-03-23
Implement hardware-port as leaf reference

pudelkoM:
This removes the field from the SwitchInterface and fixes the errors when querying it on unsupported switches.

### DIFF
--- a/stratum/hal/lib/common/yang_parse_tree_paths.cc
+++ b/stratum/hal/lib/common/yang_parse_tree_paths.cc
@@ -1022,36 +1022,19 @@ void SetUpInterfacesInterfaceStateLoopbackMode(uint64 node_id, uint32 port_id,
 
 ////////////////////////////////////////////////////////////////////////////////
 // /interfaces/interface[name=<name>]/state/hardware-port
-void SetUpInterfacesInterfaceStateHardwarePort(uint64 node_id, uint32 port_id,
-                                               TreeNode* node,
+void SetUpInterfacesInterfaceStateHardwarePort(TreeNode* node,
                                                YangParseTree* tree) {
-  // Regular method using a template cannot be used to get the OnPoll functor as
-  // std::string fields are treated differently by the PROTO-to-C++ generator:
-  // the getter returns "const std::string&" instead of "string" which leads to
-  // the template compilation error.
-  auto poll_functor = [node_id, port_id, tree](const GnmiEvent& event,
-                                      const ::gnmi::Path& path,
-                                      GnmiSubscribeStream* stream) {
-    // Create a data retrieval request.
-    DataRequest req;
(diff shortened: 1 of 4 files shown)


stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -1022,36 +1022,19 @@ void SetUpInterfacesInterfaceStateLoopbackMode(uint64 node_id, uint32 port_id,
...
-    // it is created.
-    tree->GetSwitchInterface()
-        ->RetrieveValue(node_id, req, &writer, /* details= */ nullptr)
-        .IgnoreError();
-    return SendResponse(GetResponse(path, resp), stream);
+  // This leaf is a reference to the /components/component[name=<name>]/name
+  // leaf. We look up the data from there.
+  std::string component_name = node->GetPath().elem(1).key().at("name");
+  // TODO(max): the path should end with /name, but we don't support that yet.
+  ::gnmi::Path ref_path = GetPath("components")("component", component_name)();
```
Yi-Tseng, line 1031: can we move these two lines into the functor?

pudelkoM: Probably, I didn't want to capture the node pointer in the closure, because lifetime?

If we move it into the functor, the ref path is created on every call, but it never changes.

pudelkoM: Alternatively we pass the name as string to the entire function and capture that.

Yi-Tseng: I see, I forgot that we need to pass other things like the tree to the functor if we move these two lines into the functor.
I think it is fine to keep them outside the functor

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -3269,7 +3252,7 @@ TreeNode* YangParseTreePaths::AddSubtreeInterface(
 
   node = tree->AddNode(GetPath("interfaces")(
       "interface", name)("state")("hardware-port")());
-  SetUpInterfacesInterfaceStateHardwarePort(node_id, port_id, node, tree);
+  SetUpInterfacesInterfaceStateHardwarePort(node, tree);
```
Yi-Tseng, line 3255: we can also pass the `name` to the function so we don't need to parse the node to get the name.

pudelkoM: yeah, the whole reference leaf thing is not really needed here as the name is exactly the same.
I think I'll do what you suggest, but keep this code change for later reference.

pudelkoM:
Superseded by #594, but kept for future reference, should we have to implement true openconfig ref leafs.

PR #590, opened 2021-03-24
Implement a dummy health indicator handler in all switches

pudelkoM:
Not handling the HealthIndicator request type in the switch interface leads to undesirable errors in the log. This normalizes the behavior across all switches, which now return `HEALTH_STATE_UNKNOWN`.

This behavior is in line with our decision that all SwitchInterface implementations must implement all message types.

### DIFF
--- a/stratum/hal/lib/barefoot/bf_chassis_manager.cc
+++ b/stratum/hal/lib/barefoot/bf_chassis_manager.cc
@@ -724,6 +724,14 @@ BFChassisManager::GetPortConfig(uint64 node_id, uint32 port_id) const {
           TRUNK_MEMBER_BLOCK_STATE_UNKNOWN);
       break;
     }
+    case DataRequest::Request::kHealthIndicator: {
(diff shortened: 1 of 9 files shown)


stratum/hal/lib/barefoot/bf_chassis_manager.cc
```diff
@@ -724,6 +724,14 @@ ::util::StatusOr<DataResponse> BFChassisManager::GetPortData(
           TRUNK_MEMBER_BLOCK_STATE_FORWARDING);
       break;
     }
+    case DataRequest::Request::kHealthIndicator: {
+      // Find current port health indicator (LED) for port located at:
+      // - node_id: req.health_indicator().node_id()
+      // - port_id: req.health_indicator().port_id()
+      // and then write it into the response.
+      resp.mutable_health_indicator()->set_state(HEALTH_STATE_GOOD);
```
Yi-Tseng, line 732: can we use `HEALTH_STATE_UNKNOWN` as dummy values?

PR #593, opened 2021-03-26
Use `RE2` for yang mac-address parsing

pudelkoM:
Addressed issues:

- `::google::protobuf::uint64` -> `uint64`
- `<regex>` -> RE2
- make `YangStringToMacAddress` safer
- comments
- style

This PR also fixes things introduces in commit https://github.com/stratum/stratum/commit/6e71349660bd1dbde65a9c5b765e391559e6a288.

### DIFF
--- a/stratum/hal/lib/common/utils.cc
+++ b/stratum/hal/lib/common/utils.cc
@@ -282,7 +282,7 @@ uint64 ConvertStringToSpeedBps(const std::string& speed_string) {
   } else if (speed_string == "SPEED_100GB") {
     return kHundredGigBps;
   } else {
-    return 0LL;
+    return 0ull;
   }
 }
 
--- a/stratum/hal/lib/common/utils.cc
+++ b/stratum/hal/lib/common/utils.cc
@@ -318,10 +318,10 @@ bool ConvertTrunkMemberBlockStateToBool(const TrunkMemberBlockState& state) {
 }
 
 std::string MacAddressToYangString(const uint64& mac_address) {
-  return absl::StrFormat("%x:%x:%x:%x:%x:%x", (mac_address >> 40) & 0xFF,
-                         (mac_address >> 32) & 0xFF, (mac_address >> 24) & 0xFF,
-                         (mac_address >> 16) & 0xFF, (mac_address >> 8) & 0xFF,
-                         mac_address & 0xFF);
+  return absl::StrFormat("%02x:%02x:%02x:%02x:%02x:%02x",
+                         (mac_address >> 40) & 0xFF, (mac_address >> 32) & 0xFF,
+                         (mac_address >> 24) & 0xFF, (mac_address >> 16) & 0xFF,
+                         (mac_address >> 8) & 0xFF, mac_address & 0xFF);
 }
 
 ::util::StatusOr<uint64> YangStringToMacAddress(std::string yang_string) {
--- a/stratum/hal/lib/common/utils.cc
+++ b/stratum/hal/lib/common/utils.cc
@@ -323,16 +323,16 @@ std::string MacAddressToYangString(const uint64& mac_address) {
                          (mac_address >> 8) & 0xFF, mac_address & 0xFF);
 }
 
-::util::StatusOr<uint64> YangStringToMacAddress(std::string yang_string) {
+::util::StatusOr<uint64> YangStringToMacAddress(
+    const std::string& yang_string) {
   RET_CHECK(
       RE2::FullMatch(yang_string, R"#(^[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}$)#"))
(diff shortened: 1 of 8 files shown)


stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -1706,10 +1701,10 @@ void SetUpInterfacesInterfaceEthernetStateAutoNegotiate(
 // data from the DataResponse proto received from SwitchInterface, i.e.,
 // "&DataResponse::PortCounters::message", where message field in
 // DataResponse::Counters.
-TreeNodeEventHandler GetPollCounterFunctor(
-    uint64 node_id, uint32 port_id,
-    ::google::protobuf::uint64 (PortCounters::*func_ptr)() const,
-    YangParseTree* tree) {
+TreeNodeEventHandler GetPollCounterFunctor(uint64 node_id, uint32 port_id,
+                                           uint64 (PortCounters::*func_ptr)()
+                                               const,
```
Yi-Tseng, line 1706: is there a better way to make `const` and the `func_ptr` in the same line?

stratum/hal/lib/common/yang_parse_tree_test.cc
```diff
@@ -1161,13 +1161,14 @@ TEST_F(YangParseTreeTest,
                               /* Notification will not be called */ nullptr),
               StatusIs(_, _, ContainsRegex("not a TypedValue message")));
 
-  for (auto mac_string : kMacStrings) {
+  for (auto mac_string : kInvalidMacStrings) {
```
pudelkoM, line 1164:
```suggestion
  for (const auto& mac_string : kInvalidMacStrings) {
```

stratum/hal/lib/common/yang_parse_tree_paths.cc
pudelkoM: Yes! I think the yang tree arg should come before the func_ptr, like with all the other `Get*Functor` functions.

stratum/hal/lib/common/utils.cc
```diff
@@ -318,34 +316,33 @@ bool ConvertTrunkMemberBlockStateToBool(const TrunkMemberBlockState& state) {
   return state == TRUNK_MEMBER_BLOCK_STATE_FORWARDING;
 }
 
-std::string MacAddressToYangString(
-    const ::google::protobuf::uint64& mac_address) {
+std::string MacAddressToYangString(const uint64& mac_address) {
   return absl::StrFormat("%x:%x:%x:%x:%x:%x", (mac_address >> 40) & 0xFF,
```
bocon13, line 320:
```suggestion
  return absl::StrFormat("%2x:%2x:%2x:%2x:%2x:%2x", (mac_address >> 40) & 0xFF,
```

stratum/hal/lib/common/yang_parse_tree_test.cc
```diff
@@ -1147,8 +1147,8 @@ TEST_F(YangParseTreeTest,
       "11:22:333:44:55:66",    // Too many hex digits
```
bocon13, line 1147:
```suggestion
      "0:0:0:0:0:0",         // Single digits not support
      "11:22:333:44:55:66",    // Too many hex digits
```

stratum/hal/lib/common/utils.cc
pudelkoM: It was `%02x` for zero padding in the end. Feel free to extend the new tests cases.

stratum/hal/lib/common/utils.cc
```diff
@@ -268,22 +267,21 @@ std::string ConvertSpeedBpsToString(
...
 }
 
-::google::protobuf::uint64 ConvertStringToSpeedBps(
-    const std::string& speed_string) {
-  if (speed_string.compare("SPEED_10GB") == 0) {
+uint64 ConvertStringToSpeedBps(const std::string& speed_string) {
+  if (speed_string == "SPEED_10GB") {
     return kTenGigBps;
-  } else if (speed_string.compare("SPEED_20GB") == 0) {
+  } else if (speed_string == "SPEED_20GB") {
```
pudelkoM, line 273: For reference: https://stackoverflow.com/a/9159670

stratum/hal/lib/common/utils.cc
bocon13: Yes, sorry.

stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -1343,7 +1339,7 @@ void SetUpInterfacesInterfaceEthernetConfigMacAddress(uint64 node_id,
       return status;
     }
```
pudelkoM, line 1340: This feels like `RETURN_IF_ERROR`, but check other places.

PR #597, opened 2021-03-29
Update BCM device list with SDK information and link to TAI

pudelkoM:
There often is some confusion around which SDK is needed for a given device.


bocon13 (approved):
Added some minor comments, otherwise LGTM

PR #598, opened 2021-03-29
Add the user's group ID to setup_env.sh docker run

naradahess:
Closed the old PR and created this single-issue PR just for the GID addition.

### DIFF
--- a/setup_dev_env.sh
+++ b/setup_dev_env.sh
@@ -117,6 +117,10 @@ if [ $ERR -ne 0 ]; then
     exit $ERR
 fi
 
+DOCKER_GID=$(id -g $USER)
+if [ -z "$DOCKER_GID" ]; then
+    DOCKER_GID=$(id -u $USER)
+fi
 DOCKER_RUN_OPTIONS="--rm -v $THIS_DIR:/stratum"
 if [ "$MOUNT_SSH" == YES ]; then
     DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS -v $HOME/.ssh:/home/$USER/.ssh"
@@ -132,4 +136,4 @@ if [ "$NP4_INTEL" == YES ]; then
     DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS -v /dev/intel-fpga-fme.0:/dev/intel-fpga-fme.0"
 fi
 DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS $@"
-docker run $DOCKER_RUN_OPTIONS -w /stratum --user $USER -ti $IMAGE_NAME bash
+docker run $DOCKER_RUN_OPTIONS -w /stratum --user $USER:$DOCKER_GID -ti $IMAGE_NAME bash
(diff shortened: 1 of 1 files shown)


setup_dev_env.sh
```diff
@@ -117,6 +117,10 @@ if [ $ERR -ne 0 ]; then
     exit $ERR
 fi
 
+DOCKER_GID=$(id -g $USER)
+if [ -z "$DOCKER_GID" ]; then
```
bocon13, line 121: Under what circumstances would `id -g $USER` return nothing? Is setting the GID to the UID the right thing to do?

AFAIK, they are different ID spaces.

bocon13:
Thanks @naradahess for the PR!

setup_dev_env.sh
```diff
@@ -117,6 +117,10 @@ if [ $ERR -ne 0 ]; then
...
+    DOCKER_GID=$(id -u $USER)
```
bocon13, line 122: Under what circumstances would id -g $USER return nothing? Is setting the GID to the UID the right thing to do?

AFAIK, they are different ID spaces.

naradahess: It should never fail. This was defensive programming. And yes, they are different spaces, which was the original problem. In the case of this failure, I was just reverting to the prior (incorrect) behavior, out of lack of knowing what else to do. Perhaps there is a more elegant solution, which would be to set the entire USER:GID string here instead of just setting the GID.

setup_dev_env.sh
```diff
@@ -132,4 +136,4 @@ if [ "$NP4_INTEL" == YES ]; then
     DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS -v /dev/intel-fpga-fme.0:/dev/intel-fpga-fme.0"
 fi
 DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS $@"
-docker run $DOCKER_RUN_OPTIONS -w /stratum --user $USER -ti $IMAGE_NAME bash
+docker run $DOCKER_RUN_OPTIONS -w /stratum --user $USER:$DOCKER_GID -ti $IMAGE_NAME bash
```
naradahess, line 139: Will use a single $USER_GID string here instead (see above comment).

setup_dev_env.sh
```diff
@@ -132,4 +137,4 @@ if [ "$NP4_INTEL" == YES ]; then
     DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS -v /dev/intel-fpga-fme.0:/dev/intel-fpga-fme.0"
 fi
 DOCKER_RUN_OPTIONS="$DOCKER_RUN_OPTIONS $@"
-docker run $DOCKER_RUN_OPTIONS -w /stratum --user $USER -ti $IMAGE_NAME bash
+docker run $DOCKER_RUN_OPTIONS -w /stratum --user $DOCKER_USER -ti $IMAGE_NAME bash
```
pudelkoM, line 140: I just realized that all run options should be contained in the `DOCKER_RUN_OPTIONS` var.

I'll make a separate PR for this.

bocon13 (approved):
looks good to me, thanks!

setup_dev_env.sh
bocon13: Understood :) 

I like your new approach of omitting the GID if not present.

PR #601, opened 2021-03-31
Implement LACP MAC in all switch interfaces and unify dummy MAC

pudelkoM:
This PR unifies the returned MAC address for the `/lacp/interfaces/interface[name=<name>]/state/system-id-mac` and `/interfaces/interface[name=<name>]/ethernet/config/mac-address` paths and across the switch implementations.

### DIFF
--- a/stratum/hal/lib/common/constants.h
+++ b/stratum/hal/lib/common/constants.h
@@ -2,7 +2,6 @@
 // Copyright 2018-present Open Networking Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-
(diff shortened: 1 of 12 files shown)


stratum/hal/lib/common/constants.h
```diff
@@ -68,6 +68,9 @@ constexpr uint8 kIpProtoGre = 47;
 // Precision for converting floating point to ::gnmi::Decimal64
 constexpr uint32 kDefaultPrecision = 2;
 
+// Dummy MAC address used for unsupported DataRequests.
+constexpr uint64 kDummyMacAddress = 0x112233445566ull;
```
pudelkoM, line 72: Exact value tbd

PR #613, opened 2021-04-01
Pin LLVM / clang version to TDB

pudelkoM:
The current LLVM / clang install procedure always fetches the latest LLVM release when the `build` Docker container is built. 
This means we will get random upgrades of the toolchain which will introduce incompatible changes (e.g. clang-format rules change) and will fail our CI builds spontaneously and without clear indication.

This PR pins the LLVM version to a known working value (TBD), so we can do planed upgrades at our pace.

#### Details:

 - Upstream apt repository: https://apt.llvm.org/
 - Version selection is done by picking the appropriate repo:
```
deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch main
deb-src http://apt.llvm.org/stretch/ llvm-toolchain-stretch main
# 11 
deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch-11 main
deb-src http://apt.llvm.org/stretch/ llvm-toolchain-stretch-11 main
# 12 
deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch-12 main
deb-src http://apt.llvm.org/stretch/ llvm-toolchain-stretch-12 main
```
- Versioned repos **do not provide** the convenience meta packages (`clang`) that set up all the generic symlinks, e.g. `/usr/bin/clang -> ../lib/llvm-13/bin/clang`
  - Only `/usr/bin/clang-12 -> ../lib/llvm-12/bin/clang` will exist
  - We can create the missing ones as part of the docker image (`clang`, `clang++` and `clang-format`)
  - We do not want to `export CXX=/usr/bin/clang++-12`, because this will break SDE builds and some dependencies that are also built inside the container. During CI we override this for Bazel targets we know that compile.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,7 @@ jobs:
   # Build targets and run unit tests.
   unit_tests:
     docker:
-      - image: stratumproject/build:build
+      - image: stratumproject/build:build-test
     environment:
       - CC: clang
       - CXX: clang++
@@ -99,7 +99,7 @@ jobs:
   # CDLang
   cdlang_tests:
     docker:
-      - image: stratumproject/build:build
+      - image: stratumproject/build:build-test
     environment:
       - CC: clang
       - CXX: clang++
@@ -121,7 +121,7 @@ jobs:
   # Generate coverage report and upload to codecov.io.
   coverage:
     docker:
-      - image: stratumproject/build:build
+      - image: stratumproject/build:build-test
     environment:
       - CC: clang
       - CXX: clang++
@@ -146,7 +146,7 @@ jobs:
     environment:
       - DOCKER_SCOPE: .
       - DOCKER_FILE: Dockerfile.build
-      - DOCKER_IMG: stratumproject/build:build
+      - DOCKER_IMG: stratumproject/build:build-test
     steps:
       - checkout
       - *docker_login
@@ -175,7 +175,7 @@ jobs:
 
   publish-docker-p4c-fpm:
     docker:
-      - image: stratumproject/build:build
+      - image: stratumproject/build:build-test
     environment:
       - DOCKER_SCOPE: stratum/p4c_backends/fpm
       - DOCKER_FILE: Dockerfile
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -377,13 +377,28 @@ workflows:
...
   build_and_test:
     jobs:
-      - unit_tests
-      - cdlang_tests
-      - coverage
-      - cpp-format-check
-      - cpp-lint-check
-      - license-check
-      - bazel-style-check
+      - publish-docker-build
```
pudelkoM, line 380: Unrelated to this issue; but we should consider to always build the image before running other jobs. This will catch problems like these early on. Main blocker for this is the tag, as we don't want to overwrite the main `:build` tag from PRs.

Dockerfile.build
```diff
@@ -43,12 +43,17 @@ ARG LLVM_REPO_NAME
 RUN wget --quiet -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
 RUN add-apt-repository "$LLVM_REPO_NAME"
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends clang-format clang
+    apt-get install -y --no-install-recommends clang-format-11 clang-11
+# Versioned LLVM releases don't come with the meta packages for the
+# clang -> clang-12 symlinks. We create them manually here.
+RUN ln -s /usr/bin/clang-11 /usr/bin/clang
+RUN ln -s /usr/bin/clang++-11 /usr/bin/clang++
+RUN ln -s /usr/bin/clang-format-11 /usr/bin/clang-format
```
pudelkoM, line 51: The official meta package links these directly: `/usr/bin/clang -> ../lib/llvm-13/bin/clang`.
We should do that as well.

Dockerfile.build
```diff
@@ -5,7 +5,7 @@ ARG BAZELISK_VERSION=1.7.5
 ARG PI_COMMIT=0fbdac256151eb1537cd5ebf19101d5df60767fa
 ARG BMV2_COMMIT=498202d98d83a6c580aa86a3497b9b496bcd43b2
 ARG JDK_URL=https://mirror.bazel.build/openjdk/azul-zulu11.29.3-ca-jdk11.0.2/zulu11.29.3-ca-jdk11.0.2-linux_x64.tar.gz
-ARG LLVM_REPO_NAME="deb http://apt.llvm.org/stretch/  llvm-toolchain-stretch main"
+ARG LLVM_REPO_NAME="deb http://apt.llvm.org/stretch/  llvm-toolchain-stretch-11 main"
```
pudelkoM, line 8: 11 seems to be the most likely version we had before. No formatting changes needed right now.

PR #625, opened 2021-04-09
Cleanup bf_switchd initialization

bocon13:


pudelkoM:
Why not put the init code into the SDE wrapper?

stratum/hal/bin/barefoot/main.cc
```diff
@@ -86,37 +80,13 @@ ::util::Status Main(int argc, char* argv[]) {
...
-  if (FLAGS_bf_switchd_background)
-    switchd_main_ctx->running_in_background = true;
-  else
-    switchd_main_ctx->shell_set_ucli = true;
-
-  /* determine if kernel mode packet driver is loaded */
-  switch_pci_sysfs_str_get(bf_sysfs_fname,
-                           sizeof(bf_sysfs_fname) - sizeof("/dev_add"));
-  strncat(bf_sysfs_fname, "/dev_add", sizeof("/dev_add"));
-  printf("bf_sysfs_fname %s\n", bf_sysfs_fname);
```
pudelkoM, line 109: This would also be a great opportunity to use the correct `LOG` framework.

pudelkoM: And replace [unsafe C string](https://stackoverflow.com/a/6904044) functions with `absl::StrFormat`.

pudelkoM:
Most of the changes have been merged in #640. Feel free rebase and reopen should we still need the C API.

PR #633, opened 2021-04-12
Add support for Intel Tofino SDE 9.5.0

pudelkoM:
This PR adds experimental baseline support for SDE 9.5.0. Basic functionality like startup, pipeline push and flow inserts has been confirmed on hardware, but further testing is needed.

SDE 9.5.0 is a LTS release.

TODOs:

- [x] HW test


stratum/hal/lib/barefoot/bfrt_id_mapper.cc
```diff
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include "absl/strings/match.h"
+#include "bf_rt/bf_rt_table.hpp" // TODO(max): backport fix
```
pudelkoM, line 9: should be extracted

PR #641, opened 2021-04-16
Avoid using the latest codecov script

Yi-Tseng:
Due to security reason, we should not use the latest one from the codecov
website (https://codecov.io/bash)

Recently an unknown hacker gain access and replaced the script which
can potentially send credentials back to the hacker.
For more information: https://about.codecov.io/security-update/


pudelkoM (approved):
lgtm. Let's see how often we have to update the script.

PR #648, opened 2021-04-21
Make HAL shutdown async-safe with atomic variable

pudelkoM:
The current shutdown procedure is not async-safe as it takes a mutex during the signal handler. This change fixes this by using an atomic variable and a waiter thread instead. TSAN reports no issues.

I also tested this with grpc 1.32 to verify that no mutex lock cycle occurs.

Co-authored-by: Brian O'Connor <bocon@opennetworking.org>

### DIFF
--- a/stratum/hal/lib/common/hal.cc
+++ b/stratum/hal/lib/common/hal.cc
@@ -4,7 +4,6 @@
 
 #include "stratum/hal/lib/common/hal.h"
 
-#include <atomic>
 #include <utility>
 
 #include "absl/base/macros.h"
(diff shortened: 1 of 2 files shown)


pudelkoM:
This PR is based on https://github.com/stratum/stratum/pull/349, but uses a `std::atomic<int>` variable instead of a POSIX semaphore. It has superior ergonomics as initialization is easier, there are not spurious wakeups, we preserve the signal value, and with C++20 even the [busy wait loop disappears](https://en.cppreference.com/w/cpp/atomic/atomic/wait).

stratum/hal/lib/common/hal.cc
```diff
@@ -45,12 +46,27 @@ namespace hal {
 
 namespace {
 
+// This atomic int indicates that a signal has been received and we should
+// initiate shutdown.
+std::atomic_int signal_value(0);
```
pudelkoM, line 51: check if this could be a static member of `Hal to encapsulate it further.

pudelkoM:
Closed by #652

PR #649, opened 2021-04-21
WIP packet out support

pudelkoM:


pudelkoM:
> Looks good to me , thanks for the example !
> 
> I tried to remove the padding with few experiments but not working exactly like i thought it would. Feel free to push forward think i'll come back around on the padding cleanup little later on b/c more headache than worth for now :)

Yeah, getting Packout headers right is a pain. I'll brush up the C++ code a little bit and merge.

Thanks for taking a look at this.

stratum/pipelines/stratum-tna/stratum_tna.p4
```diff
@@ -108,7 +112,23 @@ parser StratumIngressParser(packet_in pkt,
...
-        transition parse_ethernet;
+        transition check_ethernet;
+    }
+
+    state check_ethernet {
+        // We use ethernet-like headers to signal the presence of PacketOut
+        // metadata before the actual ethernet frame.
+        ethernet_t tmp = pkt.lookahead<ethernet_t>();
+        transition select(tmp.eth_type) {
+            ETHERTYPE_PACKET_OUT: parse_packet_out;
```
stolsma, line 123: @eth0xFEED Why arn't you using the previously parsed ig_intr_md.ingress_port and compare to cpu_port to find out if it is a packet out? You don't need ethertype checking and the padding then.... Thats what I do in my p4 progs... Or am I missing something here?

stolsma: If you use ethertype like this someone could send in a prepared packet on a normal port and create havoc and security nightmares....🤔

pudelkoM: The PCIe CPU port depends on the Tofino chip model (can 192 or 320). Worse, 192 is also a recirculation port. In Fabric-tna we had to use this workaround, but for this simpler pipeline matching on both can work as long as we don't use 192 otherwise.

stolsma: Ah, ok... :(

Can port metadata be of help here ? (page 16 of https://raw.githubusercontent.com/barefootnetworks/Open-Tofino/master/PUBLIC_Tofino-Native-Arch-Document.pdf). Although this is probably to far out for an example prog as also filling that port metadata table needs to be done beforehand by the controller. Not very P4 standards compliant... :(

pudelkoM: I haven't tried port metadata yet.
But yes, packetio gets very tricky on TNA.

PR #652, opened 2021-04-23
Make Hal shutdown async-safe using POSIX pipes

pudelkoM:
The current shutdown procedure is not async-safe as it takes a mutex during the signal handler. This change fixes this by using a POSIX pipe and a waiter thread instead. TSAN reports no issues.

Co-authored-by: Brian O'Connor <bocon@opennetworking.org>

### DIFF
--- a/stratum/hal/lib/common/hal.cc
+++ b/stratum/hal/lib/common/hal.cc
@@ -4,6 +4,9 @@
 
 #include "stratum/hal/lib/common/hal.h"
 
+#include <limits.h>
+#include <unistd.h>
+
 #include <utility>
 
 #include "absl/base/macros.h"
@@ -48,9 +51,11 @@ namespace {
 // Signal received callback which is registered as the handler for SIGINT and
 // SIGTERM signals using signal() system call.
 void SignalRcvCallback(int value) {
-  Hal* hal = Hal::GetSingleton();
-  if (hal == nullptr) return;
-  hal->HandleSignal(value);
+  static_assert(sizeof(value) <= PIPE_BUF,
+                "PIPE_BUF is smaller than the number of bytes that can be "
+                "written atomically to a pipe.");
+  // No reasonable error handling possible.
+  write(Hal::pipe_write_fd_, &value, sizeof(value));
 }
 
 // Set the channel arguments to match the defualt keep-alive parameters set by
@@ -71,6 +76,8 @@ void SetGrpcServerKeepAliveArgs(::grpc::ServerBuilder* builder) {
 
 Hal* Hal::singleton_ = nullptr;
 ABSL_CONST_INIT absl::Mutex Hal::init_lock_(absl::kConstInit);
+int Hal::pipe_read_fd_ = -1;
+int Hal::pipe_write_fd_ = -1;
 
 Hal::Hal(OperationMode mode, SwitchInterface* switch_interface,
          AuthPolicyChecker* auth_policy_checker,
@@ -87,7 +94,8 @@ Hal::Hal(OperationMode mode, SwitchInterface* switch_interface,
       diag_service_(nullptr),
       file_service_(nullptr),
       external_server_(nullptr),
-      old_signal_handlers_() {}
+      old_signal_handlers_(),
+      signal_waiter_tid_(0) {}
 
 Hal::~Hal() {
   // TODO(unknown): Handle this error?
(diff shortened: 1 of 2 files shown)


stratum/hal/lib/common/hal.cc
```diff
@@ -329,6 +337,15 @@ ::util::Status Hal::RegisterSignalHandlers() {
     }
     old_signal_handlers_[s] = h;
   }
+  // Create the pipe to transfer signals.
+  int pipe_fds[2];
+  CHECK_RETURN_IF_FALSE(pipe(pipe_fds) == 0) << "Could not create pipe.";
```
bocon13, line 342: Could we use a more descriptive message? "pipe for signal handling"?
```suggestion
  CHECK_RETURN_IF_FALSE(pipe(pipe_fds) == 0) << "Could not create pipe for signal handling.";
```

stratum/hal/lib/common/hal.cc
```diff
@@ -329,6 +337,15 @@ ::util::Status Hal::RegisterSignalHandlers() {
...
+  pipe_read_fd_ = pipe_fds[0];
+  pipe_write_fd_ = pipe_fds[1];
+  // Start the signal waiter thread that initiates shutdown.
+  CHECK_RETURN_IF_FALSE(pthread_create(&signal_waiter_tid_, nullptr,
+                                       SignalWaiterThreadFunc, nullptr) == 0)
+      << "Could not start the signal waiter thread.";
```
bocon13, line 348:
```suggestion
      << "Could not start the signal handling thread.";
```
Maybe?

stratum/hal/lib/common/hal.cc
```diff
@@ -4,6 +4,9 @@
 
 #include "stratum/hal/lib/common/hal.h"
 
+#include <limits.h>
```
bocon13, line 7: Do we need limits.h anymore?
```suggestion
```

bocon13 (commented):
Looks pretty good. Just a few comments.

bocon13:
closes #349

stratum/hal/lib/common/hal.cc
pudelkoM:
```
> grep -nr "PIPE_BUF" /usr/include/
/usr/include/linux/limits.h:14:#define PIPE_BUF        4096	/* # bytes in atomic write to a pipe */
```

bocon13: I think this is needed for PIPE_BUF, nevermind...

pudelkoM: We could remove the `static_assert`, though.

stratum/hal/lib/common/hal.cc
```diff
@@ -366,5 +390,19 @@ ::util::Status Hal::ProcmonCheckin() {
   return ::util::OkStatus();
 }
 
+void* Hal::SignalWaiterThreadFunc(void*) {
+  int signal_value;
+  if (read(Hal::pipe_read_fd_, &signal_value, sizeof(signal_value)) !=
+      sizeof(signal_value)) {
+    LOG(ERROR) << "Error reading signal from pipe: " << strerror(errno);
+    return nullptr;
```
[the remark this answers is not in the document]
pudelkoM: I agree, this check was before we knew `read` is atomic.

stratum/hal/lib/common/hal.cc
bocon13: I guess it doesn't hurt, but it also seems a bit unnecessary. The value in the docs is >> what we use, and it's been that way for decades.

stratum/hal/lib/common/hal.cc
pudelkoM: Then we should also rename `signal_waiter_tid_` and `SignalWaiterThreadFunc` to keep it consistent.

bocon13: On second thought, I think signal waiter makes more sense. Let's leave this as is.

stratum/hal/lib/common/hal.cc
```diff
@@ -366,5 +391,21 @@ ::util::Status Hal::ProcmonCheckin() {
...
 
+void* Hal::SignalWaiterThreadFunc(void*) {
+  int signal_value;
+  int ret = read(Hal::pipe_read_fd_, &signal_value, sizeof(signal_value));
+  if (ret == 0) {  // Pipe has been closed.
+    return nullptr;
+  } else if (ret < 0) {
+    LOG(ERROR) << "Error reading signal from pipe: " << strerror(errno);
+    return nullptr;
+  }
```
bocon13, line 402:
```suggestion
  } else if (ret != sizeof(signal_value)) {
      LOG(ERROR) << "Error reading signal from pipe: ";
      return nullptr;
    }
```

bocon13: Not sure this is necessary because I'm not sure what else would be written to the pipe, but we could include this for the sake of error handling completeness.

bocon13 (dismissed):

bocon13 (approved):
LGTM, thanks!

PR #657, opened 2021-04-27
Remove 28000 as default gRPC port

### LINKED ISSUE #499: Remove port 28000 as default
It might also be time to remove port 28000 as default.

_Originally posted by @bocon13 in https://github.com/stratum/stratum/issues/497#issuecomment-756932910_

pudelkoM:
It has been deprecated in favor of 9559 for some time and is now removed.

Closes #499


bocon13:
Thanks for taking care of this!

Issue #656, opened 2021-04-27
 How to manage device specific changes?(Adding support for new device)


oleksandrn-imt:
Hello,

I work at InterfaceMastersTechnologies and right now I'm
adding support for our https://interfacemasters.com/products/switches/10g-40g/tahoe-2624/ device(barefoot based),
but I have few questions about how to manage project-specific changes:

1) We use custom BSP, and for that, I made the following change:
    modified:   bazel/external/bfsde.BUILD

 ```
pkg_tar_with_symlinks(
     name = "bf_library_files",
     srcs = glob([
         "barefoot-bin/lib/bfshell_plugin_*.so*",
         "barefoot-bin/lib/libavago.so*",
         "barefoot-bin/lib/libbfsys.so*",
         "barefoot-bin/lib/libbfutils.so*",
         "barefoot-bin/lib/libdriver.so*",
         "barefoot-bin/lib/libdru_sim.so*",
         "barefoot-bin/lib/libpython3.4m.so*",
+        "barefoot-bin/lib/libtahoeimt_driver.so*",
+        "barefoot-bin/lib/libpltfm_mgr.so*",
+        "barefoot-bin/lib/libpltfm_driver.so*",
     ]),
     mode = "0644",
     package_dir = "/usr",
     strip_prefix = "barefoot-bin",
 )
```

The question is how do you handle such device-specific changes?
Or the only way to handle such changes is to keep them in our repo?

2) The same question for these files, which are also device specific in my opinion:
modified:   stratum/hal/bin/barefoot/deb/stratum.flags
        
```
diff --git a/stratum/hal/bin/barefoot/deb/stratum.flags b/stratum/hal/bin/barefoot/deb/stratum.flags
index 5e5919fe..c230d7e2 100644
--- a/stratum/hal/bin/barefoot/deb/stratum.flags
+++ b/stratum/hal/bin/barefoot/deb/stratum.flags
@@ -1,2 +1,4 @@
--bf_switchd_cfg=/usr/share/stratum/tofino_skip_p4_no_bsp.conf
+-bf_switchd_cfg=/usr/share/stratum/tofino_skip_p4.conf
 -bf_switchd_background=true
+-enable_onlp=false
+-bf_sim
```
        
modified:   stratum/hal/bin/barefoot/deb/stratum_bf.env
        
```
diff --git a/stratum/hal/bin/barefoot/deb/stratum_bf.env b/stratum/hal/bin/barefoot/deb/stratum_bf.env
[...]

PR #661, opened 2021-04-28
Managing device specific changes

oleksandrn-imt:
Per our discussion in #656 
This is a list of changes I did to add a new device, pure hardcode since there is no way to handle this right now. Open for discussions on how to handle such changes.


pudelkoM:
Thanks for opening the PR!

We reviewed parts of it internally and concluded that these changes are fairly specific to the Tahoe switch. Until we solve the story around general BSP support, we have to hold off on merging.

Happy to discuss next steps via email or slack.

PR #662, opened 2021-04-30
Improve signal handling code

pudelkoM:
This PR improves the signal handing code by properly saving and restoring `errno` during the handler and creating a helper functions for the pipe creation.

See: https://www.gnu.org/software/libc/manual/html_node/POSIX-Safety-Concepts.html

### DIFF
--- a/stratum/hal/lib/common/hal.cc
+++ b/stratum/hal/lib/common/hal.cc
@@ -5,7 +5,6 @@
 #include "stratum/hal/lib/common/hal.h"
 
 #include <limits.h>
-#include <unistd.h>
 
 #include <utility>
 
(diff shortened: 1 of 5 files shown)


stratum/hal/lib/common/hal.cc
```diff
@@ -54,8 +54,10 @@ void SignalRcvCallback(int value) {
   static_assert(sizeof(value) <= PIPE_BUF,
                 "PIPE_BUF is smaller than the number of bytes that can be "
                 "written atomically to a pipe.");
+  int saved_errno = errno;
   // No reasonable error handling possible.
   write(Hal::pipe_write_fd_, &value, sizeof(value));
+  errno = saved_errno;
```
bocon13, line 59: Add a note to why we are doing this

PR #666, opened 2021-05-03
Relax locking in BfrtNode for concurrent Reads and StreamMessageRequests

pudelkoM:
This change allows concurrent handling of P4RT reads and StreamMessageRequests (e.g. PacketOuts). Long-running read requests no longer block handling of PacketOuts, which can confuse controllers.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_node.cc
+++ b/stratum/hal/lib/barefoot/bfrt_node.cc
@@ -228,7 +228,7 @@ BfrtNode::~BfrtNode() = default;
     const ::p4::v1::ReadRequest& req,
     WriterInterface<::p4::v1::ReadResponse>* writer,
     std::vector<::util::Status>* details) {
-  absl::WriterMutexLock l(&lock_);
+  absl::ReaderMutexLock l(&lock_);
   CHECK_RETURN_IF_FALSE(req.device_id() == node_id_)
       << "Request device id must be same as id of this BfrtNode.";
   if (!initialized_ || !pipeline_initialized_) {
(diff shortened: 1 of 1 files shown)


bocon13 (approved):
LGTM.

I don't think `WriteForwardingEntries` updates any state protected by the lock. If we want to prevent concurrent write batches to the SDE, we should leave thing as is. Otherwise, we can also relax that lock.

pudelkoM:
> I don't think `WriteForwardingEntries` updates any state protected by the lock. If we want to prevent concurrent write batches to the SDE, we should leave thing as is. Otherwise, we can also relax that lock.

I'm still confused by the P4RT spec:

> A P4Runtime server will process the batch as follows:
>
> 1. Lock state (preventing new writes) and ...
> [...]
> 2. Unlock the state (allowing new writes);

https://s3-us-west-2.amazonaws.com/p4runtime/docs/master/P4Runtime-Spec.html#sec-batch-processing

So no concurrent writes/writes or writes/reads.

But then:

>13.4. Parallelism of Read and Write Requests
>
> A P4Runtime server may be implemented to serve at **most one** ReadRequest or WriteRequest message at a time, sequentially. It may also serve **multiple requests in parallel**, and it is expected that some client software would be easier to implement with good performance characteristics if a server did so. 

https://s3-us-west-2.amazonaws.com/p4runtime/docs/master/P4Runtime-Spec.html#sec-parallelism-of-read-and-write-requests


For now we should go with the same behavior as in BcmNode: https://github.com/stratum/stratum/blob/6c868be10a60072c618afa0cef8bf979bfa142b6/stratum/hal/lib/bcm/bcm_node.cc#L163-L167

bocon13:
Thanks for clarifying that. I'm okay with write lock on Write and read lock on Read. This will allow multiple parallel reads, but also that writes and reads are sequentially ordered and that writes can't happen concurrently at the Stratum layer.

Let's go with the PR as is.

PR #672, opened 2021-05-04
Replace `TableData::GetCounterData` with byte and packet specific versions

### LINKED ISSUE #668: Stratum prints error about reading counters that doesn't support by the table
**Describe the bug**
After update to SDE 9.5.0, and use the ONOS to control it(with TOST) Stratum start showing the following logs:

```
2021-05-04 02:06:52.818523 BF_BFRT ERROR - getDataField:1812 Data field name $COUNTER_SPEC_BYTES actionID 28485346 not found
2021-05-04 02:06:52.818574 BF_BFRT ERROR - getDataField:1812 Data field name $COUNTER_SPEC_PKTS actionID 28485346 not found
```

**Environment**
Use container image: `stratumproject/stratum-bfrt:9.5.0` (build with latest main 14c99349)

Switch: WedgeBF-32QS with ONL 4.14.49

**To Reproduce**

Load fabric-tna fabric-int pipeline with SDE 9.5.0 p4c output by using p4runtime-shell and run following:

```
te = table_entry['FabricIngress.forwarding.routing_v4']()
te.is_default = True
te.read(lambda x: print(x))
```

**Expected behavior**

 - Stratum should not show the error in the description above

pudelkoM:
This prevents doing the field name lookup for table data which has none or only one of the fields.

Fixes #668

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager.cc
@@ -486,10 +486,33 @@ std::unique_ptr<BfrtTableManager> BfrtTableManager::CreateInstance(
     result.mutable_action()->set_action_profile_group_id(selector_group_id);
   }
 
-  // Counter data
-  uint64 bytes, packets;
-  if (table_data->GetCounterData(&bytes, &packets).ok()) {
-    if (request.has_counter_data()) {
+  // Counter data, if applicable.
+  // We check for direct resources, since some calls to this function set the
+  // counter_data field for tables without direct counters.
+  if (request.has_counter_data() && table.direct_resource_ids_size()) {
+    for (const int id : table.direct_resource_ids()) {
+      uint64 bytes = 0, packets = 0;
+      ASSIGN_OR_RETURN(auto direct_counter,
+                       p4_info_manager_->FindDirectCounterByID(id));
+      switch (direct_counter.spec().unit()) {
+        case ::p4::config::v1::CounterSpec::BYTES:
+          RETURN_IF_ERROR(table_data->GetByteCounter(&bytes));
+          break;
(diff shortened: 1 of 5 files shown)


stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -486,10 +486,33 @@ ::util::StatusOr<::p4::v1::TableEntry> BfrtTableManager::BuildP4TableEntry(
...
-  if (table_data->GetCounterData(&bytes, &packets).ok()) {
-    if (request.has_counter_data()) {
+  // Counter data, if applicable.
+  // We check for direct resources, since some calls to this function set the
+  // counter_data field for tables without direct counters.
+  if (request.has_counter_data() && table.direct_resource_ids_size()) {
+    for (const int id : table.direct_resource_ids()) {
+      uint64 bytes = 0, packets = 0;
+      ASSIGN_OR_RETURN(auto direct_counter,
+                       p4_info_manager_->FindDirectCounterByID(id));
```
pudelkoM, line 496: This will fail if there is more than one direct resource on a table. Can this happen? Should we check this here or on pipeline push?

Yi-Tseng: direct meter + direct counter?

pudelkoM:
Check if we can use the same approach as in #686

stratum/hal/lib/barefoot/bfrt_table_manager.cc
Yi-Tseng: maybe we don't return if the `id` is not a direct counter id, but just skip/continue if it is not for direct counter?

pudelkoM:
Superseeded by #698

PR #677, opened 2021-05-05
Add option to make Stratum-bfrt return canonical bytestrings

pudelkoM:
This PR adds a new option `incompatible_enable_bfrt_legacy_bytestring_responses` to make Stratum-bfrt return canonical P4Runtime byte strings. For now the flag defaults to `true`, which means that the current behavior remains the same, that is, padded strings. In the future we will flip and eventually remove the legacy behavior.

TODO:

- [x] some basic testing

### DIFF
--- a/stratum/hal/lib/barefoot/bf_sde_wrapper.cc
+++ b/stratum/hal/lib/barefoot/bf_sde_wrapper.cc
@@ -44,6 +44,10 @@ int switch_pci_sysfs_str_get(char* name, size_t name_size);
 
 DEFINE_string(bfrt_sde_config_dir, "/var/run/stratum/bfrt_config",
               "The dir used by the SDE to load the device configuration.");
+DEFINE_bool(incompatible_enable_bfrt_legacy_bytestring_responses, true,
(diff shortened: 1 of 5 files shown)


stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -44,6 +44,10 @@ int switch_pci_sysfs_str_get(char* name, size_t name_size);
 
 DEFINE_string(bfrt_sde_config_dir, "/var/run/stratum/bfrt_config",
               "The dir used by the SDE to load the device configuration.");
+DEFINE_bool(incompatible_bfrt_legacy_bytestring_responses, false,
```
pudelkoM, line 47:
```suggestion
DEFINE_bool(incompatible_bfrt_legacy_bytestring_responses, true,
```

So we can merge this now without changing the default behavior.

pudelkoM:
Basic PTF test with the Fabric-tna `FabricOptimizedFieldDetectorTest` succeeded.

bocon13 (commented):
looks reasonable to me

PR #678, opened 2021-05-06
add SetSfpFrequency function to onlp_wrapper

juraul1:
Some comments on the files:

Overall, I've never learnt to write in C++ so it might be some syntax issues.

**onlp_warrper.cc:**
- <stdint.h> is for the uint8_t type
- I'm using GetSfpInfo function because I want to check if the SFP is tunable, and this information is located in sfp_info. However I don't know how to use that function direclty inside SetSfpFrequency so I just copied and pasted the code.
- I check the TYPE, and exit if it's not an SFP/SFP+. This should work but in case it doesn't, I know which register is responsible for the TYPE so I can access it using onlp_i2c_readb as well.
- Then a lot of lines of code for the ports allocation. All those functions are defined in the [APSN-ONL2 repository](https://github.com/NetworkGradeLinux/APSN-ONL2/blob/main/packages/base/any/onlp/src/onlplib/module/src/i2c.c), including the driver definition. The MUX selection is mandatory if we want to access the register with onlp_i2c_* functions. Therefore, I added an input int port_number which is the port number on the Tofino switch.
- I guess the switch (port-number) function can be located somewhere else.
- onlp_i2c_write function does not always give an error message, even if you can't write the value, that's why I put some check just after them to see if it has been written correclty.
- Grid spacing, first frequency Thz and first frequency Ghz are written on 2 registers (bytes) so I retreive the uint16_t value and convert it into Hz int value.
- The value "0.1" in first_frequency is the unit given by SFF-8690 document.
- in onlp_i2c_write, the desired value can be in hex or int format. Both are working.
- At the end, I go back to initial values so that GetSfpInfo will still be the correct values from page 1 and so that we can access other SFPs without issue.

**onlp_wrapper.h**
- not sure at all of what I added.

**onlp_wrapper.mock.h**
- Do I need to add something here? I put METHOD3 because I have 3 variables but I have not idea if it is correct.

In order to expose the phal API to the switch:
Which file should I modify? I did modify bf_chassis_manager.cc as the code is for Barefoot tofino switches but I also remember that you were taking about onlp_phal.cc or something.

**bf_chassis_manager.cc**
- I put onlp_interface_->, knowing that this might not work on this function. I wanted first to put it in onlp_phal.cc but nothing is defined there so I don't know where to call SetSfpFrequency and I don't know the difference between onlp_phal and bf_chassis_manager

### DIFF
--- a/stratum/hal/lib/phal/onlp/onlp_wrapper.cc
+++ b/stratum/hal/lib/phal/onlp/onlp_wrapper.cc
@@ -7,6 +7,7 @@
 #include <dlfcn.h>
 
 #include <string>
+#include <stdint.h>
 
 #include "absl/memory/memory.h"
 #include "absl/strings/strip.h"
@@ -127,6 +128,9 @@ template <typename T>
   LOAD_SYMBOL(onlp_led_mode_set);
   LOAD_SYMBOL(onlp_led_char_set);
   LOAD_SYMBOL(onlp_psu_info_get);
+  LOAD_SYMBOL(onlp_i2c_mux_select);
+  LOAD_SYMBOL(onlp_i2c_write);
+  LOAD_SYMBOL(onlp_i2c_readb);
 #undef LOAD_SYMBOL
 
   CHECK_RETURN_IF_FALSE(ONLP_SUCCESS(onlp_functions_.onlp_sw_init(nullptr)))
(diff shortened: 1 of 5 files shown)


bocon13:
Thanks for the PR, @juraul1!

We'll try to take a look at this soon and provide some feedback.

bocon13:
A few of us (@pudelkoM, @Yi-Tseng, and myself) had a chance to look at the PR, and one thing that we still need to figure out is how to abstract some of the platform specific (e.g. BF6064-X) details from the common ONLP implementation.

I'm not sure if there's a way to look up the mux map or whether we should consider adding this information to a configuration file, so that we can make the code generic.

The other thing we took a look at was the BF BSP interface to see if some of these details are better abstracted. I don't have a clear answer yet, but it's another thing to consider.

Perhaps we could find some time later this week to talk about some of these things and answer your questions from above?

stratum/hal/lib/phal/onlp/onlp_wrapper.cc
```diff
@@ -156,6 +160,227 @@ ::util::StatusOr<SfpInfo> OnlpWrapper::GetSfpInfo(OnlpOid oid) const {
...
+      channel_mb = 6;
+      break;
+  default: //or case 20 || 51:
+      mb_mux.devaddr = 0x77;
+      channel_mb = 7;
+      break;
+  }
+  CHECK_RETURN_IF_FALSE(
+       ONLP_SUCCESS(onlp_functions_.onlp_i2c_mux_select(mb_mux, channel_mb)))
+       << "Failed to set MB MUX.";
```
bocon13, line 333: Finding the right SFP port; should be done by the Platform API

stratum/hal/lib/phal/onlp/onlp_wrapper.cc
```diff
@@ -156,6 +160,227 @@ ::util::StatusOr<SfpInfo> OnlpWrapper::GetSfpInfo(OnlpOid oid) const {
...
+  // Change the page on slave 0x51 to access page 2
```
bocon13, line 334: Check the SFF version to make sure we have a tunable SFP

stratum/hal/lib/phal/onlp/onlp_wrapper.cc
```diff
@@ -156,6 +160,227 @@ ::util::StatusOr<SfpInfo> OnlpWrapper::GetSfpInfo(OnlpOid oid) const {
...
+  // Check if it has been done correctly
+  if (onlp_functions_.onlp_i2c_readb(0,0x51,0x91) != channel_number) {
+      fprintf(stderr, "Error: Can not write the desired frequency.\n");
+      exit(0);
+  }
+
+  // Put the page register back to 1
+  CHECK_RETURN_IF_FALSE(
+       ONLP_SUCCESS(onlp_functions_.onlp_i2c_write(0,0x51,0x7f,1,0x01,0)))
+       << "Failed to set CPU MUX.";
```
bocon13, line 376: Tuning the SFP... defined in the SFF spec, could be implemented in Stratum

stratum/hal/lib/phal/onlp/onlp_wrapper.cc
```diff
@@ -156,6 +160,227 @@ ::util::StatusOr<SfpInfo> OnlpWrapper::GetSfpInfo(OnlpOid oid) const {
...
+      mb_mux.devaddr = 0x77;
+      channel_mb = 7;
+      break;
+  }
+  CHECK_RETURN_IF_FALSE(
+       ONLP_SUCCESS(onlp_functions_.onlp_i2c_mux_select(mb_mux, channel_mb)))
+       << "Failed to set MB MUX.";
+  // Change the page on slave 0x51 to access page 2
+  CHECK_RETURN_IF_FALSE(
+       ONLP_SUCCESS(onlp_functions_.onlp_i2c_write(0,0x51,0x7f,1,0x2,0)))
```
bocon13, line 336: 0 is bus, mostly consistent
0x51 is the register (0x50 is for serial number/tmp/etc., 0x51 is register for accessing tunable parameters)
0x7f is the address, defined in the spec

stratum/hal/lib/phal/onlp/onlp_wrapper.cc
```diff
@@ -127,6 +128,9 @@ ::util::Status OnlpWrapper::Initialize() {
   LOAD_SYMBOL(onlp_led_mode_set);
   LOAD_SYMBOL(onlp_led_char_set);
   LOAD_SYMBOL(onlp_psu_info_get);
+  LOAD_SYMBOL(onlp_i2c_mux_select);
```
bocon13, line 131: LOAD_SYMBOL_IF_PRESENT(onlp_sfp_frequency_set);
LOAD_SYMBOL_IF_PRESENT(onlp_sfp_mux?_get);

Nullptr if not present, then check before using the function

stratum/hal/lib/phal/onlp/onlp_wrapper.cc
```diff
@@ -156,6 +160,227 @@ ::util::StatusOr<SfpInfo> OnlpWrapper::GetSfpInfo(OnlpOid oid) const {
...
+
+  // Change the channel number of the SFP
+  CHECK_RETURN_IF_FALSE(
+       ONLP_SUCCESS(onlp_functions_.onlp_i2c_write(0,0x51,0x91,1,channel_number,0)))
+       << "Failed to set CPU MUX.";
+
+  // Check if it has been done correctly
+  if (onlp_functions_.onlp_i2c_readb(0,0x51,0x91) != channel_number) {
+      fprintf(stderr, "Error: Can not write the desired frequency.\n");
+      exit(0);
```
bocon13, line 370: prefer returning Status as error

bocon13:
You can use `uint8`

stratum/hal/lib/phal/onlp/onlp_wrapper.cc
```diff
@@ -156,6 +160,227 @@ ::util::StatusOr<SfpInfo> OnlpWrapper::GetSfpInfo(OnlpOid oid) const {
...
+  } else {
+      fprintf(stderr, "Error: Ports limit exceeded\n"); //Port number has to be between 1 and 64 for BF6064X switch
+      exit(0);
+  }
+  CHECK_RETURN_IF_FALSE(
+      ONLP_SUCCESS(onlp_functions_.onlp_i2c_mux_select(cpu_mux, channel_cpu)))
+      << "Failed to set CPU MUX.";
+
+  switch (port_number) {
+  case 1 || 33:
```
pudelkoM, line 202: I'm fairly sure this just evaluates to `1`.
Use fall through cases instead.

juraul1:
Hi @bocon13 @pudelkoM @Yi-Tseng 
You should see some changes. I put the switch specific mapping in the vendor repository, and you can check the [PR](https://github.com/APS-Networks/OpenNetworkLinuxv2/pull/1). 
The rest of the code in onlp_wrapper is common to all SFPs, according to the SFF specs.

I haven't got the time to look deep into the BSP files yet but I suggest you have a look on the PR and then we can arrange a call and discuss about that and about the BSP option.

PR #681, opened 2021-05-07
Use `minloglevel` instead of `stderrthreshold`

pudelkoM:
`stderrthreshold` only works when logging to files is enabled (`logtostderr == false`). We disable logging to files by default. This change fixes this throughout the code base.


stratum/p4c_backends/common/build_defs.bzl
```diff
@@ -72,10 +72,6 @@ def _generate_p4c_stratum_config(ctx):
             "--p4c_annotation_map_files=" + annotation_map_files,
             "--slice_map_file=" + ctx.file.slice_map.path,
             "--target_parser_map_file=" + ctx.file.parser_map.path,
-            "--colorlogtostderr",
-            "--stderrthreshold=1",
-            "--logtostderr",
-            "--v=0",
```
pudelkoM, line 78: Don't set values if they're the same as the defaults.

PR #686, opened 2021-05-11
Prevent DirectCounterEntry modifies from changing the action data

### LINKED ISSUE #685: Action parameter(s) will be reset if set the direct counter
**Describe the bug**

When sending write request to set the direct counter, the action parameter of that table entry will also be reset to default value(e.g., zero)

**Environment**

 - Stratum bfrt container image
   - SDE 9.5.0 
   - Commit: `155588e6031853f0c0122fec823dba91571ac3a8`
 - Tofino Model

**To Reproduce**

Use P4Runtime Shell with fabric-tna:

```python3
te = table_entry['FabricIngress.stats.flows'](action='FabricIngress.stats.count')
te.priority = 10
te.match['ig_port'] = '1'
te.action['flow_id'] = '1'
te.insert()
te.read(lambda x: print(x)) # value of parameter is 1

dce = direct_counter_entry['FabricIngress.stats.flow_counter']
dce.table_entry.match['ig_port'] = '1'
dce.table_entry.priority = 10
dce.byte_count = 0
dce.packet_count = 0
dce.modify()

te.read(lambda x: print(x)) # value of parameter becomes 0
```

**Expected behavior**

The value of parameter should still be `1`

pudelkoM:
This change removes `SetOnlyCounterData` which erroneously *did* modify the action data and replaces it with the existing `SetCounterData` function. We improve the later by only setting BfRt table data fields that exists, e.g., in byte-only counters.

Fixes #685

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager.cc
@@ -685,8 +685,9 @@ std::unique_ptr<BfrtTableManager> BfrtTableManager::CreateInstance(
   RETURN_IF_ERROR(BuildTableKey(table_entry, table_key.get()));
 
   // Fetch existing entry with action data. This is needed since the P4RT
-  // request does not provide the action (id), but the SDE requires it in the
-  // later modify call.
+  // request does not provide the action ID and data, but we have to provide the
+  // current values in the later modify call to the SDE, else we would modify
+  // the table entry.
   RETURN_IF_ERROR(bf_sde_interface_->GetTableEntry(
       device_, session, table_id, table_key.get(), table_data.get()));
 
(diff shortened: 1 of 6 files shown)


pudelkoM:
@Yi-Tseng could you give this PR a try? I'm fairly certain I got it, but a second look would not hurt.

stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -685,8 +685,9 @@ ::util::Status BfrtTableManager::WriteDirectCounterEntry(
   RETURN_IF_ERROR(BuildTableKey(table_entry, table_key.get()));
 
   // Fetch existing entry with action data. This is needed since the P4RT
-  // request does not provide the action (id), but the SDE requires it in the
-  // later modify call.
+  // request does not provide the action (id) and action data, but we have to
```
pudelkoM, line 688:
```suggestion
  // request does not provide the action ID and data, but we have to
```

Yi-Tseng (approved):
LGTM, and the PTF test pass with this fix

pudelkoM:
> LGTM, and the PTF test pass with this fix

Thanks for checking!

PR #690, opened 2021-05-14
Implement `/state/last-change` node by referencing `/oper-status`

pudelkoM:
This PR implements the OpenConfig `/interfaces/interface[name=<name>]/state/last-change` path by monitoring the `/oper-status` node of the same interface and watching for port changes.

Alternatives:
- Add `uint64 last_change` field to the common `OperStatus` message and have the ChassisManagers handle it, possibly with help from the SDK

We should keep in mind that `counters/last-clear` will require a similar solution.

### DIFF
--- a/stratum/hal/lib/common/yang_parse_tree_paths.cc
+++ b/stratum/hal/lib/common/yang_parse_tree_paths.cc
@@ -914,11 +914,63 @@ void SetUpRoot(TreeNode* node, YangParseTree* tree) {
 
 ////////////////////////////////////////////////////////////////////////////////
 // /interfaces/interface[name=<name>]/state/last-change
-void SetUpInterfacesInterfaceStateLastChange(TreeNode* node) {
-  auto poll_functor = UnsupportedFunc();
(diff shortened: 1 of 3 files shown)


stratum/hal/lib/common/yang_parse_tree_paths.cc
```diff
@@ -914,11 +914,63 @@ void SetUpRoot(TreeNode* node, YangParseTree* tree) {
...
+            }
+          }
+          return false;
+        });
+    RETURN_IF_ERROR(
+        (oper_status_node->GetOnPollHandler())(event, &inline_stream));
+    if (last_state != new_state) {
+      now = absl::ToUnixNanos(absl::Now());
+      last_state = new_state;
+    }
```
pudelkoM, line 949: This logic has a accuracy problem if there are no subscribers and this path is only polled infrequently. The timestamp then is updated on the actual poll requests, making it look like the port flipped just before.

pudelkoM:
Superseded by #700 and #707

PR #691, opened 2021-05-16
stratum-bmv2: temporarily revert default canonical bytestrings

enen92:
Hey @pudelkoM

Unfortunately https://github.com/stratum/stratum/pull/660 introduced a few regressions on the use of ONOS with stratum-bmv2. Although packet-in/packet-out and some flow rules work well, simple rules such as `match IN_PORT=1, OUTPUT=2` fail to install and are kept on the pending add state forever (removed and inserted again at each cycle). This happens because, due to the default use of canonical bytestrings in PI, the received bytes for the port field are different than the ones of the inserted rule. Since ONOS compares both buffers, the flow rule is flagged as being different than the one installed. For a better understanding of the problem please check: https://gerrit.onosproject.org/c/onos/+/24608
Support for canonical bytestrings in ONOS is WIP and available in https://gerrit.onosproject.org/c/onos/+/24609.

Hence, while ONOS doesn't support canonical bytestrings in some stable version it is better to partially revert the above PR. This PR moves the PI commit a bit early in history, to the first commit before canonical bytestrings were enforced as default. Also adds a bit of documentation (similar to one of your recent PRs) to advise disabling canonical bytestrings on the client for now. There aren't pretty much any changes in PI since this point in history that are not related to canonical bytestrings anyway.
So, we can still benefit from the fixes for direct resources (meter) reading introduced in my latest PR and use ONOS as the control plane without any unexpected behaviour.

Sorry for the inconvenience/regression and thanks a lot for your time.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 ARG BAZELISK_VERSION=1.8.0
-ARG PI_COMMIT=b2760a818e0b8ade5864604d29b3008a684c6d5f
+ARG PI_COMMIT=35ac4fc80d5a3560d171dc8aa60d104bfa31d37a
 ARG BMV2_COMMIT=6dfd5b41426310eb081fedc8fee05b675f780e31
 ARG JDK_URL=https://mirror.bazel.build/openjdk/azul-zulu11.29.3-ca-jdk11.0.2/zulu11.29.3-ca-jdk11.0.2-linux_x64.tar.gz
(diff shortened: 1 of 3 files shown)


pudelkoM:
Thanks for the PR!

After some internal discussion, we decided to try to fix ONOS first. Unless we hit major road blocks, canonical byte string should not be a problem anymore soon.

enen92:
I noticed a patch to ensure read-write simmetry for p4runtime (similar to mine but extended to other match fields and actions) just went in for ONOS. I am closing this as it can be considered solve.
Once again, thanks a lot for your time and for addressing this issue quickly

PR #695, opened 2021-05-18
add tna profile member to p4runtime session lib

eth0xFEED:


stratum/pipelines/stratum-tna/stratum_tna_test.cc
```diff
@@ -107,6 +107,37 @@ class StratumTnaTest : public P4RuntimeFixture {
...
+              value: "\x00"
+            }
+            params {
+              param_id: {StratumIngress.fwd_route.dmac}
+              value: "\x00"
+            }
+          }
+      )PROTO");
+      // TODO what is the method membership space here ???? (renner)
+    // ap_entry.mutable_member_id()->set_value(member_id);
```
eth0xFEED, line 129: @pudelkoM is there an easy way to check type method membership for types in the cc generated code ? I wasnt able to untangle what the contents of an object like 
```
::p4::v1::ActionProfileMember ap_entry =
```
would be ?

pudelkoM: If your SDE supports auto-completion for generated code, you can cycle through it until you find what you need.
Reading the proto source also gives a good overview which fields are available: https://github.com/p4lang/p4runtime/blob/main/proto/p4/v1/p4runtime.proto#L286
Together with the [C++ generated code guide](https://developers.google.com/protocol-buffers/docs/reference/cpp-generated) I usually manage.

eth0xFEED: forgot i always used to use the go generated code guide.... thanks for the reminder now see that in fact there is an exact set for that field thank you sir !!
```
$ ag set_member_id
p4runtime.pb.h
4500:  void set_member_id(::PROTOBUF_NAMESPACE_ID::uint32 value);
```

PR #697, opened 2021-05-19
Add deadline param to `SwitchInterface::ReadForwardingEntries`

pudelkoM:
This PR makes `SwitchInterface::ReadForwardingEntries` deadline aware so that long running requests can be aborted early without handling all entities.

However, while this allows checking for a deadline, I'm not sure if `TryCancel` attempts from the client are covered by this.
The [gRPC error handling guide](https://grpc.io/blog/deadlines/#checking-deadlines) says to check `context->IsCancelled()`, which includes deadline, network errors and cancelations. We could pass this down as a functor?

### DIFF
--- a/stratum/hal/lib/common/p4_service.cc
+++ b/stratum/hal/lib/common/p4_service.cc
@@ -337,8 +337,13 @@ void LogReadRequest(uint64 node_id, const ::p4::v1::ReadRequest& req,
   ServerWriterWrapper<::p4::v1::ReadResponse> wrapper(writer);
   std::vector<::util::Status> details = {};
   absl::Time timestamp = absl::Now();
-  ::util::Status status =
-      switch_interface_->ReadForwardingEntries(*req, &wrapper, &details);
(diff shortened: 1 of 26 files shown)


stratum/hal/lib/common/p4_service.cc
```diff
@@ -337,8 +337,13 @@ ::grpc::Status P4Service::Read(
   ServerWriterWrapper<::p4::v1::ReadResponse> wrapper(writer);
   std::vector<::util::Status> details = {};
   absl::Time timestamp = absl::Now();
-  ::util::Status status =
-      switch_interface_->ReadForwardingEntries(*req, &wrapper, &details);
+  // auto deadline = min(context->deadline() - 100ms), 0);
+  // TODO(max): check where deadline comes from, and relation to isCancelled()
+  absl::Time deadline = absl::FromChrono(context->deadline());
+  LOG(ERROR) << absl::FormatTime(deadline);
+  auto t = absl::InfiniteFuture();
```
pudelkoM, line 344: cleanup

PR #733, opened 2021-06-04
Don't discard `last_change` timestamp on `BfChassisManager::PushChassisConfig`

pudelkoM:
This PR changes the handling of old `last_change` timestamp on `PushChassisConfig`. We no longer discard it, but keep the last known value. This prevents the return of 0 timestamps between a config push and port event.


pierventre:
Verified - +1 from me

PR #741, opened 2021-06-09
Expose QoS config on Tofino

pudelkoM:


stratum/hal/lib/common/common.proto
```diff
@@ -692,10 +692,89 @@ message TofinoConfig {
...
+      PRIO_6 = 7;
+      PRIO_7 = 8;
+    }
+    message PoolConfig {
+      ApplicationPool pool = 1;  // required
+      uint32 pool_size = 2;
+      bool enable_color_drop = 3;
+    }
+    message PpgConfig {
+      uint32 sdk_port = 1;  // required
```
pudelkoM, line 738: consider supporting singleton port ids

pudelkoM: Since we currently pass the entire config to the SdeWrapper, which has no knowledge about singleton ports, supporting those is infeasible.

PR #745, opened 2021-06-16
Add a parser test for Tofino port maps and fix them

### LINKED ISSUE #744: File x86-64-netberg-aurora-750-r0/port_map.json has syntax error
File stratum/hal/config/x86-64-netberg-aurora-750-r0/port_map.json has a JSON syntax error on line 243:
    "rx_pn_swap": ,
The value is missing before the comma. Not sure what the correct value should be, so cannot submit the fix myself.

NOTE: stratum/hal/config/x86-64-netberg-aurora-710-r0/port_map.json also has the following issues:
- Missing the terminal comma on line 554.
- Missing a value on line 618. The value for "tx_lane" should be 1.

Found using a JSON validation tool. Does not break the build. But these should be fixed.

pudelkoM:
This PR adds an additional validation check to the config files. Port map JSON files are now parsed to check for syntax errors. This change also fixes any existing errors.

Fixes #744


naradahess (approved):
Nice. Glad you have added JSON validation. Perhaps consider the same for YAML files (though you may need some options/exclusions).

PR #706, opened 2021-05-25
First stab at BSP support for Wedge switches

pudelkoM:
This PR adds support for BSP-backed Stratum binaries on Wedge switches.

Addl flag: `-bf_switchd_cfg=/usr/share/stratum/tofino_skip_p4.conf`

Logs:

```
bf_switchd: processing P4 configuration...
P4 profile for dev_id 0
  p4_name: dummy
    libpd:
    libpdthrift:
    context:
    config:
  Agent[0]: /usr/lib/libpltfm_mgr.so
  diag:
  accton diag:
  non_default_port_ppgs: 0
  SAI default initialize: 1
bf_switchd: library /usr/lib/libpltfm_mgr.so loaded
bf_switchd: agent[0] initialized
Health monitor started
Operational mode set to ASIC
Initialized the device types using platforms infra API
ASIC detected at PCI /sys/class/bf/bf0/device
ASIC pci device id is 16
bf_switchd: drivers initialized
Skipping P4 program load for dev_id 0
Setting core_pll_ctrl0=cd44cbfe
```

PR #748, opened 2021-06-21
Minimal BSP support for Tofino Wedge100BF switches

pudelkoM:
This PR adds initial support for BSP-backed Stratum binaries on Wedge switches and makes it the default on those devices.
As before, the selection can be overwritten with a flag: `-bf_switchd_cfg=/usr/share/stratum/tofino_skip_p4[_no_bsp].conf`

Closes #706


stratum/hal/bin/barefoot/docker/build-stratum-bf-container.sh
```diff
@@ -30,6 +30,7 @@ Additional environment variables:
     DOCKER_IMG: Docker image to use for building (Default: stratumproject/build:build)
     RELEASE_BUILD: Optimized build with stripped symbols (Default: false)
     BAZEL_CACHE: Path to Bazel cache (Default: <empty>)
+    BSP: Path to optional BSP package directory (Default: <empty>)
```
pudelkoM, line 33: This is an env var because this script does no support/use flag style arguments for the other parameters.

bazel/external/bfsde.BUILD
```diff
@@ -52,6 +52,11 @@ cc_library(
 
 pkg_tar_with_symlinks(
     name = "bf_library_files",
+    # Using a wildcard glob here to match the shared libraries makes this rule
+    # more generic than a normal source list, as it does not require that all
+    # targets are present, which is the case for non-BSP SDE builds. Extenting
```
bocon13, line 57: typo
```suggestion
    # targets are present, which is the case for non-BSP SDE builds. Extending
```

stratum/hal/bin/barefoot/README.build.md
```diff
@@ -287,8 +287,16 @@ information. **This is the recommended mode. No changes to the SDE needed.**
 
 Otherwise, you need to build & install the BSP. You will not be able to use
 the Stratum ONLP support. The exact instructions vary by the BSP vendor, here is
-how it works for the Wedge reference switch. Pass the BSP sources to the p4studio_build
-script with the `--bsp-path` flag.
+how it works for the Wedge reference switch. Either set the `BSP` environment
+variable before running the [`build-stratum-bf-container.sh`](#method-1-build-with-docker-in-one-shot) script:
```
bocon13, line 291: Line length
```suggestion
variable before running the
[`build-stratum-bf-container.sh`](#method-1-build-with-docker-in-one-shot) script:
```

stratum/hal/bin/barefoot/deb/start-stratum.sh
```diff
@@ -50,6 +50,15 @@ if ! grep hugetlbfs /etc/mtab | grep -q /mnt/huge; then
...
+   [[ "$PLATFORM" == "x86-64-accton-wedge100bf-32qs-r0" ]] || \
+   [[ "$PLATFORM" == "x86-64-accton-wedge100bf-65x-r0" ]]; then
+    BF_SWITCHD_CFG=/usr/share/stratum/tofino_skip_p4.conf
+else
+    BF_SWITCHD_CFG=/usr/share/stratum/tofino_skip_p4_no_bsp.conf
+fi
+
 # Set up port map for device
 PORT_MAP="/etc/stratum/$PLATFORM/port_map.json"
 if [ ! -f "$PORT_MAP" ]; then
```
bocon13, line 64: We can skip this step if we are using the BSP. But it also doesn't hurt to have it.

bocon13 (approved):
Looks reasonable and show allow others to build on it. Some minor nits.

stratum/hal/bin/barefoot/deb/start-stratum.sh
pudelkoM: Don't we still need the map if a user overrides the `bf_switchd_cfg` flag to non-bsp?

PR #754, opened 2021-06-24
Missing the conversion from bits to bytes

Yi-Tseng:
The second parameter of P4RuntimeByteStringToPaddedByteString is a byte, need
to use NumBitsToNumBytes to convert bits first before pass to it.

### DIFF
--- a/stratum/hal/lib/barefoot/bf_sde_wrapper.cc
+++ b/stratum/hal/lib/barefoot/bf_sde_wrapper.cc
@@ -2641,7 +2641,7 @@ namespace {
       table->dataFieldSizeGet(field_id, &data_field_size_bits));
   // The SDE expects a string with the full width.
   std::string value = P4RuntimeByteStringToPaddedByteString(
-      register_data, data_field_size_bits);
+      register_data, NumBitsToNumBytes(data_field_size_bits));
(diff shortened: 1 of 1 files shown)


pudelkoM (approved):

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -2641,7 +2641,7 @@ ::util::Status BfSdeWrapper::WriteRegister(
       table->dataFieldSizeGet(field_id, &data_field_size_bits));
   // The SDE expects a string with the full width.
   std::string value = P4RuntimeByteStringToPaddedByteString(
-      register_data, data_field_size_bits);
+      register_data, NumBitsToNumBytes(data_field_size_bits));
   RETURN_IF_BFRT_ERROR(table_data->setValue(
       field_id, reinterpret_cast<const uint8*>(value.data()), value.size()));
```
pudelkoM, line 2646: Maybe this entire block can be replaced by `SetParam`?

pudelkoM: It can not, because it's a raw `BfRtTableData` not a wrapped `TableData`.

PR #757, opened 2021-06-25
Translate P4 table ID to BfRt ID when reading registers

Yi-Tseng:
The register id(and name) in bfrt.json can be different from the one in p4info, here is an example from the P4 info

bfrt.json

```
    {
      "name" : "pipe.FabricEgress.int_egress.queue_report_quota",
      "id" : 2342202179,
...
```

p4info

```
registers {
  preamble {
    id: 379267907
    name: "FabricEgress.int_egress.queue_report_quota"
    alias: "queue_report_quota"
  }
...
```


pudelkoM:
that comment lol

Thanks!

pudelkoM (approved):

PR #760, opened 2021-06-28
Make SerDes config optional on Stratum-bcm

pudelkoM:
We currently don't have any serdes database and on platforms with poor ONLP support, `GetFrontPanelPortInfo` will fail. This change make Configuring SerDes optional, i.e. error are not blocking.


abhilashendurthi (approved):
LGTM

abhilashendurthi (approved):

PR #768, opened 2021-07-01
Add BSP Debian dependencies to Stratum-bf[rt] packages

pudelkoM:
The Wedge BSP (more specifically, the accton driver) requires additional libraries at runtime. This PR adds them to the dependencies of the Stratum Debian package.

See: https://github.com/stratum/fabric-tna/pull/294

### DIFF
--- a/stratum/hal/bin/barefoot/BUILD
+++ b/stratum/hal/bin/barefoot/BUILD
@@ -166,6 +166,14 @@ pkg_tar(
     ],
 )
 
+# Extra SDE runtime dependencies when using the BSP.
+# TODO(max): consider packaging these deps with the SDE itself somehow.
+sde_bsp_debian_deps = [
+    # Edgecore Wedge100bf BSP dependencies.
+    "libusb-1.0-0",
+    "libcurl4-gnutls-dev",
+]
+
 pkg_deb(
     name = "stratum_bf_deb",
     architecture = "amd64",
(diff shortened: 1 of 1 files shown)


stratum/hal/bin/barefoot/BUILD
```diff
@@ -166,6 +166,8 @@ pkg_tar(
     ],
 )
 
+sde_bsp_debian_deps = ["libusb-1.0-0"]
```
pudelkoM, line 169: This should really go into `bfsde.build`, but sharing variables requires using `.bzl` files, which might be overkill for a single instance.

Alternatively, we could package the SDE into a deb with the correct deps, and depend on that later. @bocon13 floated that idea before.

stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -38,10 +38,12 @@ if [ -n "$CHASSIS_CONFIG" ]; then
 fi
 
 # Start Stratum.
+# --network host: the BSP needs access to the usb0 interface.
```
pudelkoM, line 41: @bocon13 @Yi-Tseng

I found no other way yet. For reference: the BSP will do a http request against the BMC in `bf_pltfm_bd_type_init`.

If this is our solution, we have to update the run README.

PR #769, opened 2021-07-01
Use Docker host networking for Stratum-bf[rt] containers on Wedge

pudelkoM:
The BSP on Wedge100bf switches talks to the BMC over the `usb0` interface during startup. By default, docker isolates the interface from the container, causing the startup procedure to fail with `2021-07-01 22:06:27.571944 BF_PLTFM ERROR - Error Getting the Board info from BMC. Hence exiting ****`. This change binds the Stratum container to the host network, when run on applicable switches, fixing the error.

### DIFF
--- a/stratum/hal/bin/barefoot/docker/start-stratum-container.sh
+++ b/stratum/hal/bin/barefoot/docker/start-stratum-container.sh
@@ -17,7 +17,12 @@ elif [[ -z "$PLATFORM" ]]; then
 fi
 
 # Set Docker network options.
-if [[ "$PLATFORM" == 'barefoot-tofino-model' ]]; then
+# On tofino-model and certain switches we run Stratum directly on the host
+# network. The BSP on Wedge devices needs access to the usb0 interface.
+if [[ "$PLATFORM" == 'barefoot-tofino-model' ]] || \
+   [[ "$PLATFORM" == "x86-64-accton-wedge100bf-32x-r0" ]] || \
+   [[ "$PLATFORM" == "x86-64-accton-wedge100bf-32qs-r0" ]] || \
+   [[ "$PLATFORM" == "x86-64-accton-wedge100bf-65x-r0" ]]; then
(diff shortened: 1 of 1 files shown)


stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -17,7 +17,12 @@ elif [[ -z "$PLATFORM" ]]; then
 fi
 
 # Set Docker network options.
-if [[ "$PLATFORM" == 'barefoot-tofino-model' ]]; then
+# On tofino-model or switches with BSP support, we bind to the host network
+# directly. The BSP on Wedge devices needs access to the usb0 interface.
+if [[ "$PLATFORM" == 'barefoot-tofino-model' ]] || \
+   [[ "$PLATFORM" == "x86-64-accton-wedge100bf-32x-r0" ]] || \
+   [[ "$PLATFORM" == "x86-64-accton-wedge100bf-32qs-r0" ]] || \
+   [[ "$PLATFORM" == "x86-64-accton-wedge100bf-65x-r0" ]]; then
```
pudelkoM, line 25: @bocon13 @Yi-Tseng

I found no other way yet. For reference: the BSP will do a http request against the BMC in `bf_pltfm_bd_type_init`.

If this is our solution, we have to update the run README.

bocon13: I don't see a big problem with this.

If we don't want to use the host network, we could consider moving the usb0 interface into the container's netns. (Downside is that the host wouldn't be able to access the BMC.)

Or, we could create a bridge... But that's definitely going to be more complicated.

pudelkoM: Sounds good.

I looked at the `README.run.md`, but all mentions of `--network` apply to tofino model.

Yi-Tseng (approved):
Right, EdgeCore uses OpenBMC restful API to get some information

PR #774, opened 2021-07-08
Merge `BfrtActionProfileManager` into `BfrtTableManager`

pudelkoM:
The separation of `BfrtActionProfileManager` and `BfrtTableManager` was largely a logical one, not because of state or complexity concerns.
P4Runtime one-shot action profile groups blur the lines between those two and would be hard to implement within the current structure. This change merges the two managers into one.


stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -934,6 +934,309 @@ ::util::Status BfrtTableManager::WriteMeterEntry(
...
+    const ::p4::v1::Update::Type type, const ::p4::v1::ExternEntry& entry) {
+  absl::WriterMutexLock l(&lock_);
+  ASSIGN_OR_RETURN(uint32 bfrt_table_id,
+                   bf_sde_interface_->GetBfRtId(entry.extern_id()));
+  switch (entry.extern_type_id()) {
+    case kTnaExternActionProfileId: {
+      ::p4::v1::ActionProfileMember act_prof_member;
+      CHECK_RETURN_IF_FALSE(entry.entry().UnpackTo(&act_prof_member))
+          << "Entry " << entry.ShortDebugString()
+          << " is not an action profile member.";
```
pudelkoM, line 948: If this goes into bfrt_node, we could delete a lot of code.

stratum/hal/lib/barefoot/bfrt_node.cc
```diff
@@ -420,13 +408,25 @@ ::util::Status BfrtNode::WriteExternEntry(
...
-    case kTnaExternActionSelectorId:
-      return bfrt_action_profile_manager_->WriteActionProfileEntry(session,
-                                                                   type, entry);
+    case kTnaExternActionProfileId: {
+      ::p4::v1::ActionProfileGroup act_prof_group;
+      CHECK_RETURN_IF_FALSE(entry.entry().UnpackTo(&act_prof_group))
+          << "Entry " << entry.ShortDebugString()
+          << " is not an action profile group.";
+      return bfrt_table_manager_->WriteActionProfileGroup(session, type,
+                                                          act_prof_group);
```
pudelkoM, line 417: @Yi-Tseng any reason no to unpack the extern here and then just call the regular `WriteActionProfileGroup` API?

Yi-Tseng: I forgot why I unpack it in the manager, but I agree that we can just unpack it here.

Yi-Tseng: By the way, the original logic is to unpack action profile member if it is an ActionProfile and unpack action profile group if it is an ActionSelector

pudelkoM: good catch! I swapped it back.

PR #781, opened 2021-07-09
Replace P4Runtime controller management with PINS' code

### LINKED ISSUE #287: Support P4Runtime "Role"
In P4Runtime spec:

```Partitioning of the control plane: Multiple controllers may have orthogonal, non-overlapping, “roles” (or “realms”) and should be able to push forwarding entities simultaneously. The control plane can be partitioned into multiple roles and each role will have a set of controllers, one of which is the master and the rest are slaves. Role definition, i.e. how P4 entities get assigned to each role, is out-of-scope of this document.```

### LINKED ISSUE #380: Update client arbitration to match P4RT 1.3.0
Our current implementation doesn't quite match the spec:
https://p4.org/p4runtime/spec/v1.2.0/P4Runtime-Spec.html#sec-master-slave-arbitration-and-controller-replication

### PR DESCRIPTION
**Pull request by pudelkoM**
This PR replaces the Stratum-internal P4Runtime server connection management code, part of `P4Service`, with a library extracted from the PINS project. The library implements the current version of the spec with the current mastership arbitration algorithm; **Stratum will not longer pick a new master from the backup controller pool**.

Re-using existing externally maintained code should give us the benefit of future updates for cheap. This is why we strive for minimal modifications to the imported code. In particular, we do not include Stratum deps so that other can use it as well.

Not addressed here is the persistent election ID issue.

fixes #287
fixes #380

### CODE REVIEW

**Comment by pudelkoM** [comment]
We're starting to discover [issues](https://jira.opennetworking.org/browse/AETHER-1071) with Stratum's non-spec compliant P4RT mastership handling. It seems to be time to address this after the Aether 2.0 release, and maybe after the Stratum 22.03 release as well?

File: stratum/hal/lib/common/p4_service_test.cc [hunk 1]
```diff
@@ -130,6 +131,36 @@ class P4ServiceTest : public ::testing::TestWithParam<OperationMode> {
     p4_service_->node_id_to_controllers_[node_id].insert(controller);
   }
 
+  grpc::Status AddFakeMasterController2(uint64 node_id, uint64 connection_id,
```
pudelkoM, line 134: cleanup

stratum/hal/lib/common/p4_service_test.cc
```diff
@@ -998,34 +1045,40 @@ TEST_P(P4ServiceTest, StreamChannelSuccess) {
...
-  ASSERT_EQ(absl::Uint128Low64(kElectionId2),
-            resp.arbitration().election_id().low());
-  ASSERT_EQ(::google::rpc::ALREADY_EXISTS, resp.arbitration().status().code());
+  // ASSERT_TRUE(stream1->Read(&resp));
+  // ASSERT_EQ(absl::Uint128High64(kElectionId2),
+  //           resp.arbitration().election_id().high());
+  // ASSERT_EQ(absl::Uint128Low64(kElectionId2),
+  //           resp.arbitration().election_id().low());
+  // ASSERT_EQ(::google::rpc::ALREADY_EXISTS,
+  // resp.arbitration().status().code());
```
pudelkoM, line 1073: figure out what to do here

stratum/hal/lib/common/p4_service_test.cc
```diff
@@ -1082,21 +1145,28 @@ TEST_P(P4ServiceTest, StreamChannelSuccess) {
...
+
+  // FIXME: SdnControllerManager does not send updates for a mastership change
+  // with a non-increasing election ID. ASSERT_TRUE(stream1->Read(&resp));
+  // LOG(ERROR) << "13: " << resp.ShortDebugString();
+  // ASSERT_EQ(absl::Uint128High64(kElectionId2),
+  //           resp.arbitration().election_id().high());
+  // ASSERT_EQ(absl::Uint128Low64(kElectionId2),
+  //           resp.arbitration().election_id().low());
+  // ASSERT_EQ(::google::rpc::ALREADY_EXISTS,
+  // resp.arbitration().status().code());
```
pudelkoM, line 1161: same

stratum/hal/lib/common/p4_service.h
```diff
@@ -184,23 +185,37 @@ class P4Service final : public ::p4::v1::P4Runtime::Service {
...
   // controller is disconnected).
   void RemoveController(uint64 node_id, uint64 connection_id)
       LOCKS_EXCLUDED(controller_lock_);
+  void RemoveController(uint64 node_id, p4runtime::SdnConnection* connection)
+      LOCKS_EXCLUDED(controller_lock_);
 
   // Returns true if given (election_id, uri) for a Write request belongs to the
   // master controller stream for a node given by its node ID.
   bool IsWritePermitted(uint64 node_id, absl::uint128 election_id,
                         const std::string& uri) const
```
pudelkoM, line 203: check if still needed

stratum/hal/lib/common/p4_service.h
```diff
@@ -184,23 +185,37 @@ class P4Service final : public ::p4::v1::P4Runtime::Service {
...
       LOCKS_EXCLUDED(controller_lock_);
+  ::util::Status AddOrModifyController(
+      uint64 node_id, const p4::v1::MasterArbitrationUpdate& update,
+      p4runtime::SdnConnection* controller) LOCKS_EXCLUDED(controller_lock_);
 
   // Removes an existing controller from the controllers_ set given its stream.
   // To be called after stream from an existing controller is broken (e.g.
   // controller is disconnected).
   void RemoveController(uint64 node_id, uint64 connection_id)
       LOCKS_EXCLUDED(controller_lock_);
```
pudelkoM, line 196: check if still needed

stratum/hal/lib/common/p4_service.h
```diff
@@ -184,23 +185,37 @@ class P4Service final : public ::p4::v1::P4Runtime::Service {
                                        const std::string& uri,
                                        ServerStreamChannelReaderWriter* stream)
```
pudelkoM, line 186: check if still needed

stratum/hal/lib/common/p4_service.h
```diff
@@ -184,23 +185,37 @@ class P4Service final : public ::p4::v1::P4Runtime::Service {
...
+  bool IsWritePermitted(uint64 node_id, const p4::v1::WriteRequest& req) const
+      LOCKS_EXCLUDED(controller_lock_);
+  bool IsWritePermitted(uint64 node_id,
+                        const p4::v1::SetForwardingPipelineConfigRequest& req)
+      const LOCKS_EXCLUDED(controller_lock_);
 
   // Returns true if the given connection_id belongs to the master controller
   // stream for a node given by its node ID.
   bool IsMasterController(uint64 node_id, uint64 connection_id) const
       LOCKS_EXCLUDED(controller_lock_);
```
pudelkoM, line 214: check if still needed

stratum/hal/lib/common/p4_service.h
```diff
@@ -179,27 +134,29 @@ class P4Service final : public ::p4::v1::P4Runtime::Service {
...
   // packet I/O.
-  ::util::Status AddOrModifyController(uint64 node_id, uint64 connection_id,
-                                       absl::uint128 election_id,
-                                       const std::string& uri,
-                                       ServerStreamChannelReaderWriter* stream)
+  ::util::Status AddOrModifyController(
+      uint64 node_id, const p4::v1::MasterArbitrationUpdate& update,
+      p4runtime::SdnConnection* controller) LOCKS_EXCLUDED(controller_lock_);
+
+  // Removes an existing controller from the controllers manager given its
```
Yi-Tseng, line 141:
```suggestion
  // Removes an existing controller from the controller manager given its
```

stratum/hal/lib/common/p4_service.cc
```diff
@@ -713,168 +712,59 @@ ::util::Status P4Service::AddOrModifyController(
     // Store Channel and tid for Teardown().
     stream_response_reader_tids_.push_back(tid);
     stream_response_channels_[node_id] = channel;
-    node_id_to_controllers_[node_id] = {};
-    it = node_id_to_controllers_.find(node_id);
+    node_id_to_controller_manager_.emplace(node_id, node_id);
```
Yi-Tseng, line 715: Should we put a `SdnControllerManager` instead of a `uint64` to the map as the value?
Since the type of `node_Id_to_controller_manager_` is `std::map<uint64, p4runtime::SdnControllerManager>`

pudelkoM: It actually does that, but it looks confusing. Here's what's going on:

- `SdnControllerManager` cannot be copied, only moved
- hence we do `emplace` instead of `insert`
- but we still can't construct the manager object here (e.g. `emplace(node_id, SdnControllerManager())`) as that would be a copy
- so we use [perfect forwarding](https://en.cppreference.com/w/cpp/utility/forward) semantics to construct the manager directly _inside_ the map only
- as a result, here, we're only providing the constructor args of the manager to be forwarded, hence the second `node_id`.

Hope this makes sense.

Yi-Tseng: Now I get it. Totally make sense now. It is ok to leave this code here.
Maybe we can put some comments here so any other people won't confuse like me in the future?

.circleci/config.yml
```diff
@@ -411,9 +411,9 @@ workflows:
           filters:
             branches:
               only: main
-      - publish-docker-mininet:
-          requires:
-            - publish-docker-build
```
pudelkoM, line 416: revert before merging

bocon13:
fixes #287
fixes #380

.circleci/config.yml
pierventre: we are good to go from my perspective

stratum/hal/lib/common/p4_service.cc
```diff
@@ -83,8 +83,7 @@ ::util::Status P4Service::Setup(bool warmboot) {
 ::util::Status P4Service::Teardown() {
   {
     absl::WriterMutexLock l(&controller_lock_);
-    node_id_to_controllers_.clear();
-    connection_ids_.clear();
+    node_id_to_controller_manager_.clear();
```
pierventre, line 86: why are you not purging `num_controller_connections_` ?

stratum/hal/lib/common/p4_service.cc
```diff
@@ -547,21 +542,23 @@ ::grpc::Status P4Service::StreamChannel(
   // 3- At any point of time, only the master stream is capable of sending
   //    and receiving packets.
 
-  // First thing to do is to find a new ID for this connection.
-  auto ret = FindNewConnectionId();
+  // First thing to do is to check the connection limit and increment it.
```
pierventre, line 545: it is not obvious to me what connection limit means. Maybe add an additional comment saying that we don't handle more than N connections at time

stratum/lib/p4runtime/sdn_controller_manager.cc
```diff
@@ -0,0 +1,445 @@
...
+#include "absl/status/status.h"
+#include "absl/strings/str_cat.h"
+#include "absl/strings/str_format.h"
+#include "absl/strings/str_join.h"
+#include "absl/types/optional.h"
+#include "glog/logging.h"
+#include "p4/v1/p4runtime.pb.h"
+#include "stratum/glue/gtl/map_util.h"
+
+// TODO(max): version 1.3.0 of P4Runtime does not fully support roles yet.
```
pierventre, line 17: is this a current TODO or for a future pr ?

stratum/lib/p4runtime/sdn_controller_manager.cc
```diff
@@ -0,0 +1,445 @@
...
+// Flip and remove once landed.
+#define P4RT_HAS_ROLES 0
+
+namespace stratum {
+namespace p4runtime {
+namespace {
+
+std::string PrettyPrintRoleName(const absl::optional<std::string>& name) {
```
pierventre, line 25: should we consider to move these two Pretty* function to an utils file

pierventre (dismissed):
minor comments

stratum/lib/p4runtime/sdn_controller_manager.cc
pudelkoM: future, needs a new p4rt release

stratum/hal/lib/common/p4_service.cc
pudelkoM: good point, should be reset as well.

pierventre (dismissed):

stratum/lib/p4runtime/sdn_controller_manager.cc
```diff
@@ -0,0 +1,445 @@
...
+        update.election_id().high(), update.election_id().low());
+  }
+
+  const bool new_connection = !controller->IsInitialized();
+
+  if (new_connection) {
+    // First arbitration message sent by this controller.
+
+    // Verify the request's device ID is being sent to the correct device.
+    if (update.device_id() != device_id_) {
```
Yi-Tseng, line 135: Please correct me if I am wrong
After the client creates a new gRPC stream, it will send a new arbitration update message to the server, which includes a `device_id` in the message and we store this data in `node_id` in `p4_service`([line 573](https://github.com/stratum/stratum/blob/82b2ff66659fc2324f213e3b3b10da081b4a704b/stratum/hal/lib/common/p4_service.cc#L573)).
Everytime when the client sends a `StreamMessageRequest` we will check if the `device_id` changed(`p4_service` [line 574](https://github.com/stratum/stratum/blob/82b2ff66659fc2324f213e3b3b10da081b4a704b/stratum/hal/lib/common/p4_service.cc#L574))
Does this mean we don't need to check the id again in this function? Or we can simplify the check in the `p4_service`?

Yi-Tseng: Also, I was thinking if this condition will always be false since we always get the `sdn_controller_manager` based on the `device_id`/`node_id` ([line 718](https://github.com/stratum/stratum/blob/82b2ff66659fc2324f213e3b3b10da081b4a704b/stratum/hal/lib/common/p4_service.cc#L684))

Yi-Tseng: comments also apply to line 161

stratum/lib/p4runtime/sdn_controller_manager.cc
```diff
@@ -0,0 +1,445 @@
...
+
+bool SdnControllerManager::PrimaryConnectionExists(
+    const absl::optional<std::string>& role_name) {
+  absl::optional<absl::uint128> election_id_past_for_role =
+      election_id_past_by_role_[role_name];
+
+  for (const auto& connection : connections_) {
+    if (connection->GetRoleName() == role_name &&
+        connection->GetElectionId() == election_id_past_for_role) {
+      return election_id_past_for_role.has_value();
```
Yi-Tseng, line 350: Is there a case that the election id for a role is an empty value? I believe that we always we always add or modify the SdnConnection with a valid election id (https://github.com/stratum/stratum/blob/82b2ff66659fc2324f213e3b3b10da081b4a704b/stratum/hal/lib/common/p4_service.cc#L583-L586)

stratum/lib/p4runtime/sdn_controller_manager.cc
pudelkoM: There is a duplicate check happening, true. I think the sdn_controller should be as simple and safe to use as possible, hence should be in charge of doing the check. Then we can probably remove it from p4 service.

stratum/hal/lib/common/p4_service.cc
```diff
@@ -649,41 +656,34 @@ ::grpc::Status P4Service::StreamChannel(
...
-  response->set_p4runtime_api_version(STRINGIFY(P4RUNTIME_VER));
+::grpc::Status P4Service::Capabilities(::grpc::ServerContext* context,
+                                       const ::p4::v1::CapabilitiesRequest* req,
+                                       ::p4::v1::CapabilitiesResponse* resp) {
+  resp->set_p4runtime_api_version(STRINGIFY(P4RUNTIME_VER));
   return ::grpc::Status::OK;
 }
 
-::util::StatusOr<uint64> P4Service::FindNewConnectionId() {
+::util::Status P4Service::CheckAndCountConnectionLimit() {
```
pudelkoM, line 666:
```suggestion
::util::Status P4Service::CheckAndIncrementConnectionCount() {
```

?

stratum/lib/p4runtime/sdn_controller_manager.cc
pudelkoM: Not really needed, I think. The public interface of `SdnConnection` is `GetRoleName()` which already calls this function. It's really a private helper function, hence its placement in an anonymous namespace.

stratum/lib/p4runtime/sdn_controller_manager.cc
pudelkoM: I think the spec allows backups without any election ID set:

The sender need not specify an election_id. If t

> he election_id is not specified, the sender's election_id is considered lower than any election_id, and the sender will thus never become primary.

Yi-Tseng: got it, maybe we can remove the check from the `p4_service`?

pudelkoM: I think I understood the logic here.


https://github.com/stratum/stratum/blob/4db73796d082973e63d3309b373343927ef19d4c/stratum/lib/p4runtime/sdn_controller_manager.h#L152-L154

Constructing the map with `absl::optional` as values allows liberal use of the index operator (`map[]`) without having to worry about inserting a key with a default value by accident. Which is what happens in a statement like this: `a = map[key];`. Looks like a read, but is actually a write into the map. With `absl::optional` values the key "presence check" becomes `map[key].has_value()` instead of `map.find()`.

Nonetheless, I simplified the check a bit by using the same patter as in other places.

stratum/lib/p4runtime/sdn_controller_manager.cc
pudelkoM: done

Yi-Tseng (commented):
Overall looks good to me, I guess you still need to clean up some changes? (e.g. config.yml)

stratum/lib/p4runtime/sdn_controller_manager.h
```diff
@@ -0,0 +1,170 @@
...
+      grpc_stream_;  // not owned.
+};
+
+class SdnControllerManager {
+ public:
+  explicit SdnControllerManager(uint64_t device_id) : device_id_(device_id) {}
+
+  grpc::Status HandleArbitrationUpdate(
+      const p4::v1::MasterArbitrationUpdate& update, SdnConnection* controller)
+      ABSL_LOCKS_EXCLUDED(lock_);
```
Yi-Tseng, line 78: I just realized that we are returning the gRPC status instead of Stratum status here.
In this case, we convert a gRPC status to Stratum status from the caller and convert to another gRPC status and send it back to the controller.
Should we change these functions to use Stratum status?

Yi-Tseng: `SDN controller manager` --`grpc::Status`--> `P4Service::AddOrModifyController` --`util::Status`--> `P4Service::StreamChannel` --`grpc::Status`--> gRPC

pudelkoM: That is somewhat unfortunate, yes. One of the goals with using the external PINS code was getting code that is maintained by them and we can simply update. Modifying it too much or pulling in Stratum deps would make these updates harder.

Yi-Tseng (approved):

stratum/hal/lib/common/p4_service.cc
```diff
@@ -649,41 +652,34 @@ ::grpc::Status P4Service::StreamChannel(
...
-::util::StatusOr<uint64> P4Service::FindNewConnectionId() {
+::util::Status P4Service::CheckAndIncrementConnectionCount() {
   absl::WriterMutexLock l(&controller_lock_);
-  if (static_cast<int>(connection_ids_.size()) >=
-      FLAGS_max_num_controller_connections) {
+
+  if (num_controller_connections_ >= FLAGS_max_num_controller_connections) {
     return MAKE_ERROR(ERR_NO_RESOURCE)
            << "Can have max " << FLAGS_max_num_controller_connections
            << " active/inactive streams for all the node.";
```
pudelkoM, line 668:
```suggestion
           << " active/inactive streams for all the nodes.";
```

stratum/hal/lib/common/p4_service.h
```diff
@@ -179,27 +135,29 @@ class P4Service final : public ::p4::v1::P4Runtime::Service {
   // takes the lock. After successful completion of this function, the first
   // element in controllers_ set will have the master controller stream for
   // packet I/O.
```
pudelkoM, line 137: Update comment

stratum/hal/lib/common/p4_service.h
```diff
@@ -232,12 +190,20 @@ class P4Service final : public ::p4::v1::P4Runtime::Service {
...
+  // is used for PacketIO, and is the only connection allowed to write updates.
+  //
+  // It is possible for connections to be made for specific roles. In which case
+  // one primary connection is allowed for each distinct role.
+  std::map<uint64, p4runtime::SdnControllerManager>
+      node_id_to_controller_manager_ ABSL_GUARDED_BY(controller_lock_);
+
+  // Holds the number of currently open StreamChannels across all nodes. This is
+  // tracked for resource limiting. Note that this count can be different from
+  // the number of connected controllers reported by all controller managers, as
```
pudelkoM, line 204:
```suggestion
  // the sum of connected controllers reported by all controller managers, as
```

stratum/lib/p4runtime/BUILD
```diff
@@ -35,3 +36,32 @@ stratum_cc_library(
...
     ],
 )
+
+stratum_cc_library(
+    name = "sdn_controller_manager",
+    srcs = ["sdn_controller_manager.cc"],
+    hdrs = ["sdn_controller_manager.h"],
+    deps = [
+        "//stratum/glue:logging",
+        "@com_github_google_glog//:glog",
```
pudelkoM, line 46: possibly remove

stratum/lib/p4runtime/BUILD
```diff
@@ -35,3 +36,32 @@ stratum_cc_library(
...
+        "@com_github_grpc_grpc//:grpc++",
+        "@com_github_p4lang_p4runtime//:p4runtime_cc_grpc",
+        "@com_github_p4lang_p4runtime//:p4runtime_cc_proto",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/numeric:int128",
+        "@com_google_absl//absl/status",
```
pudelkoM, line 52: check if used

stratum/hal/lib/common/p4_service.h
```diff
@@ -232,12 +190,20 @@ class P4Service final : public ::p4::v1::P4Runtime::Service {
...
-  std::map<uint64, std::set<Controller, ControllerComp>> node_id_to_controllers_
-      GUARDED_BY(controller_lock_);
+  // P4Runtime can accept multiple connections to a single switch for
+  // redundancy. When there is >1 connection the switch chooses a primary which
+  // is used for PacketIO, and is the only connection allowed to write updates.
+  //
+  // It is possible for connections to be made for specific roles. In which case
+  // one primary connection is allowed for each distinct role.
+  std::map<uint64, p4runtime::SdnControllerManager>
+      node_id_to_controller_manager_ ABSL_GUARDED_BY(controller_lock_);
```
pudelkoM, line 200: Check if this can be a absl hash map. We do not depend on the ordering anymore.

stratum/lib/p4runtime/sdn_controller_manager.cc
```diff
@@ -0,0 +1,431 @@
...
+  return absl::StrCat("(role_name: ", PrettyPrintRoleName(role_name_),
+                      ", election_id: ", PrettyPrintElectionId(election_id_),
+                      ", uri: ", grpc_context_->peer(), ")");
+}
+
+void SdnConnection::SendStreamMessageResponse(
+    const p4::v1::StreamMessageResponse& response) {
+  VLOG(2) << "Sending response: " << response.ShortDebugString();
+  if (!grpc_stream_->Write(response)) {
+    LOG(ERROR) << "Could not send arbitration update response to gRPC context '"
```
pudelkoM, line 99: should probably have a `.log_by_default(false)` modifier.

stratum/lib/p4runtime/BUILD
pudelkoM: is used

stratum/lib/p4runtime/BUILD
pudelkoM: `"//stratum/glue:logging"` should be removed

stratum/hal/lib/common/p4_service.h
pudelkoM: I encountered problems with the absl flat_hash_map and in-place object creation, it still tried to move the controller. For now this will remain a `std::unordered_map`.

Yi-Tseng (dismissed):

stratum/hal/lib/common/p4_service.h
```diff
@@ -219,7 +177,7 @@ class P4Service final : public ::p4::v1::P4Runtime::Service {
                                     const ::p4::v1::StreamMessageResponse& resp)
       LOCKS_EXCLUDED(controller_lock_);
 
-  // Mutex lock used to protect node_id_to_controllers_ which is updated
+  // Mutex lock used to protect node_id_to_controller_manager_ which is updated
   // every time mastership for any of the controllers connected to each node is
   // modified, or when a controller is disconnected.
```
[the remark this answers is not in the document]
pudelkoM: I agree on most of the comment, but the node->manager mapping is still there?
```c++
 std::unordered_map<uint64, p4runtime::SdnControllerManager> node_id_to_controller_manager_;
```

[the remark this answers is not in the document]
pudelkoM: I pushed a new comment, let me know what you think.

PR #784, opened 2021-07-14
Reject range "don't care" match fields in Stratum-bfrt

pudelkoM:
According to the [P4Runtime spec](https://opennetworking.org/wp-content/uploads/2020/10/P4Runtime-Specification-120-wd.html#sec-match-format) "don't care" range match field must be omitted from a request. This change enables a check to ensure that a clients does not send those.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager.cc
@@ -176,7 +176,9 @@ std::unique_ptr<BfrtTableManager> BfrtTableManager::CreateInstance(
                                 ::p4::config::v1::MatchField::EXACT)
               << "Found match field of type EXACT does not fit match field "
               << expected_match_field.ShortDebugString() << ".";
-          CHECK_RETURN_IF_FALSE(!IsDontCareMatch(mk.exact()));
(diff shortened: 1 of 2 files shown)


stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -176,7 +176,9 @@ ::util::Status BfrtTableManager::BuildTableKey(
                                 ::p4::config::v1::MatchField::EXACT)
               << "Found match field of type EXACT does not fit match field "
               << expected_match_field.ShortDebugString() << ".";
-          CHECK_RETURN_IF_FALSE(!IsDontCareMatch(mk.exact()));
```
pierventre, line 179: I believe there is no "don't care" for exact match

pierventre (approved):
the implementation for exact match returns always false but it is quite misleading. Up to you.

stratum/hal/lib/barefoot/bfrt_table_manager.cc
pudelkoM: correct, that function will always return false.

pierventre: maybe add a comment :)

PR #786, opened 2021-07-16
Add support for BF SDE 9.6.0

bocon13:
SDE 9.6.0 has changed the build process. `p4studio_build.py` is now `p4studio`.

`stratum_bf` and `stratum_bfrt` start up cleanly against `tofino-model`

Notable changes
- `libbf_switchd_lib.a` no longer exists and required symbols appear to be in `libdriver.so`
- The `libpiall` dependency transitioned from a static library to shared library
- `libJudy` transitioned from a static library to shared library

TODO
- [ ] conditionally depend on `libbf_switchd_lib.a` based on SDE version
- [ ] verify the dependencies are being included

This PR doesn't address the build script changes required for 9.6.0; you need to build `bf-sde-9.6.0-install.tgz` by hand with a few patches to the CMake files. Follow #795 for details on that.


stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -5,7 +5,8 @@ set -e
 
 LOG_DIR=${LOG_DIR:-/var/log}
 SDE_VERSION=${SDE_VERSION:-9.3.2}
-DOCKER_IMAGE=${DOCKER_IMAGE:-stratumproject/stratum-bf}
+#FIXME update "stratum-bf" to "stratum-bfrt" globally
```
bocon13, line 8: make this a separate PR

pudelkoM: Done: #789

Yi-Tseng:
there is  a `libbfrt.so` that includes bfrt symbols we are usually using, I guess we also need to include it?

stratum/hal/bin/barefoot/docker/start-stratum-container.sh
smittal-bf: Thanks Max and Brian for this quick support!

bocon13: thanks!

bocon13:
We are skipping SDE 9.6.0

PR #788, opened 2021-07-16
Pass .circleci/bazelrc as flag to Bazel

bocon13:
Avoid overwriting .bazelrc to keep the git tree clean

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,17 +35,11 @@ analyze_bazel_profile: &analyze_bazel_profile
   run:
     name: Analyze Bazel profile
     command: |
-      bazel analyze-profile /tmp/bazel.profile.json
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -35,18 +35,12 @@ analyze_bazel_profile: &analyze_bazel_profile
   run:
     name: Analyze Bazel profile
     command: |
-      bazel analyze-profile /tmp/bazel.profile.json
+      bazel --bazelrc .circleci/bazelrc analyze-profile /tmp/bazel.profile.json
```
bocon13, line 38: Consider extracting `--bazelrc .circleci/bazelrc` into a variable; e.g. `$BAZEL_OPTS`

pudelkoM (approved):
Looks fine to me. Considering that forgetting `--bazelrc` is just as easy as forgetting the `*set_bazelrc` step, it's not a loss or gain.

PR #791, opened 2021-07-20
Push tools image to Aether harbor registry

pudelkoM:
With the docker.io rate limits and the depreciation of the bf-pipeline-builder image, I decided to upload the `stratum-tools` image to the Aether registry as well. It can be found here: https://registry.aetherproject.org/harbor/projects/4/repositories/stratum-tools. ONF projects should use that image going forward.

Pros of pushing the image from here (circleci):
- Full control over build flags and tags
- Easy to change, update and maintain in one atomic commit

Cons:
- Increased risk of credential leak (will be addressed in the future with project specific credentials)
- Two names for the same image `tost/stratum-tools` and `stratumproject/stratum-tools`

Any other concerns?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,12 @@ docker_login: &docker_login
   run:
     name: Docker login
     command: echo $DOCKER_PWD | docker login -u $DOCKER_LOGIN --password-stdin
+
+harbor_login: &harbor_login
+  run:
+    name: Harbor login
+    command: echo $AETHER_HARBOR_PWD | docker login -u $AETHER_HARBOR_LOGIN --password-stdin registry.aetherproject.org
+
 docker_build: &docker_build
   run:
     name: Build Docker image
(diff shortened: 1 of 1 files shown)


stolsma:
I hope this doesn't mean that in the future access to Stratum container images will become for ONF members only as is the case for the Aether registry. That would be a real bummer for independent developers....😔

pudelkoM:
> @bocon13 @pudelkoM I hope this doesn't mean that in the future access to Stratum container images will become for ONF members only as is the case for the Aether registry. That would be a real bummer for independent developers....😔

Thanks for voicing your concerns. Here are some points to hopefully dispel them:
- We pushing the image to an **additional** registry 
- This job would otherwise run on our internal jenkins CI (See https://github.com/stratum/stratum-ci)
- The source code to build the images is still available

bocon13 (approved):
lgtm

bocon13:
@stolsma, thanks for reaching out! Nothing is changing on the public Docker Hub side. It's our intention to keep Stratum source code fully open and to continue to publish public images for new releases as we've done for the past ones. :)

PR #799, opened 2021-07-28
Update PI and BMv2 versions

bocon13:
Add veth_setup.sh and veth_teardown.sh to build Docker image
Update BMv2 reference chassis config will all 9 ports


bocon13:
@ccascone @pudelkoM @Yi-Tseng any objections to bumping the versions on PI and BMv2?

Also, I noticed that the PI version imported by Bazel and the version that we install into the build container are a little out of sync.

pudelkoM:
> @ccascone @pudelkoM @Yi-Tseng any objections to bumping the versions on PI and BMv2?
> 
> Also, I noticed that the PI version imported by Bazel and the version that we install into the build container are a little out of sync.

No concern from my side. Last time we updated PI it broke a few assumptions clients (e.g. ONOS) made about bmv2.

ccascone:
LGTM

bocon13:
Looks like the diff is pretty small. There is a fix in both PI and BMv2 for meters, so that may be the only place that we could see anything different.

pudelkoM (approved):

PR #801, opened 2021-07-30
WIP support for one-shot action profile groups

pudelkoM:


stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -678,6 +704,195 @@ ::util::Status BfrtTableManager::ReadTableEntry(
...
+      }
+      std::vector<uint32> all_member_ids;
+      for (const auto& member_group : member_ids) {
+        all_member_ids.insert(all_member_ids.end(), member_group.begin(),
+                              member_group.end());
+      }
+      std::sort(all_member_ids.begin(), all_member_ids.end());
+      if (!all_member_ids.empty()) {
+        next_free_member_id = *(all_member_ids.rbegin()) + 1;
+      }
```
pudelkoM, line 850: This search will eventually overflow back to 0 and potentially fail with a duplicate.

stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -678,6 +704,195 @@ ::util::Status BfrtTableManager::ReadTableEntry(
...
+                          table_entry.table_id(), true);
+  if (!uses_one_shot_action) {
+    RETURN_ERROR(ERR_INVALID_PARAM)
+        << "One-shot and traditional action profile programming styles "
+           "cannot be mixed on the same action profile "
+        << action_profile.preamble().name() << " or table "
+        << table_entry.table_id() << ".";
+  }
+
+  if (type == ::p4::v1::Update::DELETE) {
```
pudelkoM, line 793: I think modify is a superset of delete and insert, the logic should be here already.

stratum/hal/lib/barefoot/bfrt_table_manager.cc
```diff
@@ -678,6 +704,195 @@ ::util::Status BfrtTableManager::ReadTableEntry(
...
+        all_member_ids.insert(all_member_ids.end(), member_group.begin(),
+                              member_group.end());
+      }
+      std::sort(all_member_ids.begin(), all_member_ids.end());
+      if (!all_member_ids.empty()) {
+        next_free_member_id = *(all_member_ids.rbegin()) + 1;
+      }
+    }
+    // 2. Create the members.
+    const auto& set = table_entry.action().action_profile_action_set();
```
Yi-Tseng, line 853: rename to something like `action_set` for better readability?

Yi-Tseng:
> I think `MODIFY` is a superset of delete and insert, the logic should be here already. 

I think this is true, we can actually create a new group, modify the table entry to use the new group and remove the old group.

stratum/hal/lib/barefoot/bfrt_table_manager.cc
pudelkoM: The variable was only used in a single place, so I just removed it.

Yi-Tseng (commented):
Just wondering if we can convert the table entry to the one-shot one in the `BuildP4TableEntry` function so we can reuse the logic when reading single, wildcard, and default entry?

For example:
```
BuildP4TableEntry {
  // Build a P4 table entry....
  if table_is_one_shot_type {
    convert_to_one_shot(result)
  }
  return result
}
```

pudelkoM:
> Just wondering if we can convert the table entry to the one-shot one in the `BuildP4TableEntry` function so we can reuse the logic when reading single, wildcard, and default entry?
> 
> For example:
> 
> ```
> BuildP4TableEntry {
>   // Build a P4 table entry....
>   if table_is_one_shot_type {
>     convert_to_one_shot(result)
>   }
>   return result
> }
> ```

Main hurdle I see right now is that one-shot profiles are built from multiple table data objects.

PR #804, opened 2021-08-02
[stratum-bcm] Support the P4 DirectMeterEntryRead operation

miguelbf-alb:
## Description

This PR implements the `DirectMeterEntryRead` operation, required for the controller to read the direct meter resources. Since the meter_config is already on the ACL table entries implementing this message is as simple as copying the values from there to the DirectMeter mutablecConfig. Comparing stratum-bcm to stratum-bmv2 we found that the table entry action and meter_config are not part of the response. So, we are explicitly removing those fields from the response.
I am a bit unsure if this approach is 100% correct so your feedback is appreciated. Btw, DirectCounterRead operations are currently broken in stratum-bcm and should follow a similar approach to this one (I'll look into that later).

## Todo:

- [ ] Unit Tests
- [ ] Describe the runtime validation approach

### DIFF
--- a/stratum/hal/lib/bcm/bcm_table_manager.cc
+++ b/stratum/hal/lib/bcm/bcm_table_manager.cc
@@ -1604,6 +1604,44 @@ std::set<uint32> BcmTableManager::GetAllAclTableIDs() const {
   return ::util::OkStatus();
 }
 
+::util::Status BcmTableManager::ReadDirectMeterTableEntries(
+    const uint32 table_id, ::p4::v1::ReadResponse* resp) const {
+  // Lookup from the ACL tables.
+  const AclTable* acl_lookup = gtl::FindOrNull(acl_tables_, table_id);
+  if (acl_lookup) {
+    // We shouldn't continue if static flows are detected.
+    if (acl_lookup->IsConst()) {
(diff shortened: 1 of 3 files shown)


miguelbf-alb:
That's fine @pudelkoM, take your time.
Regards

stratum/hal/lib/bcm/bcm_table_manager.cc
```diff
@@ -1604,6 +1604,46 @@ ::util::Status BcmTableManager::ReadTableEntries(
...
+        entry_ptr->mutable_config()->set_cburst(
+            entry_ptr->mutable_table_entry()->meter_config().cburst());
+        entry_ptr->mutable_config()->set_pir(
+            entry_ptr->mutable_table_entry()->meter_config().pir());
+        entry_ptr->mutable_config()->set_pburst(
+            entry_ptr->mutable_table_entry()->meter_config().pburst());
+        // clear meter_config from the table entry
+        entry_ptr->mutable_table_entry()->clear_meter_config();
+      } else {
+        entry_ptr->mutable_config()->set_cir(-1);
```
pierventre, line 1636: does it not break the r/w symmetry ?

pierventre (commented):
if we assume the r/w symmetry as leading principle - this patch breaks it

PR #805, opened 2021-08-06
Allow the configuration of the ADMIN state in StratumBmv2Switch

pierventre:
Add `administate` as constructor parameter to allow users to specify the initial admin state of the ports.


charlesmcchan (approved):

pudelkoM:
@pierventre can you link to a place where this is used? I'd like to get a feel how the interface should be shaped.

pierventre:
> @pierventre can you link to a place where this is used? I'd like to get a feel how the interface should be shaped.
you can find some examples in the up4 example topologies. Here for your reference:

`leaf2 = self.addSwitch('leaf2', cls=StratumBmv2Switch, cpuport=CPU_PORT, adminstate="DISABLED")`

pudelkoM (approved):

PR #808, opened 2021-08-10
Make Stratum-bfrt QoS configurable with singleton port IDs

pudelkoM:
This PR add the option to use `SingletonPort` IDs instead of sdk ports in the QoS config for Tofino.

TODO:

- [x] test on HW


stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -1344,13 +1344,25 @@ ::util::Status BfSdeWrapper::ConfigureQos(
...
   device_to_ppg_handles_[device].clear();
   for (const auto& ppg_config : qos_config.ppg_configs()) {
+    uint32 sdk_port;
+    switch (ppg_config.port_type_case()) {
+      case TofinoConfig::TofinoQosConfig::PpgConfig::PortTypeCase::kSdkPort:
+        sdk_port = ppg_config.sdk_port();
+        break;
+      case TofinoConfig::TofinoQosConfig::PpgConfig::PortTypeCase::kPort: {
+        ASSIGN_OR_RETURN(sdk_port, GetPortIdFromPortKey(
+                                       device, PortKey(0, ppg_config.port())));
```
Yi-Tseng, line 1354: Correct me if I am wrong
`PortKey(0, ppg_config.port())` will create a port key with -1 as channel value and `GetPortIdFromPortKey` will return failed status.

Yi-Tseng: Same for line 1392

pudelkoM: Could be, will test later.

pudelkoM:
The latest commit is an experiment with adding a common `TofinoPort`. Has the drawback that the `oneof` does not directly embed in the parent message anymore, thus needing an extra `port { sdk_port: 1 }` wrapper in the text config.

ccascone (approved):
The proto changes look good to me

pudelkoM:
The common tofino port message did not really safe any code once I found a way to shorten the namespace for the port type case, so I reverted it back to direct embed.

ccascone (approved):

PR #819, opened 2021-08-17
Remove Google specific defines

pudelkoM:
These cause compiler warnings and serve no purpose anymore. `STLStringResizeUninitialized` is removed, as there are no usages anywhere. Should it be required at a later point, we can fall back to Abseils internal version of it: https://github.com/abseil/abseil-cpp/blob/aad2c8a3966424bbaa2f26027d104d17f9c1c1ae/absl/strings/internal/resize_uninitialized.h#L61-L68

### DIFF
--- a/stratum/glue/gtl/stl_util.h
+++ b/stratum/glue/gtl/stl_util.h
@@ -49,18 +49,6 @@ inline const T* vector_as_array(const std::vector<T, Allocator>* v) {
   return v->data();
 }
 
-// Like str->resize(new_size), except any new characters added to "*str" as a
-// result of resizing may be left uninitialized, rather than being filled with
-// '0' bytes. Typically used when code is then going to overwrite the backing
-// store of the string with known data. Uses a Google extension to ::string.
-inline void STLStringResizeUninitialized(std::string* s, size_t new_size) {
-#if __google_stl_resize_uninitialized_string
-  s->resize_uninitialized(new_size);
-#else
-  s->resize(new_size);
-#endif
-}
-
 // Calls delete (non-array version) on the SECOND item (pointer) in each pair in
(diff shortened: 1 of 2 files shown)


bocon13:
Sounds reasonable. We can probably also rename `InitGoogle` to `InitLogging`.

stratum/glue/gtl/stl_util.h
```diff
@@ -54,11 +54,7 @@ inline const T* vector_as_array(const std::vector<T, Allocator>* v) {
 // '0' bytes. Typically used when code is then going to overwrite the backing
 // store of the string with known data. Uses a Google extension to ::string.
 inline void STLStringResizeUninitialized(std::string* s, size_t new_size) {
-#if __google_stl_resize_uninitialized_string
-  s->resize_uninitialized(new_size);
-#else
   s->resize(new_size);
-#endif
 }
```
pudelkoM, line 58: With the removal of the `#if` this whole function can go. There are no usages anyway.

stratum/glue/init_google.h
```diff
@@ -5,21 +5,10 @@
...
-
 #include "gflags/gflags.h"
 #include "stratum/glue/logging.h"
 #include "stratum/glue/stamping.h"
 
-// TODO(unknown) need to define transformation or comment this out on
-// Google's side
-using gflags::ParseCommandLineFlags;
-
 inline void InitGoogle(const char* usage, int* argc, char*** argv,
```
pudelkoM, line 12: Rename to `InitLogging`

pudelkoM: On second thoughts, this function is really about gflags, not logging. And renaming it would make the filename incorrect as well. Let's postpone this for later.

PR #826, opened 2021-09-20
Reject invalid `MeterConfig`s in Stratum-bfrt

pudelkoM:
P4RT meters are trTCM according to [RFC 2698](https://datatracker.ietf.org/doc/html/rfc2698).
Not all P4 `MeterConfig`s possible are actually valid. We don't want to solely rely on the SDE to reject those, therefore we introduce logic in Stratum to catch them early and with helpful error messages.

### DIFF
--- a/stratum/hal/lib/p4/utils.cc
+++ b/stratum/hal/lib/p4/utils.cc
@@ -117,6 +117,32 @@ std::string ByteStringToP4RuntimeByteString(std::string bytes) {
   return bytes;
 }
 
+::util::Status IsValidMeterConfig(const ::p4::v1::MeterConfig& meter_config) {
+  if (meter_config.cir() > meter_config.pir()) {
+    RETURN_ERROR(ERR_INVALID_PARAM)
+        << "Meter configuration " << meter_config.ShortDebugString()
+        << " is invalid: committed rate cannot be greater than peak rate.";
+  }
+  if (meter_config.cburst() > meter_config.pburst()) {
+    RETURN_ERROR(ERR_INVALID_PARAM)
+        << "Meter configuration " << meter_config.ShortDebugString()
+        << " is invalid: committed burst size cannot be greater than peak burst"
+        << " size.";
+  }
+  if (meter_config.cburst() == 0) {
+    RETURN_ERROR(ERR_INVALID_PARAM)
+        << "Meter configuration " << meter_config.ShortDebugString()
+        << " is invalid: committed burst size cannot be zero.";
+  }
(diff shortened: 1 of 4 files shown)


stratum/hal/lib/p4/utils.cc
```diff
@@ -117,6 +117,32 @@ std::string ByteStringToP4RuntimeByteString(std::string bytes) {
...
+    RETURN_ERROR(ERR_INVALID_PARAM)
+        << "Meter configuration " << meter_config.ShortDebugString()
+        << " is invalid: committed rate cannot be greater than peak rate.";
+  }
+  if (meter_config.cburst() > meter_config.pburst()) {
+    RETURN_ERROR(ERR_INVALID_PARAM)
+        << "Meter configuration " << meter_config.ShortDebugString()
+        << " is invalid: committed burst size cannot be greater than peak burst"
+        << " size.";
+  }
```
pudelkoM, line 131: This check does not seem to be an actual requirement from the RFC. Maybe remove?

stratum/hal/lib/p4/utils.cc
```diff
@@ -117,6 +117,26 @@ std::string ByteStringToP4RuntimeByteString(std::string bytes) {
   return bytes;
 }
 
+::util::Status IsValidMeterConfig(const ::p4::v1::MeterConfig& meter_config) {
+  if (meter_config.cir() > meter_config.pir()) {
```
pierventre, line 121: can rate and burst be negative numbers ? according to the spec they are int64

pierventre (commented):
other than that looks good to me

stratum/hal/lib/p4/utils.cc
pudelkoM: Good point:

```json
        {
          "mandatory" : false,
          "read_only" : false,
          "singleton" : {
            "id" : 65547,
            "name" : "$METER_SPEC_PIR_KBPS",
            "repeated" : false,
            "annotations" : [],
            "type" : {
              "type" : "uint64",
              "default_value" : 18446744073709551615
            }
          }
        },
```

bocon13 (commented):
This looks good after addressing @pierventre's comment

pierventre (approved):

bocon13 (approved):

pudelkoM:
I'm totally ok with merging, one remaining thought was that the check should happen in p4_service to cover all backends.
But that opens up the gate to all kinds of validations, p4 pipeline aware and not.
Maybe in the future...

PR #832, opened 2021-10-08
Add LICENSE and NOTICE for stratum_bf[rt] package

bocon13:
- [x] Update BUILD explanation
- [x] Update Intel License

### DIFF
--- a/.circleci/check-bazel-format.sh
+++ b/.circleci/check-bazel-format.sh
@@ -18,10 +18,12 @@ bazel/patches/BUILD
 bazel/rules/BUILD
 bazel/rules/build_tools.bzl
 bazel/rules/lexyacc.bzl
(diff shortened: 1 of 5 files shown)


bocon13:
I plan to follow this up with another change that breaks up NOTICE file by directory and then combines them using a `genrule` before adding it to the binary, but that's not required for the initial version.

pudelkoM:
```
# MISSING COPYRIGHT AND LICENSING INFORMATION

The following files have no licensing information:
* stratum/hal/bin/barefoot/legal/NOTICE
```

Failed license check on a LICENSE file lol

pudelkoM:
Two points

- should we add these files to the package now already?
- I'd be ok with putting them into `stratum/stratum/hal/bin/barefoot/` directly

PR #837, opened 2021-10-09
Drop support for Intel SDE 9.2.0 and 9.3.0

bocon13:
SDE 9.5.0 is now recommended


bocon13:
> There must be a few more bazel targets/defines to remove. Also PI, I think. grep `9_2_0`

Good point. We should remove the enums if we aren't building/testing with the SDE. Updated PR with all refs.

Issue #843, opened 2021-11-02
Remove dependency on Linux in gNOI Admin Service


bocon13:
Stratum currently depends on Linux headers, which means it can't be built natively on Mac.

For example,
```
warning: unknown warning option '-Werror=maybe-uninitialized'; did you mean '-Werror=uninitialized'? [-Wunknown-warning-option]
stratum/hal/lib/common/admin_utils.cc:5:10: fatal error: 'linux/reboot.h' file not found
#include <linux/reboot.h>
         ^~~~~~~~~~~~~~~~
1 warning and 1 error generated.
```

We can use platform defines (e.g. `__linux__`) to help identify the target, so we can use Bazel to determine on a target by target basis what the behavior should be with a select and some custom defines.

For example, we probably don't want want to reboot the machine for stratum_dummy or stratum_bmv2, but we could kill the process.

On Mac in general, we probably just want to return gRPC UNIMPLENTED (instead of rebooting the developer machine).

--------

There's also a dependency in `hal.cc`:
```
In file included from stratum/hal/lib/common/hal.cc:5:
./stratum/hal/lib/common/hal.h:156:28: error: use of undeclared identifier 'sighandler_t'
  absl::flat_hash_map<int, sighandler_t> old_signal_handlers_;
```
This might have some hints on how to address the signal handler issue: https://github.com/MichaelSasser/pdfcrack-ng/pull/2/files

Michaellin250:
Can I take on this issue? :) First time contributing In a while!

bocon13:
@Michaellin250 Go for it! Please let us know if the issue or objective isn't clear enough, and we can provide more details/guidance.

Michaellin250:
Hi. I did a pull request and made the change. Let me know if it works!!

pudelkoM:
Fixed by #1013

PR #846, opened 2021-11-02
Remove dependency on Linux in gNOI Admin Service

Michaellin250:
Fixes issue #843

### DIFF
--- a/stratum/hal/lib/common/admin_utils.cc
+++ b/stratum/hal/lib/common/admin_utils.cc
@@ -2,7 +2,9 @@
 // Copyright 2018-present Open Networking Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-#include <linux/reboot.h>
+#ifdef __linux__
+  #include <linux/reboot.h>
+#endif
 #include <openssl/md5.h>
 #include <openssl/sha.h>
 #include <sys/reboot.h>
--- a/stratum/hal/lib/common/hal.cc
+++ b/stratum/hal/lib/common/hal.cc
@@ -312,7 +312,11 @@ Hal* Hal::GetSingleton() {
   // Register the signal handlers and save the old handlers as well.
   std::vector<int> sig = {SIGINT, SIGTERM, SIGUSR2};
   for (const int s : sig) {
-    sighandler_t h = signal(s, SignalRcvCallback);
+    #ifdef __APPLE__
+    	sig_t h = signal(s, SignalRcvCallback);
+    #else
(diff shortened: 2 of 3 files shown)


stratum/hal/lib/common/hal.cc
```diff
@@ -312,7 +312,11 @@ ::util::Status Hal::RegisterSignalHandlers() {
   // Register the signal handlers and save the old handlers as well.
   std::vector<int> sig = {SIGINT, SIGTERM, SIGUSR2};
   for (const int s : sig) {
-    sighandler_t h = signal(s, SignalRcvCallback);
+    #ifdef __APPLE__
+    	sig_t h = signal(s, SignalRcvCallback);
+    #else
```
pudelkoM, line 317: Not a huge fan of having platform specific code inline with implementations. This could probably be a `typedef`.

pudelkoM:
With these `ifdef`s I counted 4 MacOS specific code variations.
I wonder if it makes sense to collect them into a single `glue/platform.h` file? @bocon13

stratum/hal/lib/common/hal.cc
pudelkoM: In fact, it's the same type:


```c
typedef void (*sig_t) (int);
```

https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/signal.3.html


```c
typedef void (*sighandler_t)(int);
```

https://man7.org/linux/man-pages/man2/signal.2.html

stratum/hal/lib/common/admin_utils.cc
```diff
@@ -2,7 +2,9 @@
 // Copyright 2018-present Open Networking Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-#include <linux/reboot.h>
+#ifdef __linux__
+  #include <linux/reboot.h>
+#endif
```
bocon13, line 7: We also need to make a change in `AdminServiceUtilsInterface::Reboot()` below.

This might be a good candidate for the `glue/platforms.h` file. Define Reboot there and then we can call it here.

pudelkoM:
I looked into this issue a little more and unfortunately found that the `AuthPolicyChecker` uses epoll, a Linux specific polling mechanism: https://github.com/stratum/stratum/blob/main/stratum/lib/security/auth_policy_checker.cc#L9

`AuthPolicyChecker` is used in all externally facing gRPC services, like P4RT, gNMI and gNOI.

Even if we fix the `AdminService`, it still would not build on MacOS.

PR #850, opened 2021-11-05
Update grpc, and update credentials_manager to use FileWatcherCertificateProvider

koconnor29:
Also include a few other small things:

- bump gRPC version from 1.33.2 to 1.35.0
- add a test for credentials_manager
- update go toolchain
- add a missing include statement to admin_service.cc

Fixes: #844

PR #851, opened 2021-11-05
Add missing signal.h includes

bocon13:
Required for failing tests in #850 

Not sure why compilation failures are happening now... possibly due to a dropped include in newer versions of gRPC

cc: @koconnor29


stratum/tools/gnmi/gnmi_cli.cc
```diff
@@ -1,6 +1,8 @@
 // Copyright 2019-present Open Networking Foundation
 // SPDX-License-Identifier: Apache-2.0
 
+#include <signal.h>
```
pudelkoM, line 4: Already includes `<csignal>`

PR #857, opened 2021-12-04
Reject duplicate P4Runtime election IDs from multiple controllers

pudelkoM:
P4Service currently does not reject duplicate election IDs from controllers. This leads to a data corruption when a slave controller sends a MastershipUpdate with the same election ID as the current master:

- 3 Controllers connected: 1 (1111, slave), 2 (2222, master), 3 (1212, slave)
- C1 sends arb. update with election id 2222, same as current master
- Call entry: `AddOrModifyController 1, 2222, ipv4:127.0.0.1:56994.`
- Controller set at beginning of call: `Controller (2, 2222, ipv4:127.0.0.1:56994), Controller (3, 1212, ipv4:127.0.0.1:56994), Controller (1, 1111, ipv4:127.0.0.1:56994), .`
- C1 gets deleted from set: `Controller (2, 2222, ipv4:127.0.0.1:56994), Controller (3, 1212, ipv4:127.0.0.1:56994), .`
- Try to insert C1 into set, but election ID already exists, C1 lost: `Controller (2, 2222, ipv4:127.0.0.1:56994), Controller (3, 1212, ipv4:127.0.0.1:56994), .`
- Stratum falsely logs C1 as new master: `Controller (connection_id: 1, election_id: 2222, uri: ipv4:127.0.0.1:56994) is connected as MASTER for node (aka device) with ID 123123123.`
- C2  receives update: `Controller #2 mastership update: arbitration { device_id: 123123123 election_id { low: 2222 } status { } }`
- C3  receives update: `Controller #3 mastership update: arbitration { device_id: 123123123 election_id { low: 2222 } status { code: 6 message: "You are not my master!" } }`
- No updates sent to C1 and the stream is still open, but not recognized by Stratum as a controller anymore
- C1 might thinks it's master, but all calls will lead to `ERR_PERMISSION_DENIED` errors

This happens because we store Controllers in a `set` with a custom comparison function. Controllers are not hashed, but only compared by their election ID. Thus, two controllers with the same election ID are not possible.
https://github.com/stratum/stratum/blob/34a75cdf38dc0b58062a281228583a767e3cad38/stratum/hal/lib/common/p4_service.h#L235-L240

https://github.com/stratum/stratum/blob/34a75cdf38dc0b58062a281228583a767e3cad38/stratum/hal/lib/common/p4_service.h#L81-L87

This PR changes this by checking for insertion success. Failure indicates a duplicate and we close the stream, all according to spec:

[P4RT Spec - Section 5.3](https://s3-us-west-2.amazonaws.com/p4runtime/docs/master/P4Runtime-Spec.html#sec-arbitration-updates):
> Otherwise, if the MasterArbitrationUpdate message is received from an already connected controller: 

> If the election_id is set and is already used by another controller (excluding the controller making the request) for the same (device_id, role_id), the P4Runtime server shall terminate the stream by returning an INVALID_ARGUMENT error. 

### Additional considerations

Stratum is not fully P4RT spec compliant, with or without this fix. While this change brings us closer, a gap remains. At some point we should consider delegating P4RT related code to an external, well tested library.

Outright closing the stream without telling the offending controller the currently highest election ID puts the controller an an interesting situation:
 - To become master it has to know the current ID, which is only broadcast to backups
 - To become backup it has to find a lower, free election ID by guessing or bruteforce


pudelkoM:
@pierventre FYI

pierventre (approved):

pierventre:
@pudelkoM regarding your additional comments, I agree on using a well tested library. I am not fully sure about your final remarks - in theory an ONOS cluster should not need this and each instance can potentially guess the election id of the other instances. Happy to discuss today. Thanks a lot for this hot fix.

pierventre:
tested - works great

bocon13 (approved):
LGTM

Issue #863, opened 2021-12-09
Revert pi


bocon13:
https://github.com/stratum/stratum/blob/f3a7dacd786ee559b9fa323142f4ce60b12f96f0/stratum/hal/bin/barefoot/build-bf-sde.sh#L161

Issue #862, opened 2021-12-09
Skip stratum_bf for SDE 9.7.0


bocon13:
https://github.com/stratum/stratum/blob/5ef66d996423a9bca0256678c8acf965d8f187a4/tools/release-stratum.sh#L152

PR #859, opened 2021-12-08
Only link against `bfshell_plugin.so` and enable `pi` for SDE 9.7.0 build

Yi-Tseng:
`"barefoot-bin/lib/bfshell_plugin*",` can include some unexpected files when building with older version of SDEs which cause the build failed.

```
ERROR: /root/.cache/bazel/_bazel_root/1b07b1db06ec9ee22c71eac6f0bd4549/external/local_barefoot_bin/BUILD:13:11: in srcs attribute of cc_library rule @local_barefoot_bin//:bfsde: source file '@local_barefoot_bin//:barefoot-bin/lib/bfshell_plugin_bf_rt.la' is misplaced here (expected .cc, .cpp, .cxx, .c++, .C, .cu, .cl, .c, .h, .hh, .hpp, .ipp, .hxx, .h++, .inc, .inl, .tlh, .tli, .H, .tcc, .S, .s, .asm, .a, .lib, .pic.a, .lo, .lo.lib, .pic.lo, .so, .dylib, .dll, .o, .obj or .pic.o). Since this rule was created by the macro 'cc_library', the error might have been caused by the macro implementation
```

PR #864, opened 2021-12-10
Fix stratum_bfrt with SDE 9.7.0

bocon13:
- Partially revert #859
- Don't include PI in 9.7.0 build
- Skip stratum_bf for SDE 9.7.0 in release script


bocon13:
closes #862
closes #863

pudelkoM:
Having `pi` enabled causes a dependency on `libpiall.so` in one of the SDE libs, even for BFRT-only builds. For now we decided to drop Stratum-bf support.

PR #868, opened 2021-12-13
Check meter config only if we are trying to set the meter

Yi-Tseng:
We don't check the meter config when we are trying to reset the meter, which contains an empty meter config in the write request.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager.cc
@@ -932,7 +932,6 @@ BfrtTableManager::ReadDirectCounterEntry(
   CHECK_RETURN_IF_FALSE(meter_entry.meter_id() != 0)
       << "Missing meter id in MeterEntry " << meter_entry.ShortDebugString()
       << ".";
-  RETURN_IF_ERROR(IsValidMeterConfig(meter_entry.config()));
 
   bool meter_units_in_packets;  // or bytes
   {
@@ -960,6 +959,7 @@ BfrtTableManager::ReadDirectCounterEntry(
     meter_index = meter_entry.index().index();
   }
   if (meter_entry.has_config()) {
+    RETURN_IF_ERROR(IsValidMeterConfig(meter_entry.config()));
     RETURN_IF_ERROR(bf_sde_interface_->WriteIndirectMeter(
         device_, session, meter_id, meter_index, meter_units_in_packets,
         meter_entry.config().cir(), meter_entry.config().cburst(),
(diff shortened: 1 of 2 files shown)


pierventre (dismissed):

bocon13 (approved):
LGTM

Should we add a test?

pudelkoM (dismissed):

pudelkoM:
Test without this code change:

```
[ RUN      ] BfrtTableManagerTest.ResetIndirectMeterEntryTest
E20211213 21:26:00.668447 130118 utils.cc:127] StratumErrorSpace::ERR_INVALID_PARAM: 'meter_config.cburst() > 0' is false. Meter configuration  is invalid: committed burst size cannot be less than or equal to zero.
E20211213 21:26:00.668800 130118 bfrt_table_manager.cc:935] Return Error: IsValidMeterConfig(meter_entry.config()) failed with StratumErrorSpace::ERR_INVALID_PARAM: 'meter_config.cburst() > 0' is false. Meter configuration  is invalid: committed burst size cannot be less than or equal to zero.
stratum/hal/lib/barefoot/bfrt_table_manager_test.cc:259: Failure
Value of: bfrt_table_manager_->WriteMeterEntry( session_mock, ::p4::v1::Update::MODIFY, entry)
Expected: is OK
  Actual: StratumErrorSpace::ERR_INVALID_PARAM: 'meter_config.cburst() > 0' is false. Meter configuration  is invalid: committed burst size cannot be less than or equal to zero. (of type util::Status)
stratum/hal/lib/barefoot/bfrt_table_manager_test.cc:239: Failure
Actual function call count doesn't match EXPECT_CALL(*bf_sde_wrapper_mock_, GetBfRtId(kP4MeterId))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
stratum/hal/lib/barefoot/bfrt_table_manager_test.cc:246: Failure
Actual function call count doesn't match EXPECT_CALL(*bf_sde_wrapper_mock_, WriteIndirectMeter( kDevice1, _, kBfRtTableId, Optional(kMeterIndex), false, kUnsetMeterThresholdReset, kUnsetMeterThresholdReset, kUnsetMeterThresholdReset, kUnsetMeterThresholdReset))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
[  FAILED  ] BfrtTableManagerTest.ResetIndirectMeterEntryTest (1 ms)
```

pudelkoM (approved):

Yi-Tseng:
> LGTM
> 
> Should we add a test?

was planning to add, but looks like @pudelkoM did. Thanks!

bocon13 (approved):

PR #872, opened 2021-12-14
Delete `loopback` P4 pipeline and Python rules

pudelkoM:
It's time to clean up some leftovers from ye olden times.

That whole `loopback` directory:

- does not build/run
- is not tested in CI
- not used (?)
- pollutes the dependencies with Python

If you are a user of that P4 pipeline or have other concerns, please reply here. Deletion is postponed until Jan 2022.


bazel/deps.bzl
```diff
@@ -78,13 +78,6 @@ def stratum_deps():
             sha256 = P4RUNTIME_SHA,
         )
 
-    if "build_stack_rules_proto" not in native.existing_rules():
-        remote_workspace(
-            name = "build_stack_rules_proto",
-            remote = "https://github.com/stackb/rules_proto",
-            commit = "2f4e4f62a3d7a43654d69533faa0652e1c4f5082",
-        )
-
```
pudelkoM, line 87: This dep is needed by `rules_python`.
The only remaining Python scripts are https://github.com/stratum/stratum/blob/deddede3f8c5c643c21f8ece5f3c453e3ec659fd/stratum/hal/bin/dummy/BUILD#L15-L32

I'll see if a C++ rewrite or deletion is feasible.

Yi-Tseng:
We should finally be able to merge PR #1 😅
Also we should be able to resolve #341

pudelkoM:
After this change, `bazel query '//...'` will work.

bazel/deps.bzl
```diff
@@ -78,13 +78,6 @@ def stratum_deps():
             sha256 = P4RUNTIME_SHA,
         )
 
-    if "build_stack_rules_proto" not in native.existing_rules():
-        remote_workspace(
-            name = "build_stack_rules_proto",
-            remote = "https://github.com/stackb/rules_proto",
-            commit = "2f4e4f62a3d7a43654d69533faa0652e1c4f5082",
-        )
```
Yi-Tseng, line 86: @pudelkoM I realized that this makes all Jenkins build fail, we need this to build PI related things

PR #876, opened 2021-12-18
Canonicalize `cert_utils` code

pudelkoM:
This PR touches up the recently added `cert_utils`.

- Use correct include guards (`#ifdef`)
- `NULL` -> `nullptr`
- hide internal helper functions
- use `absl::Time` for times
- move `using` directives from header to impl only
- rename file to `certificate`, same as class


stratum/lib/security/BUILD
```diff
@@ -149,12 +149,10 @@ stratum_cc_library(
 
 stratum_cc_test(
     name = "credentials_manager_test",
-    srcs = [
-        "credentials_manager_test.cc",
-    ],
+    srcs = ["credentials_manager_test.cc"],
     data = glob(["testdata/certs*/**"]),
```
bocon13, line 153:
```suggestion
```

No longer needed

stratum/lib/security/BUILD
```diff
@@ -203,27 +201,29 @@ stratum_cc_library(
 )
 
 stratum_cc_library(
-    name = "cert_utils",
+    name = "certificate",
+    # TODO(bocon): seed random number generator before using this in production
+    # code
```
bocon13, line 206: By default, OpenSSL on Linux should seed the RNG automatically from `/dev/urandom` the first time bytes are need to generate an RSA keypair. This is definitely sufficient for testing, and probably sufficient for production use cases.

See https://wiki.openssl.org/index.php/Random_Numbers#Initialization for more details on OpenSSL seeding.

Per the (man page)[https://man7.org/linux/man-pages/man4/random.4.html],
```
The /dev/random interface is considered a legacy interface, and
       /dev/urandom is preferred and sufficient in all use cases, with
       the exception of applications which require randomness during
       early boot time; for these applications, getrandom(2) must be
       used instead, because it will block until the entropy pool is
       initialized.
```

pudelkoM: For context, that's your comment I moved from the [header](https://github.com/stratum/stratum/pull/876/files#diff-46a71815adfa8f084e63ee53e5ae319baa327104993409a9d6d3fb1d39ec23eeL65-L66) to the Bazel rule that a future dev will change to use this lib in non-test code.

If you think it's solved, we can remove the comment and `testonly` restriction.

stratum/lib/security/certificate.cc
```diff
@@ -54,43 +63,42 @@ util::Status GenerateRSAKeyPair(EVP_PKEY* evp, int bits) {
...
 }
 
 util::Status GenerateUnsignedCert(X509* unsigned_cert,
                                   EVP_PKEY* unsigned_cert_key,
                                   const std::string& common_name, int serial,
-                                  int days) {
+                                  absl::Time valid_until) {
   CHECK_RETURN_IF_FALSE(
       ASN1_INTEGER_set(X509_get_serialNumber(unsigned_cert), serial));
   CHECK_RETURN_IF_FALSE(X509_gmtime_adj(X509_get_notBefore(unsigned_cert), 0));
```
bocon13, line 84: This is the start the certificate validity now instead of 1970. :)

stratum/lib/security/certificate.cc
```diff
@@ -54,43 +63,42 @@ util::Status GenerateRSAKeyPair(EVP_PKEY* evp, int bits) {
...
   }
+
   return util::OkStatus();
 }
 
 util::Status GenerateUnsignedCert(X509* unsigned_cert,
                                   EVP_PKEY* unsigned_cert_key,
                                   const std::string& common_name, int serial,
-                                  int days) {
+                                  absl::Time valid_until) {
```
bocon13, line 81:
```suggestion
                                  absl::Time issued_time,
                                  absl::Duration valid_duration) {
```

Not sure if we should have valid_until or a duration if we also include a start time param.

stratum/lib/security/certificate.cc
bocon13, line 84:
```suggestion
  CHECK_RETURN_IF_FALSE(X509_gmtime_adj(X509_getm_notBefore(unsigned_cert), 0));
```

`X509_get_notBefore` is deprecated, use: `X509_getm_notBefore`

stratum/lib/security/certificate.cc
```diff
@@ -99,20 +107,23 @@ util::Status SignCert(X509* unsigned_cert, EVP_PKEY* unsigned_cert_key,
...
+
   return util::OkStatus();
 }
 
 util::Status GenerateSignedCert(X509* unsigned_cert,
                                 EVP_PKEY* unsigned_cert_key, X509* issuer,
                                 EVP_PKEY* issuer_key,
                                 const std::string& common_name, int serial,
-                                int days) {
+                                absl::Time valid_until) {
```
bocon13, line 118: Update signature if `SignCert` changes

stratum/lib/security/BUILD
bocon13: Yup, got it. And, here is a better place for the comment, so thanks for moving it.

Regarding whether this is solved, I think it largely is. I'm not sure what "early" means in the man page, but I think that by the time the system boots, Docker loads, etc., there should be sufficient entropy to generate a short-lived machine-to-machine communication key pair if someone wanted to do that

I think we drop the `testonly` restriction if that's okay with you.

bocon13:
@pudelkoM, thanks for cleaning this up! It looks pretty good. I left a few comments on things that I didn't catch the first time around on the initial PR. We can address them now or in a follow-up PR.

stratum/lib/security/certificate.h
```diff
@@ -0,0 +1,48 @@
...
+
+#ifndef STRATUM_LIB_SECURITY_CERTIFICATE_H_
+#define STRATUM_LIB_SECURITY_CERTIFICATE_H_
+
+#include <memory>
+#include <string>
+
+#include "absl/time/time.h"
+#include "openssl/bio.h"
+#include "openssl/bn.h"
```
bocon13, line 12:
```suggestion
```
Not required in the header.

stratum/lib/security/certificate.cc
pudelkoM: Creating durations longer than hours is a bit involved, as the `absl::<time unit>` helpers stop at `Hours`, for good reason. Absolute `Time` points (`valid_after`, `valid_util` ?) seem intuitive to me.

stratum/lib/security/certificate.cc
bocon13: I can't find docs on this, but experimentally, setting `notBefore` to `0` seems to use `now`.

```
    [issuer] => Array
        (
            [commonName] => Stratum CA
        )

    [version] => 2
    [serialNumber] => 01
    [validFrom] => 220104012144Z
    [validTo] => 220203012144Z
```

stratum/lib/security/BUILD
pudelkoM: removed

stratum/lib/security/certificate.cc
bocon13: Absolute times work for me.

stratum/lib/security/certificate.cc
pudelkoM: Now an input param

bocon13 (approved):
lgtm

PR #877, opened 2021-12-18
Create `pipeline_pusher` tool to replace `update_config.py`

pudelkoM:
This PR creates a minimal P4RT client lib (adapted from PINS), which is then used in the new `pipeline_pusher` tool. The tool takes a p4info and device config as input, establishes mastership on a device and loads the new pipeline.


stratum/lib/p4runtime/p4runtime_session.h
```diff
@@ -0,0 +1,152 @@
...
+  // This stream channel and context are used to perform master arbitration,
+  // but can afterwards also be used for packet IO.
+  std::unique_ptr<grpc::ClientContext> stream_channel_context_;
+  std::unique_ptr<grpc::ClientReaderWriter<p4::v1::StreamMessageRequest,
+                                           p4::v1::StreamMessageResponse>>
+      stream_channel_;
+};
+
+// Creates P4Runtime stub with appropriate channel configuration.
+std::unique_ptr<p4::v1::P4Runtime::Stub> CreateP4RuntimeStub(
```
Yi-Tseng, line 138: We use this function only in one place, can move codes from this function to the `Create` function to avoid creating a shallow module

stratum/tools/p4runtime/pipeline_pusher.cc
```diff
@@ -0,0 +1,60 @@
...
+              "Path to an optional P4Info text proto file. If specified, file "
+              "content will be serialized into the p4info field in "
+              "ForwardingPipelineConfig proto and pushed to the switch.");
+DEFINE_string(p4_pipeline_config_file, "",
+              "Path to an optional P4PipelineConfig bin proto file. If "
+              "specified, file content will be serialized into the "
+              "p4_device_config field in ForwardingPipelineConfig proto "
+              "and pushed to the switch.");
+DEFINE_uint64(device_id, 1, "P4Runtime device ID.");
+
```
Yi-Tseng, line 21: Add election id as optional parameter?

stratum/lib/p4runtime/p4runtime_session.cc
```diff
@@ -0,0 +1,193 @@
...
+  }
+  if (response.arbitration().election_id().low() !=
+      session->election_id_.low()) {
+    RETURN_ERROR(ERR_INTERNAL)
+        << "Lowest 64 bits of received election id doesn't match: "
+        << response.ShortDebugString();
+  }
+
+  // Move is needed to make the older compiler happy.
+  // See: go/totw/labs/should-i-return-std-move.
```
Yi-Tseng, line 131: Please fix the URL here

stratum/lib/p4runtime/p4runtime_session.h
pudelkoM: You're right, as far as I can tell this code is structured with unit tests in mind. PINS has `p4_runtime_session_test.cc`. I'd include that here too, but it requires generating gRPC mocks in the p4runtime repo.

pudelkoM: `P4RuntimeSession::Create` can take a `::p4::v1::P4Runtime::MockStub` as input in unit tests.

`CreateP4RuntimeStub` is the public "production" interface with full setup.

pudelkoM: After some digging, I this is how it works:

`CreateP4RuntimeStub` is used on other tests that interact with P4runtime, but don't want an actual `P4RuntimeSession` with networking and so on.

`P4RuntimeSession::Create` is the interface production apps use. One-shot, batteries included.

pudelkoM: One more point: if we keep the code similar to the upstream PINS sources, we'll have an easier time integrating future updates.

Yi-Tseng (dismissed):
lgtm overall

stratum/lib/p4runtime/p4runtime_session.h
Yi-Tseng: sounds good to me

stratum/lib/p4runtime/p4runtime_session.h
```diff
@@ -0,0 +1,164 @@
...
+  // The election id that has been used to perform master arbitration.
+  ::p4::v1::Uint128 election_id_;
+  // The P4Runtime stub of the switch that this session belongs to.
+  std::unique_ptr<::p4::v1::P4Runtime::Stub> stub_;
+  // This stream channel and context are used to perform master arbitration,
+  // but can now also be used for packet IO.
+  std::unique_ptr<grpc::ClientContext> stream_channel_context_;
+  std::unique_ptr<grpc::ClientReaderWriterInterface<
+      ::p4::v1::StreamMessageRequest, ::p4::v1::StreamMessageResponse>>
+      stream_channel_;
```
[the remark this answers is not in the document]
pudelkoM: I'm missing some detail here. Why is the order important?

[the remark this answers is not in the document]
Yi-Tseng: I think the order of initializing the member in the constructor is based on the order we declare the member in the class

```
stub_(std::move(stub)),
stream_channel_context_(absl::make_unique<grpc::ClientContext>()),
stream_channel_(stub_->StreamChannel(stream_channel_context_.get())) 
```

We need to make sure `stub_` and `stream_channel_context_` is initialized before `stream_channel_`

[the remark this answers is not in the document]
pudelkoM: oh god, yes. Good catch! I'll add a comment.

PR #883, opened 2022-01-04
Add X509v3 extensions to certificate before signing

bocon13:


stratum/lib/security/certificate.cc
```diff
@@ -91,24 +93,79 @@ util::Status GenerateUnsignedCert(X509* unsigned_cert,
...
+          << "Could not add extension " << name << " with value: " << value
+          << ".\n"
+          << std::string(mem->data, mem->length);
+    }
+  }
+
+  // Add the extension to the end of the list
+  CHECK_RETURN_IF_FALSE(X509_add_ext(cert, ext.get(), -1));
+
+  RETURN_OK();
```
pudelkoM, line 124: First time seeing this. wow

bocon13: I was tired of the seg faults with no errors. 😂

Turns out most OpenSSL commands will write to an error buffer before returning; we just need to read them. We could use in other places that we make OpenSSL calls, but here's one way to get some of the error data.

bocon13: And if the wow was for how complicated it is to read a string from OpenSSL... I agree.

We can (should) probably add a helper that takes a BIO and returns a string.

stratum/lib/security/certificate.cc
```diff
@@ -9,14 +9,16 @@
...
 namespace stratum {
 
-using BIGNUM_ptr = std::unique_ptr<BIGNUM, decltype(&::BN_free)>;
-using BIO_ptr = std::unique_ptr<BIO, decltype(&::BIO_free)>;
-using EVP_PKEY_ptr = std::unique_ptr<EVP_PKEY, decltype(&::EVP_PKEY_free)>;
-using X509_ptr = std::unique_ptr<X509, decltype(&::X509_free)>;
+/* TODO(bocon): Copied from a newer version of x509.h
+                Remove when present in the version we use.
+ */
+#define X509_VERSION_3 2
```
pudelkoM, line 21: Can you explain how this define works? Normally does go before the header include, no?

bocon13: It's just a define that can be used to set the version number.

```cpp
  CHECK_RETURN_IF_FALSE(X509_set_version(unsigned_cert, X509_VERSION_3));
```

https://github.com/openssl/openssl/blob/79b2a2f2eedb9d6b24a3f6748332328cf54568fb/include/openssl/x509.h.in#L636

pudelkoM: Got it. I was on a completely different trail, like `_GNU_SOURCE` or similar.

stratum/lib/security/certificate.cc
```diff
@@ -9,132 +9,262 @@
...
-  // Returns a reference to the underlying key; no need to free.
-  RSA* rsa = EVP_PKEY_get0_RSA(pkey);
-  CHECK_RETURN_IF_FALSE(rsa) << "Key is not an RSA key.";
 
-  BIO_ptr bio(BIO_new(BIO_s_mem()), BIO_free);
-  CHECK_RETURN_IF_FALSE(bio.get()) << "Failed to allocate string buffer.";
-  CHECK_RETURN_IF_FALSE(PEM_write_bio_RSAPrivateKey(
-      bio.get(), rsa, nullptr, nullptr, 0, nullptr, nullptr))
-      << "Failed to write private key to buffer.";
+inline bssl::UniquePtr<BIO> NewBio() {
```
pudelkoM, line 30: FYI: inline is only needed for functions definitions in headers so that the ODR is not violated. For helpers in anon namespaces, or in implementation files in general, it's not necessary.

https://en.cppreference.com/w/cpp/language/definition

stratum/lib/security/certificate.cc
```diff
@@ -9,132 +9,262 @@
...
+#include "openssl/x509.h"
+#include "openssl/x509v3.h"
 #include "stratum/lib/macros.h"
 
-namespace stratum {
+/* TODO(bocon): Copied from a newer version of x509.h
+                Remove when present in the version we use.
+ */
+#ifndef X509_VERSION_3
+#define X509_VERSION_3 2
```
pudelkoM, line 20: Now that I understand the purpose, is this not a `constexpr int`?

stratum/lib/security/certificate.cc
pudelkoM: I guess we'd one of those commits:
https://github.com/google/boringssl/commit/89d1241487fb6b715f6516a1f107e846dd760a85
https://github.com/google/boringssl/commit/a96f4dd3822efdfbd371abc4d8ad0e1b9b497866#diff-878093ea6426091505b4c49c59b78924f42859af0eb4ce39b8089bda9577e013R456-R460

pudelkoM: I guess we'd one of those commits:
https://github.com/google/boringssl/commit/89d1241487fb6b715f6516a1f107e846dd760a85
https://github.com/google/boringssl/commit/a96f4dd3822efdfbd371abc4d8ad0e1b9b497866#diff-878093ea6426091505b4c49c59b78924f42859af0eb4ce39b8089bda9577e013R456-R460

stratum/lib/security/certificate.cc
```diff
@@ -148,9 +278,106 @@ util::Status Certificate::GenerateKeyPair(int bits) {
...
+      RETURN_ERROR(ERR_INVALID_PARAM)
+          << "Certificate and private key do not match.";
+    default:
+      auto bio = NewBio();
+      ERR_print_errors(bio.get());
+      RETURN_ERROR(ERR_INVALID_PARAM) << "OpenSSL error checking private key.\n"
+                                      << BioToString(bio);
+  }
+
+  RETURN_OK();
```
pudelkoM, line 375: Did you pick this macro up from PINS? I have not seen any other use of `RETURN_OK` in Stratum.

stratum/lib/security/certificate.cc
bocon13: Good to know. Seems like these days `inline` is more about preventing ODR violations than performance, and the compiler will just inline whatever it can where it makes sense.

stratum/lib/security/certificate.cc
```diff
@@ -148,9 +286,119 @@ util::Status Certificate::GenerateKeyPair(int bits) {
   return GenerateRSAKeyPair(key_.get(), bits);
 }
 
+util::StatusOr<bool> Certificate::IsCA() {
+  switch (X509_check_ca(certificate_.get())) {
+    case 0:  // not CA certificate
+             // int X509_check_issued(X509 *issuer, X509 *subject);
```
bocon13, line 292:
```suggestion
```

stratum/lib/security/certificate.cc
bocon13: We don't have any uses of it, but it's been in the Stratum tree since the beginning:
https://github.com/stratum/stratum/blob/main/stratum/glue/status/status_macros.h#L383

I just happened to notice it and thought I'd try it out. Happy to switch back to:
```c++
return ::util::OkStatus();
```

In PINS, it's mostly: `return absl::OkStatus();`

bocon13:
This PR is basically feature complete now if you want to have a look.

cc: @koconnor29 in case you're interested

stratum/lib/security/certificate.cc
pudelkoM: Interesting. I guess Google had to keep both around because of widespread use.

For us `::util::OkStatus()` would be more consistent, don't you agree?

bocon13: Yup, I agree. changed them over to `util::OkStatus()`

stratum/lib/security/certificate.cc
```diff
@@ -9,132 +9,270 @@
...
-  RSA* rsa = EVP_PKEY_get0_RSA(pkey);
-  CHECK_RETURN_IF_FALSE(rsa) << "Key is not an RSA key.";
 
-  BIO_ptr bio(BIO_new(BIO_s_mem()), BIO_free);
-  CHECK_RETURN_IF_FALSE(bio.get()) << "Failed to allocate string buffer.";
-  CHECK_RETURN_IF_FALSE(PEM_write_bio_RSAPrivateKey(
-      bio.get(), rsa, nullptr, nullptr, 0, nullptr, nullptr))
-      << "Failed to write private key to buffer.";
+// RFC 5280 allows serials numbers up to 20 bytes
+constexpr int SERIAL_NUMBER_LENGTH_BITS = 16 * 8;  // 16 bytes
```
pudelkoM, line 31:
```suggestion
constexpr int kSerialNumberLenghtBits = 16 * 8;  // 16 bytes
```

stratum/lib/security/certificate.cc
```diff
@@ -9,132 +9,270 @@
...
+// A macro for running an OpenSSL function and returning a
+// on error with a util::Status that contains the OpenSSL error.
+#define OPENSSL_RETURN_IF_ERROR(cond)                 \
+  ERR_clear_error();                                  \
+  if (!ABSL_PREDICT_TRUE(cond)) {                     \
+    auto bio = NewBio();                              \
+    ERR_print_errors(bio.get());                      \
+    RETURN_ERROR(ERR_INVALID_PARAM)                   \
+        << "OpenSSL call '" << #cond << "' failed.\n" \
+        << BioToString(bio);                          \
```
pudelkoM, line 68: needs `do while` wrapper

pudelkoM:
Would you mind pointing out the use case for these changes? I have not noticed any callers to the functions.

stratum/lib/security/certificate.cc
bocon13: good catch

bocon13:
@pudelkoM 

There was a little bit of (maybe unnecessary) scope creep in this PR. For testing, X509 v1 certificates are definitely sufficient. However, if we want to start using this in other settings, we should be use v3 certs with proper extensions. That is the original purpose of the PR.

Along the way, debugging was tough, so better error handling was implemented.

> Would you mind pointing out the use case for these changes? I have not noticed any callers to the functions. Just wondering if there are more changes coming.

```cpp
  util::Status LoadCertificate(const std::string& cert_pem,
                               const std::string& key_pem);
  util::StatusOr<bool> IsCA();
  util::Status CheckCommonName(const std::string& name);
  const std::string& GetCommonName();
  util::Status CheckIssuer(const Certificate& issuer);
  util::StatusOr<std::string> GetSerialNumber();
```

These functions were added initially for testing: to validate that the certificates are being generated correctly and to verify the signing chain. However, in the future, they should also be useful for validation for loaded credentials:
https://github.com/stratum/stratum/blob/849cdd6c4fea7a000070363058b2e770919135da/stratum/lib/security/credentials_manager.cc#L67

I didn't want to put them as helps in the test, then move them in the next PR on the topic. However, I'm open to doing so. What do you think?

pudelkoM:
Thanks for explaining!
I was wondering if you were aiming to replace `tools/tls/generate-certs.sh`. Hypothetically, Stratum could be "secure by default" by generating certs on the fly, when not provided by the user.

bocon13:
It should be pretty easy to replace `tools/tls/generate-certs.sh` with a simple C++ tool. It was on my radar. :)

Running Stratum by default with a cert, even a self-signed one, isn't a bad idea, and provides some security benefits. Technically, the gNMI spec requires TLS:

```
All communication between gRPC client and server must be encrypted by TLS (TLS 1.2). The gRPC channel must be in an encrypted state before it is considered usable.
```
https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-authentication.md

PR #885, opened 2022-01-04
Add `CredentialsManager` function to create secure gRPC client `ChannelCredentials`

pudelkoM:
The code to setup secure gRPC channel credentials is currently duplicated in many tools. This PR moves that code into the `CredentialsManager` which already provides an similar interface for servers.

TODO:

- [x] find correct place for this code. `/security`?
- [x] Update ~all~ most tools/CLIs to use this function


stratum/lib/utils.h
```diff
@@ -247,6 +249,28 @@ inline U ByteStreamToUint(const std::string& bytes) {
   return val;
 }
 
+// Create secure TLS gRPC channel credentials for use in clients.
+inline ::util::StatusOr<std::shared_ptr<::grpc::ChannelCredentials>>
+CreateSecureClientGrpcChannelCredentials(
```
bocon13, line 254:
```suggestion
CreateGrpcClientChannelCredentials(
```

We have the same code block in all usages:
```c++
  std::shared_ptr<::grpc::ChannelCredentials> channel_credentials;
  if (!FLAGS_ca_cert.empty()) {
    ASSIGN_OR_RETURN(channel_credentials,
                     CreateSecureClientGrpcChannelCredentials(
                         FLAGS_client_key, FLAGS_client_cert, FLAGS_ca_cert));
  } else {
    channel_credentials = ::grpc::InsecureChannelCredentials();
  }
```
I would suggest moving the if logic into the helper and simplifying the tools.

pudelkoM:
I saw that the `CredentialsManager` has a very similar function for use in gRPC servers: `GenerateExternalFacingServerCredentials`. The manager could provide the same function for clients.

stratum/lib/utils.h
pudelkoM: Done, now part the cred manager

stratum/lib/security/credentials_manager.cc
```diff
@@ -14,53 +14,86 @@
...
 #include "stratum/lib/utils.h"
 
-DEFINE_string(ca_cert_file, "", "CA certificate path");
-DEFINE_string(server_key_file, "", "gRPC Server private key path");
-DEFINE_string(server_cert_file, "", "gRPC Server certificate path");
+DEFINE_string(ca_cert_file, "", "Path to CA certificate file");
+DEFINE_string(server_key_file, "", "Path to gRPC server private key file");
+DEFINE_string(server_cert_file, "", "Path to gRPC server certificate file");
+DEFINE_string(client_key_file, "", "Path to gRPC client key file");
+DEFINE_string(client_cert_file, "", "Path to gRPC client certificate file");
```
pudelkoM, line 21: Could probably merged into one. Having client and server in one binary is unlikely? But I like the `client/server` distinction in the flag name.

pudelkoM: Updating the test code showed that the separation is useful.

stratum/lib/security/credentials_manager.h
```diff
@@ -28,10 +28,19 @@ class CredentialsManager {
...
+  // Generates client credentials for contacting an external facing gRPC server.
+  virtual std::shared_ptr<::grpc::ChannelCredentials>
+  GenerateExternalFacingClientCredentials() const;
+
+  // Loads new server credentials.
+  virtual ::util::Status LoadNewServerCredentials(const std::string& ca_cert,
+                                                  const std::string& cert,
+                                                  const std::string& key);
+
+  // Loads new server credentials.
```
pudelkoM, line 40:
```suggestion
  // Loads new client credentials.
```

stratum/lib/security/credentials_manager.h
```diff
@@ -28,10 +28,19 @@ class CredentialsManager {
   virtual std::shared_ptr<::grpc::ServerCredentials>
   GenerateExternalFacingServerCredentials() const;
 
-  // Loads new credentials.
-  virtual ::util::Status LoadNewCredential(const std::string& ca_cert,
-                                           const std::string& cert,
-                                           const std::string& key);
+  // Generates client credentials for contacting an external facing gRPC server.
+  virtual std::shared_ptr<::grpc::ChannelCredentials>
+  GenerateExternalFacingClientCredentials() const;
```
Yi-Tseng, line 33: Maybe we should change the `Generate` to `Get` in the future since we actually generate client/server credentials when initializing the manager.

Yi-Tseng (approved):
The overall implementation and tool refactor looks good to me. one suggestion below.

stratum/lib/security/credentials_manager.cc
```diff
@@ -9,68 +9,121 @@
...
+#include "absl/time/time.h"
 #include "gflags/gflags.h"
 #include "stratum/glue/logging.h"
 #include "stratum/lib/macros.h"
 #include "stratum/lib/utils.h"
 
-DEFINE_string(ca_cert_file, "", "CA certificate path");
-DEFINE_string(server_key_file, "", "gRPC Server private key path");
-DEFINE_string(server_cert_file, "", "gRPC Server certificate path");
+DEFINE_string(server_ca_cert_file, "", "Path to CA certificate file");
```
bocon13, line 19: There probably isn't much of a need for a separate `server_ca_cert_file` and `client_ca_cert_file`.

I guess the only use case would be a program that is both a server and client where the CA's are different, but that seems a bit contrived to me.

stratum/lib/security/credentials_manager.cc
```diff
@@ -9,68 +9,121 @@
...
   ::util::Status status;
   // TODO(Kevin): Validate the provided key material if possible
   // TODO(max): According to the API of FileWatcherCertificateProvider, any key
   // and certifcate update must happen atomically. The below code does not
   // guarantee that.
-  status.Update(WriteStringToFile(root_certs, FLAGS_ca_cert_file));
+  status.Update(WriteStringToFile(root_certs, FLAGS_server_ca_cert_file));
   status.Update(WriteStringToFile(cert_chain, FLAGS_server_cert_file));
   status.Update(WriteStringToFile(private_key, FLAGS_server_key_file));
+  absl::SleepFor(absl::Seconds(kFileRefreshIntervalSeconds));
```
bocon13, line 109: This is probably fine. However, the convention in gRPC tests seems to be to wait an extra second to avoid a very unlikely race:

https://github.com/grpc/grpc/blob/2d4f3c56001cd1e1f85734b2f7c5ce5f2797c38a/test/core/security/grpc_tls_certificate_provider_test.cc#L316
```suggestion
  absl::SleepFor(absl::Seconds(kFileRefreshIntervalSeconds + 1));
```

pudelkoM: sounds good, will also update the tests.

lol at the comment above the linked code

stratum/lib/security/credentials_manager.cc
pudelkoM: Reverted back to single cert

PR #888, opened 2022-01-06
Re-add `build_stack_rules_proto` dep

Yi-Tseng:
PR #872 removed a few (seemingly) unneeded dependencies. As CI is not building the `stratum_bf/stratum_bfrt` targets we did not catch the missing dependency for older PI versions, which requires `build_stack_rules_proto`.

Note:
Bazel also tries to build PI when building stratum_bfrt target, need to find out the reason.


pudelkoM (approved):
Oh well...

PR #889, opened 2022-01-06
Add bmv2-commit label to mn-stratum image

EmanueleGallone:
This PR adds a label to image `mn-stratum` showing the commit used to build bmv2.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -3,7 +3,7 @@
 
 ARG BAZELISK_VERSION=1.8.0
 ARG PI_COMMIT=a5fd855d4b3293e23816ef6154e83dc6621aed6a
-ARG BMV2_COMMIT=b0fb01ecacbf3a7d7f0c01e2f149b0c6a957f779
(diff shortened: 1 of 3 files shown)


pudelkoM:
Why not just read the BMv2 commit from the Stratum hash? One-liner in bash:

```bash
git show $(docker image inspect --format='{{json .ContainerConfig.Labels}}' opennetworking/mn-stratum:latest | jq -r '."org.opencontainers.image.version"'):Dockerfile.build | grep "BMV2_COMMIT="
```

EmanueleGallone:
Thanks @pudelkoM; While building the image from a more recent bmv2 commit I thought it would've been nice to directly show the bmv2 commit as a label, but your one-liner convinced me 😄 
Closing PR.

pudelkoM:
Happy to help.

My example was primarily to show that the info is already there, albeit a little hidden.

If the this bmv2 commit is something you need very often and by hand, we can revisit this issue.
For one-off lookups or automated processes the above should work.

PR #890, opened 2022-01-06
Pull SDE tars from S3 for automated post-merge Docker builds by CI

pudelkoM:
This PR pulls the SDE tars from S3 storage to enable building of BFRT related targets and pushing of docker images directly from CI. As of now we'll build and push all supported SDE versions in a post-merge to main job. If pre-merge testing all versions becomes important in the future, we can follow up on this later.


.circleci/config.yml
```diff
@@ -23,13 +26,21 @@ save_bazel_cache: &save_bazel_cache
...
 
 clean_bazel_cache: &clean_bazel_cache
   run:
     name: Clean Bazel disk cache of files that have not been modified in 30 days
     # mtime is the only time preserved after untaring the cache.
     command: /usr/bin/find /tmp/bazel-disk-cache -mtime +30 -exec rm -v {} \;
 
+fetch_sde_tars: &fetch_sde_tars
+  aws-s3/sync:
+    # arguments: '--dryrun'
```
pudelkoM, line 39: remove

.circleci/config.yml
```diff
@@ -383,6 +426,12 @@ workflows:
       - publish-docker-stratum-bcm-opennsa:
           requires:
             - publish-docker-build
+      - publish-docker-stratum-bfrt:
+          # requires:
+          #   - publish-docker-build
+          matrix:
+            parameters:
+              sde_version: ["9.5.0", "9.7.0"]
```
pudelkoM, line 434: Add remaining versions

.circleci/config.yml
```diff
@@ -383,6 +426,12 @@ workflows:
       - publish-docker-stratum-bcm-opennsa:
           requires:
             - publish-docker-build
+      - publish-docker-stratum-bfrt:
+          # requires:
+          #   - publish-docker-build
```
pudelkoM, line 431: enable before merge

.circleci/config.yml
```diff
@@ -98,9 +108,11 @@ jobs:
     environment:
       - CC: clang
       - CXX: clang++
+      - SDE_INSTALL_TAR: /tmp/bf-sde-tars/bf-sde-9.7.0-install.tgz
```
pudelkoM, line 111: I'm not sure if the resources spent on building with all SDE version justifies the small benefits. It's really only the sde wrapper and main.cc (which have no tests) that depend on it.

bocon13: S3 sync seems pretty fast (~3s), and the build should mostly be cached (negligible resources). So the most expensive thing here is probably extracting the tarball, which isn't easy to quantify without some additional logging.

Building the binary does force the linker to bring everything together which can uncover some issues.

I'd be curious to see the overhead before making the call not to do it.

pudelkoM: I agree, more tracing is needed. From our TST notes, we concluded to pursue and "main-only" approach at first. Let's solve this later once we have a few runs.

pudelkoM:
There are some size gains to be had on the SDE tars:



```
> tar cf bf-sde-9.8.0-install.tar ./bf-sde-9.8.0-install/
> xz --compress --keep --best ./bf-sde-9.8.0-install.tar
> ll
-rw-r--r--  1 max  wheel   1.0G Feb 27 14:13 bf-sde-9.8.0-install.tar
-rw-r--r--  1 max  wheel   164M Feb 27 14:13 bf-sde-9.8.0-install.tar.xz
-rw-r--r--  1 max  staff   269M Feb  8 12:21 bf-sde-9.8.0-install.tgz
```

pudelkoM:
@zdw We're introducing a CI job that pulls resources from an ONF S3 bucket. The expected bandwidth is low due to caching, but can we monitor this for a while?

PR #892, opened 2022-01-07
Support running Stratum-bfrt on SONiC-based OS

Yi-Tseng:
Use `/etc/sonic/sonic-environment` or ` /etc/sonic/config_db.json` file to load the PLATFORM string.


stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -4,13 +4,18 @@
 set -e
 
 LOG_DIR=${LOG_DIR:-/var/log}
-SDE_VERSION=${SDE_VERSION:-9.5.0}
+SDE_VERSION=${SDE_VERSION:-9.7.0}
```
pudelkoM, line 7: I don't think we should update the default version here/now. How about waiting for the flip in fabric-tna, then updating the default everywhere?

stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -52,5 +57,6 @@ docker run -it --rm --privileged \
     $ONLP_MOUNT \
     $CHASSIS_CONFIG_MOUNT \
     -v $LOG_DIR:/var/log/stratum \
+    --name stratum \
```
pudelkoM, line 60: Good idea, I'd like to apply this to every container

stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -11,6 +11,11 @@ DOCKER_IMAGE_TAG=${DOCKER_IMAGE_TAG:-$SDE_VERSION}
 # Try to load the platform string if not already set.
 if [[ -z "$PLATFORM" ]] && [[ -f "/etc/onl/platform" ]]; then
     PLATFORM=$(cat /etc/onl/platform)
+elif [[ -z "$PLATFORM" ]] && [[ -f "/etc/sonic/sonic-environment" ]]; then
+    source /etc/sonic/sonic-environment
+    PLATFORM=$(echo "$PLATFORM" | sed 's/_/-/g')
```
bocon13, line 16: Do we care about "polluting" this script with other variables from `sonic-environment`?

Note: The variables are not exported to subprocesses unless you call `set -o allexport` before `source`. So this really only affects the code in this file.

If we do care, here's one solution:
```suggestion
    PLATFORM=$(source /etc/sonic/sonic-environment; echo $PLATFORM | sed 's/_/-/g')
```

Test:
```bash
$ echo $(source /etc/sonic/sonic-environment; echo $PLATFORM | sed 's/_/-/g')
x86-64-accton-wedge100bf-32x-r0
```

bocon13, line 16: Do we care about "polluting" this script with other variables from `sonic-environment`?

Note: The variables are not exported to subprocesses unless you call `set -o allexport` before `source`. So this really only affects the code in this file.

If we do care, here's one solution:
```suggestion
    PLATFORM=$(source /etc/sonic/sonic-environment; echo "$PLATFORM" | sed 's/_/-/g')
```

Test:
```bash
$ echo $(source /etc/sonic/sonic-environment; echo $PLATFORM | sed 's/_/-/g')
x86-64-accton-wedge100bf-32x-r0
```

stratum/hal/bin/barefoot/docker/start-stratum-container.sh
```diff
@@ -11,6 +11,11 @@ DOCKER_IMAGE_TAG=${DOCKER_IMAGE_TAG:-$SDE_VERSION}
...
+    # Disable ONLP when running Stratum on SONiC.
+    set -- ${@} --enable_onlp=false
```
bocon13, line 18: This is a little bit cryptic to me.

We might consider explicitly defining:
```
STRATUM_ARGS=${@}
```

Then, just the following here:
```
STRATUM_ARGS+=" --enable_onlp=false"
```

What do you think?

Yi-Tseng: Sounds good to me

bocon13 (approved):
lgtm

pudelkoM (dismissed):

pudelkoM (approved):
LGTM, feel free to merge once you finished your testing on SONiC base OS.

PR #895, opened 2022-01-07
Touch up `CredentialsManager`

pudelkoM:
Some cleanup and application of best practices and convention:

- string const ref
- Member function order
- Comment reflow
- no `using` in headers
- missing virtual on `LoadNewCredential`

### DIFF
--- a/stratum/lib/security/credentials_manager.cc
+++ b/stratum/lib/security/credentials_manager.cc
@@ -20,19 +20,24 @@ DEFINE_string(server_cert_file, "", "gRPC Server certificate path");
 
 namespace stratum {
 
-CredentialsManager::~CredentialsManager() {}
+using ::grpc::experimental::FileWatcherCertificateProvider;
+using ::grpc::experimental::TlsServerCredentials;
+using ::grpc::experimental::TlsServerCredentialsOptions;
+
 CredentialsManager::CredentialsManager() {}
 
-std::shared_ptr<::grpc::ServerCredentials>
-CredentialsManager::GenerateExternalFacingServerCredentials() const {
-  return server_credentials_;
-}
+CredentialsManager::~CredentialsManager() {}
 
 ::util::StatusOr<std::unique_ptr<CredentialsManager>>
 CredentialsManager::CreateInstance() {
   auto instance_ = absl::WrapUnique(new CredentialsManager());
   RETURN_IF_ERROR(instance_->Initialize());
   return std::move(instance_);
+}
+
+std::shared_ptr<::grpc::ServerCredentials>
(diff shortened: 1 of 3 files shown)


stratum/lib/security/credentials_manager.cc
```diff
@@ -56,10 +61,11 @@ ::util::Status CredentialsManager::Initialize() {
 }
 
 ::util::Status CredentialsManager::LoadNewCredential(
-    const std::string root_certs, const std::string cert_chain,
-    const std::string private_key) {
+    const std::string& root_certs, const std::string& cert_chain,
+    const std::string& private_key) {
   ::util::Status status;
   // TODO(Kevin): Validate the provided key material if possible
+  // TODO(max): update must be atomic
```
bocon13, line 68: Agreed. However, this race is unlikely to cause an issue.

1. Certificates rotate relatively infrequently / never.
2. The FileWatcher polls every second making the race extremely unlikely.
3. If the race does occur, it should be resolved on the next poll for a maximum outage of the polling interval.
4. The client will likely automatically retry in the event of a failed connection.

All of that being said, if there's a simple fix, let's do it. But I don't think I'd be too interested in something super complicated.

bocon13: Two other thoughts:

We can force an update after writing the files: `FileWatcherCertificateProvider::ForceUpdate()`
This doesn't solve the race, but reduces the error duration after the race occurs.

I tried to look through to code to see if a key mismatch is picked up by gRPC.
```
@grpc/src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.cc
@grpc/src/core/lib/security/credentials/tls/grpc_tls_certificate_distributor.cc
```
On quick glance, it's hard to tell if/what types of errors are handled. It might be easier to verify behavior experimentally.

bocon13 (dismissed):
One comment on atomicity; otherwise LGTM

stratum/lib/security/credentials_manager.cc
pudelkoM: yeah, I took this as a note for later when reading the API:
https://github.com/grpc/grpc/blob/ea389c00c2c83e0364c5b1cfb73820c0101daa14/include/grpcpp/security/tls_certificate_provider.h#L83-L90

bocon13: Ah, that makes sense.
```suggestion
  // TODO(max): update must be atomic (See FileWatcherCertificateProvider comments in tls_certificate_provider.h)
```

pudelkoM: good idea with the reference, will do that.

PR #898, opened 2022-01-13
Simplify bazel coverage flags

Yi-Tseng:
Bazel now uses '@bazel_tools//tools/test:coverage_report_generator' as
default coverage report generator and it should work.

TODO: see if CI can generate correct report

### DIFF
--- a/.circleci/bazelrc
+++ b/.circleci/bazelrc
@@ -28,5 +28,5 @@ test --test_verbose_timeout_warnings
 build --profile=/tmp/bazel.profile.json
 
 # Coverage
-coverage --combined_report=lcov --coverage_report_generator @bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main
+coverage --combined_report=lcov
 


pudelkoM:
Didn't we commit the entire script because of some exploit a while ago? I'm fine with this if we believe that circleci does the checking for us.

Yi-Tseng:
> Didn't we commit the entire script because of some exploit a while ago? I'm fine with this if we believe that circleci does the checking for us.

Yes, but I think Codecov adds some checks to verify the hash of the script

Yi-Tseng:
I am also fine to keep the script in Stratum so we don't need to use the orb

.circleci/config.yml
```diff
@@ -3,6 +3,8 @@
 
 version: 2.1
 
+orbs:
+  codecov: codecov/codecov@3.2.2
 # Cache Bazel output root (see bazelrc) to speed up job execution time.
```
pudelkoM, line 8: Maybe add a newline here

pudelkoM (approved):

PR #899, opened 2022-01-13
[SDFAB-901] Support P4Runtime translation in Stratum-bfrt

Yi-Tseng:
This PR adds P4Runtime translation support.

- The P4RuntimeBfrtTranslator class implementation contains most of the translation logic. (And unit tests)
- Include the translator in every `Bfrt*Manager` class.
  - Translate entry when reading and writing
- Initialize the translator in the main function of the stratum_bfrt.
- Add a new flag `--experimental_enable_p4runtime_translation` to enable translation, disabled by default.

TODO:
- [x] Refactor with deeper module design (Only expose a few APIs from the translator)
- [x] TableEntry support
- [x] #904 
- [x] #905 
- [x] #907 
- [x] #908
- [x] Meter, DirectMeter
- [x] Register
- [x] Verify the implementation with Tofino Model (See: https://github.com/stratum/fabric-tna/pull/455)


stratum/hal/lib/barefoot/p4runtime_bfrt_translator.h
```diff
@@ -0,0 +1,165 @@
...
+#include "stratum/hal/lib/barefoot/bf_sde_interface.h"
+#include "stratum/hal/lib/common/common.pb.h"
+#include "stratum/hal/lib/common/writer_interface.h"
+#include "stratum/lib/macros.h"
+
+namespace stratum {
+namespace hal {
+namespace barefoot {
+
+class P4RuntimeBfrtTranslator {
```
Yi-Tseng, line 26: Or just `BfrtTranslator`?

stratum/hal/lib/barefoot/bfrt_node_test.cc
```diff
@@ -193,68 +207,34 @@ class BfrtNodeTest : public ::testing::Test {
   static constexpr int kLogicalPortId = 35;
   static constexpr uint32 kPortId = 941;
   static constexpr uint32 kL2McastGroupId = 20;
-  static constexpr char kBfConfigPipelineString[] = R"PROTO(
+  static constexpr char kBfConfigPipelineString[] = R"pb(
```
pudelkoM, line 210: This change comes from a newer version of clang-format (I think 12 or 13). For now we can leave it at `PROTO`.

stratum/hal/lib/barefoot/utils.h
```diff
@@ -47,6 +47,11 @@ ::util::StatusOr<int32> ConvertPriorityFromBfrtToP4rt(uint64 priority);
 // byte string.
 int NumBitsToNumBytes(int num_bits);
 
+// Converts unsigned integer to a byte array that fits certain bit width.
+// Returns error if the value is bigger than the size of bit width can store.
+::util::StatusOr<std::string> Uint32ToBytes(uint32 value, size_t bitwidth);
+::util::StatusOr<uint32> BytesToUint32(std::string value);
```
pudelkoM, line 53: Could `ByteStreamToUint` and `Uint32ToByteStream` cover these two functions? Main difference is the error handling.

stratum/hal/lib/barefoot/bfrt_node.cc
```diff
@@ -20,6 +21,9 @@
 #include "stratum/lib/utils.h"
 #include "stratum/public/proto/error.pb.h"
 
+DEFINE_bool(experimental_enable_p4runtime_translation, false,
```
pudelkoM, line 24: Using Bazels flag policy: https://docs.bazel.build/versions/main/command-line-reference.html#option-metadata-tags

stratum/hal/lib/barefoot/utils.h
Yi-Tseng: Didn't know we have this utility, will use that

stratum/hal/lib/barefoot/bfrt_node.cc
```diff
@@ -195,43 +209,96 @@ ::util::Status BfrtNode::WriteForwardingEntries(
...
+        ::p4::v1::TableEntry table_entry = update.entity().table_entry();
+        if (translation_enabled_) {
+          ASSIGN_OR_RETURN(table_entry,
+                           p4runtime_bfrt_translator_->TranslateTableEntry(
+                               table_entry, true));
+        }
+        status = bfrt_table_manager_->WriteTableEntry(session, update.type(),
+                                                      table_entry);
         break;
+      }
```
pudelkoM, line 222: A thought on how to cut down the boilerplate code by having a generic `TranslatedEntity` function that we call before the `switch` with `update.entity()`. Inside that we check for the flag and either return it unmodified or translated.
What do you think?

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.h
```diff
@@ -0,0 +1,165 @@
...
+    : public WriterInterface<::p4::v1::ReadResponse> {
+ public:
+  bool Write(const ::p4::v1::ReadResponse& msg) override;
+  ~P4RuntimeBfrtTranslationWriterWrapper() override {}
+  static std::unique_ptr<P4RuntimeBfrtTranslationWriterWrapper> CreateInstance(
+      WriterInterface<::p4::v1::ReadResponse>* writer,
+      P4RuntimeBfrtTranslator* p4runtime_bfrt_translator) {
+    return absl::WrapUnique(new P4RuntimeBfrtTranslationWriterWrapper(
+        writer, p4runtime_bfrt_translator));
+  }
```
pudelkoM, line 138: Since we're not doing any work here, we can probably use a normal contructor. See: `ServerWriterWrapper` and `ChannelWriterWrapper`. The managers generally have a `CreateInstance` for consistency, because _some_ do work.

stratum/hal/lib/barefoot/utils.h
pudelkoM: You can roll over your new test cases if you want.

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.h
pudelkoM: I'm not sure if this could be a generic P4 component independent from bfrt. So we can do the same for bmv2 or bcm?

stratum/hal/lib/barefoot/bfrt_node.cc
pudelkoM: This could also move `translation_enabled_` and the flag into the translator file.

Yi-Tseng: Yes, actually I am working on having generic functions for read/write request and stream messages

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.h
Yi-Tseng: The `TranslateValue` function is specific for TNA and Barefoot's device

pudelkoM: Just the `"tna/PortId_t"` string, right?

Yi-Tseng: Right, both `TranslateValue` and `TranslateTnaPortId` are specific for TNA/BF

Yi-Tseng: Oh, also when chassis config is pushed, we need to collect port mapping from SDE API

stratum/hal/lib/barefoot/utils.cc
```diff
@@ -7,6 +7,7 @@
 #include <algorithm>
 #include <utility>
 
+#include "stratum/glue/gtl/stl_util.h"
```
Yi-Tseng, line 10: to be removed

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.h
```diff
@@ -0,0 +1,175 @@
...
+  // Private constructor. Use CreateInstance() to create an instance of this
+  // class.
+  P4RuntimeBfrtTranslator(BfSdeInterface* bf_sde_interface, int device_id,
+                          bool translation_enabled)
+      : device_id_(device_id),
+        bf_sde_interface_(bf_sde_interface),
+        translation_enabled_(translation_enabled),
+        pipeline_require_translation_(false) {}
+  virtual ::util::StatusOr<::p4::v1::Entity> TranslateEntity(
+      const ::p4::v1::Entity& entity, bool to_sdk) SHARED_LOCKS_REQUIRED(lock_);
```
pudelkoM, line 71: Since these functions are private, it's not a huge issue, but https://abseil.io/tips/94 comes to mind.

Yi-Tseng: I will add a comment to the parameter since it will be more clear to other developers.

pudelkoM: comment seems the best, enum would have been overkill probably.

stratum/hal/lib/barefoot/bfrt_constants.h
```diff
@@ -67,6 +70,14 @@ constexpr int kMaxQueuesPerPort = 32;
 
 constexpr absl::Duration kDefaultSyncTimeout = absl::Seconds(1);
 
+// P4Runtime Translation.
+constexpr int32 kTnaPortIdBitWidth = 9;
+constexpr char kUriTnaPortId[] = "tna/PortId_t";
+const absl::flat_hash_map<std::string, int32> kUriToBitWidth = {
+    {kUriTnaPortId, kTnaPortIdBitWidth}};
+constexpr uint32 kSdnTnaRecirculationPortBase = 0xFFFFFF00;
+constexpr uint32 kTnaRecirculationPortBase = 0x44;
```
pudelkoM, line 79: I think those are not used outside the translator, thus could be local to that class.

stratum/hal/lib/barefoot/bfrt_node.cc
```diff
@@ -131,6 +138,14 @@ ::util::Status BfrtNode::CommitForwardingPipelineConfig() {
...
 
   // Push pipeline config to the managers.
+  RETURN_IF_ERROR(p4runtime_bfrt_translator_->PushForwardingPipelineConfig(
+      bfrt_config_.programs(0).p4info()));
+
+  // Augment the P4Info so managers will use the original bitwith from P4 code
+  // for every fields (e.g, 9-bit port number instead of 32-bits).
+  ASSIGN_OR_RETURN(const auto& low_level_p4info,
+                   p4runtime_bfrt_translator_->GetLowLevelP4Info());
+  *bfrt_config_.mutable_programs(0)->mutable_p4info() = low_level_p4info;
```
pudelkoM, line 148: This is not a good program flow / API design. In the node we first save the p4info, then push it to a sub-manager, then read it back from it and overwrite to local version.

stratum/hal/lib/barefoot/bfrt_node.cc
```diff
@@ -188,11 +202,12 @@ ::util::Status BfrtNode::WriteForwardingEntries(
...
     return MAKE_ERROR(ERR_NOT_INITIALIZED) << "Not initialized!";
   }
-
+  ASSIGN_OR_RETURN(const auto& request,
+                   p4runtime_bfrt_translator_->TranslateWriteRequest(req));
   bool success = true;
   ASSIGN_OR_RETURN(auto session, bf_sde_interface_->CreateSession());
   RETURN_IF_ERROR(session->BeginBatch());
-  for (const auto& update : req.updates()) {
+  for (const auto& update : request.updates()) {
```
pudelkoM, line 210: I think this violates the p4rt guarantees around batching when translation fails:

> CONTINUE_ON_ERROR: This is the default behavior and the default enum value. Each operation within the batch must be attempted even if one or more encounter errors.

Should we translate one entity at a time?

stratum/hal/lib/barefoot/bfrt_node_test.cc
```diff
@@ -1131,6 +1165,9 @@ TEST_F(BfrtNodeTest, WriteForwardingEntriesSuccess_InsertActionProfileGroup) {
 //   EXPECT_CALL(*bfrt_table_manager_mock_,
 //               AddActionProfileGroup(EqualsProto(*group), kEgressIntfId))
 //       .WillOnce(Return(::util::OkStatus()));
+//   EXPECT_CALL(*p4runtime_bfrt_translator_mock_,
+//   TranslateWriteRequest(EqualsProto(req)))
+//       .WillOnce(Return(::util::StatusOr<::p4::v1::WriteRequest>(req)));
```
pudelkoM, line 1170: Can this commented in again?

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+  // Port mapping for P4Runtime translation.
+  singleton_port_to_sdk_port_.clear();
+  sdk_port_to_singleton_port_.clear();
+  // Initialize with special ports.
+  singleton_port_to_sdk_port_[kSdnUnspecifiedPortId] = 0;
+  ASSIGN_OR_RETURN(const auto& cpu_sdk_port,
+                   bf_sde_interface_->GetPcieCpuPort(device_id_));
+  singleton_port_to_sdk_port_[kSdnCpuPortId] = cpu_sdk_port;
+  sdk_port_to_singleton_port_[cpu_sdk_port] = kSdnCpuPortId;
+  for (int pipe = 0; pipe < 4; pipe++) {
```
pudelkoM, line 88:
```suggestion
  for (int pipe = 0; pipe < kTnaMaxNumPipes; pipe++) {
```

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+  low_level_p4info_ = p4info;
+  pipeline_require_translation_ = false;
+  if (!translation_enabled_) {
+    return ::util::OkStatus();
+  }
+  // Enable P4Runtime translation when user define a new type with
+  // p4runtime_translation.
+  if (p4info.has_type_info()) {
+    // First, store types that need to be translated(will check the type_name
+    // later)
```
pudelkoM, line 121:
```suggestion
    // First, store types that need to be translated (will check the type_name
    // later).
```

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+          ::p4::config::v1::P4NewTypeSpec::kTranslatedType) {
+        pipeline_require_translation_ = true;
+        // TODO(Yi Tseng): Verify URI string
+        type_name_to_uri[type_name] = value.translated_type().uri();
+        if (value.translated_type().sdn_type_case() ==
+            ::p4::config::v1::P4NewTypeTranslation::kSdnBitwidth) {
+          type_name_to_bit_width[type_name] =
+              value.translated_type().sdn_bitwidth();
+        } else {
+          // TODO(Yi Tseng): support SDN String translation.
```
pudelkoM, line 137: Should we fail here? Silent drop seems unwise.

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+    sdk_port_to_singleton_port_[sdk_port_id] = singleton_port_id;
+  }
+
+  return ::util::OkStatus();
+}
+
+::util::Status P4RuntimeBfrtTranslator::PushForwardingPipelineConfig(
+    const ::p4::config::v1::P4Info& p4info) {
+  ::absl::WriterMutexLock l(&lock_);
+  low_level_p4info_ = p4info;
```
pudelkoM, line 112: Generally, it's better to commit new state at the end after success. I.e. prepare a local low level map, then set it last.


See: https://github.com/stratum/stratum/blob/p4rt-translation/stratum/hal/lib/barefoot/bf_chassis_manager.cc#L582-L595

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+      for (auto& match_field : *table.mutable_match_fields()) {
+        if (match_field.has_type_name()) {
+          const auto& type_name = match_field.type_name().name();
+          const auto& table_id = table.preamble().id();
+          const auto& match_field_id = match_field.id();
+          std::string* uri = gtl::FindOrNull(type_name_to_uri, type_name);
+          if (uri) {
+            table_to_field_to_type_uri_[table_id][match_field_id] = *uri;
+            // Replace bitwidth to the low level one.
+            CHECK_RETURN_IF_FALSE(kUriToBitWidth.count(*uri));
```
pudelkoM, line 168: move check first

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+  const auto& member_id = act_prof_mem.member_id();
+  ASSIGN_OR_RETURN(*(translated_apm.mutable_action()),
+                   TranslateAction(translated_apm.action(), to_sdk));
+  return translated_apm;
+}
+
+::util::StatusOr<::p4::v1::MeterEntry>
+P4RuntimeBfrtTranslator::TranslateMeterEntry(const ::p4::v1::MeterEntry& entry,
+                                             bool to_sdk) {
+  // TODO(Yi Tseng): Will support this in another PR.
```
pudelkoM, line 464: I don't think there is anything to translate for indirect meters. Or just `TranslateIndex`?

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+  return entry;
+}
+
+::util::StatusOr<::p4::v1::DirectMeterEntry>
+P4RuntimeBfrtTranslator::TranslateDirectMeterEntry(
+    const ::p4::v1::DirectMeterEntry& entry, bool to_sdk) {
+  // TODO(Yi Tseng): Will support this in another PR.
```
pudelkoM, line 471: Just call `TranslateTableEntry`?

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+                     bf_sde_interface_->GetPortIdFromPortKey(
+                         device_id_, singleton_port_key));
+    singleton_port_to_sdk_port_[singleton_port_id] = sdk_port_id;
+    sdk_port_to_singleton_port_[sdk_port_id] = singleton_port_id;
+  }
+
+  return ::util::OkStatus();
+}
+
+::util::Status P4RuntimeBfrtTranslator::PushForwardingPipelineConfig(
```
pudelkoM, line 109: > Note that it may be infeasible to translate the value-mask pair for ternary matches: LPM, TERNARY or RANGE match kinds. The P4Runtime server may require that for these match kinds the port match be either de facto “exact” (0xffffffff mask for TERNARY, prefix-length of 32 for LPM, or same low and high bounds for RANGE) or “don't care”. 

~~Are we checking this somewhere?~~ ok, we do in `TranslateTableEntry`. Can we move this check into `PushForwardingPipelineConfig`? Much earlier.

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+  ::p4::v1::TableEntry translated_entry(entry);
+  const auto& table_id = translated_entry.table_id();
+  if (table_to_field_to_type_uri_.count(table_id) &&
+      table_to_field_to_bit_width_.count(table_id)) {
+    for (::p4::v1::FieldMatch& field_match :
+         *translated_entry.mutable_match()) {
+      const auto& field_id = field_match.field_id();
+      std::string* uri =
+          gtl::FindOrNull(table_to_field_to_type_uri_[table_id], field_id);
+
```
pudelkoM, line 346:
```suggestion
  if (!uri) continue;
```

?

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+            ASSIGN_OR_RETURN(const std::string& new_val,
+                             TranslateValue(field_match.exact().value(), *uri,
+                                            to_sdk, to_bit_width));
+            field_match.mutable_exact()->set_value(new_val);
+            break;
+          }
+          case ::p4::v1::FieldMatch::kTernary: {
+            // We only allow the "exact" type of ternary match, which means
+            // all bits from mask must be one.
+            std::string all_one = std::string(from_bit_width / 8, '\xff');
```
pudelkoM, line 370: Is truncation intended here? `7 / 8 = 0`

```suggestion
            std::string all_one = std::string(NumBitsToNumBytes(from_bit_width), '\xff');
```

ok, I see that the next line inserts the leading byte. I noticed that `RangeDefaultHigh` does the same thing, chance of code reuse?

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+}
+
+::util::StatusOr<::p4::v1::PacketOut>
+P4RuntimeBfrtTranslator::TranslatePacketOut(
+    const ::p4::v1::PacketOut& packet_out) {
+  ::p4::v1::PacketOut translated_packet_out(packet_out);
+  for (auto& md : *translated_packet_out.mutable_metadata()) {
+    const std::string* uri =
+        gtl::FindOrNull(packet_out_meta_to_type_uri_, md.metadata_id());
+    if (uri) {
```
pudelkoM, line 607:
```suggestion
    if (!uri)
      continue;
```

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
```diff
@@ -0,0 +1,742 @@
...
+         // translated.
+    }
+  }
+  return translated_action;
+}
+
+::util::StatusOr<std::string> P4RuntimeBfrtTranslator::TranslateValue(
+    const std::string& value, const std::string& uri, bool to_sdk,
+    int32 bit_width) {
+  if (uri.compare(kUriTnaPortId) == 0) {
```
pudelkoM, line 709:
```suggestion
  if (uri == kUriTnaPortId) {
```

stratum/hal/bin/barefoot/BUILD
```diff
@@ -74,6 +74,7 @@ stratum_cc_binary(
         "//stratum/hal/lib/barefoot:bfrt_pre_manager",
         "//stratum/hal/lib/barefoot:bfrt_table_manager",
         "//stratum/hal/lib/barefoot:bfrt_counter_manager",
+        "//stratum/hal/lib/barefoot:p4runtime_bfrt_translator",
```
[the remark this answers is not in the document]
Yi-Tseng: Sounds good, We can also rename the class to BfrtP4RuntimeTranslator

stratum/hal/lib/barefoot/bfrt_node.cc
Yi-Tseng: I think that makes sense. Since we should not reject the entire request if there is an invalid entry.
Do you think we can just call `TranslateEntity` at the beginning of the for-loop?

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
Yi-Tseng: I add a new function `MaxValueOfBits` which will be used by `RangeDefaultHigh` and here.

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
Yi-Tseng: Sorry, I think I don't understand your question.
We need to check the value/mask/prefix_len/range of match fields for each table entry when we write table entries. That's the reason I put check in the `TranslateTableEntry`

stratum/hal/lib/barefoot/bfrt_node.cc
Yi-Tseng: Maybe we can also have the same logic for the translator (save and commit). 
And we store a translated version when we are creating `bfrt_config_`?

https://github.com/stratum/stratum/blob/main/stratum/hal/lib/barefoot/bfrt_node.cc#L109
```diff
- *program->mutable_p4info() = config.p4info();
+ *program->mutable_p4info() = bfrt_p4runtimr_translator_.TranslateP4Info(config.p4info());
```

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
Yi-Tseng: Right, this is an easy task. I include the support and test for Meter, DirectMeter, and Register in the PR

stratum/hal/lib/barefoot/p4runtime_bfrt_translator.cc
pudelkoM: my bad, I confused two things here. The check for `0xffffffff`/prefix len/don't care can, of course, only happen at flow push time.

stratum/hal/lib/barefoot/bfrt_node.cc
pudelkoM: > Do you think we can just call TranslateEntity at the beginning of the for-loop?

That should work.

stratum/hal/lib/barefoot/bfrt_node.cc
Yi-Tseng: Maybe we should always keep the original version(one form the control plane to the bfrt node) and push the translated version to managers

pudelkoM: I think there is a way to structure this. Everything above BfrtNode is (potentially) untranslated. In the write path, the node translates all requests to SDK form and passes them on to the managers. Therefore, everything below node is translated. So we can and should pass the low level p4info down to the managers.

pudelkoM: Looking deeper into the `P4TableMapper`, I think that design is the way to go.

Yi-Tseng: > I think there is a way to structure this:
>     * Everything above BfrtNode is (potentially) untranslated.
>     * In the write path, the node translates all requests to SDK form and passes them on to the managers.
>     * Therefore, everything below the node is translated.
>     * So we can and should pass the low level p4info down to the managers.

I think that's what I am currently doing now.
I add a new change to the translator and the bfrt node. Now bfrt node stores the untouched pipeline config.
When committing pipeline config to bfrt mangers, bfrt node will create a copy for the pipeline config and modify it and push to those managers.

I can also add the translator to those managers, the reason I didn't put is simply want to reduce the total number of dependency.

pudelkoM: > I think that's what I am currently doing now. (unless I missed something?)

no, that's right. It's a summary of the current state.

Yi-Tseng: got it.
I am trying to figure out roles for different components, for example

bfrt_node: pass P4Runtime requests to managers that can handle them
bfrt_x_manager: Translate P4Runtime messages to the internal data structure and pass to the SDE interface.
bfrt_p4runtime_translator: Translate P4Runtime message to another P4Runtime message with a different value

And looks like P4TableMapper does both translating the data structure(P4- > common) and maybe the value. And bcm managers translate the common entry to the bcm specific one.

Yi-Tseng: I think we can move the translator to those managers to keep the bfrt node clean.
And calls the translation in manager functions when bfrt node pass P4Runtime requests to the manager
so we can keep some logic in the lower level

pudelkoM: jep, this might work

stratum/hal/bin/barefoot/main_bfrt.cc
```diff
@@ -23,6 +24,8 @@ DEFINE_bool(bf_switchd_background, false,
             "Run bf_switchd in the background with no interactive features");
 DEFINE_string(bf_switchd_cfg, "stratum/hal/bin/barefoot/tofino_skip_p4.conf",
               "Path to the BF switchd json config file");
+DEFINE_bool(experimental_enable_p4runtime_translation, false,
+            "Enable experimental P4runtime translation feature.");
```
pudelkoM, line 28:
```suggestion
            "Enable experimental P4Runtime translation feature.");
```

stratum/hal/lib/barefoot/bfrt_counter_manager.h
```diff
@@ -52,14 +54,20 @@ class BfrtCounterManager {
  private:
   // Private constructor, we can create the instance by using `CreateInstance`
   // function only.
-  explicit BfrtCounterManager(BfSdeInterface* bf_sde_interface_, int device);
+  explicit BfrtCounterManager(
+      BfSdeInterface* bf_sde_interface_,
+      BfrtP4RuntimeTranslator* bfrt_p4runtime_translator, int device);
```
pudelkoM, line 59: Can you move the sde interface second to last?
It's like in BfrtNode: <managers>, sde interface, device

stratum/hal/lib/barefoot/bfrt_p4runtime_translator.h
```diff
@@ -0,0 +1,172 @@
...
+      : device_id_(0),
+        bf_sde_interface_(nullptr),
+        translation_enabled_(false),
+        pipeline_require_translation_(false) {}
+
+ private:
+  // Private constructor. Use CreateInstance() to create an instance of this
+  // class.
+  BfrtP4RuntimeTranslator(BfSdeInterface* bf_sde_interface, int device_id,
+                          bool translation_enabled)
```
pudelkoM, line 95:
```suggestion
  BfrtP4RuntimeTranslator(bool translation_enabled, 
                          BfSdeInterface* bf_sde_interface, int device_id)
```

sde interface and device last.

stratum/hal/lib/barefoot/utils.h
```diff
@@ -47,6 +47,9 @@ ::util::StatusOr<int32> ConvertPriorityFromBfrtToP4rt(uint64 priority);
 // byte string.
 int NumBitsToNumBytes(int num_bits);
 
+// Gets the max value of certain bitwidth.
+std::string MaxValueOfBits(size_t bitwidth);
+
```
pudelkoM, line 52: I like the naming PINS uses for something similar:

```c++
// Returns if a (normalized) byte string is all zeros.
bool IsAllZeros(const std::string &s);
```

```suggestion
// Returns a byte string with bitwidth one bits set.
std::string AllOnesByteString(size_t bitwidth);

```

stratum/hal/lib/barefoot/utils.h
```diff
@@ -47,6 +47,9 @@ ::util::StatusOr<int32> ConvertPriorityFromBfrtToP4rt(uint64 priority);
 // byte string.
 int NumBitsToNumBytes(int num_bits);
 
+// Gets the max value of certain bitwidth.
```
pudelkoM, line 50:
```suggestion
// Returns a byte string with bitwidth bits set to zero.
```

pudelkoM (approved):

pudelkoM (approved):

stratum/hal/lib/barefoot/bfrt_node.cc
```diff
@@ -125,21 +132,23 @@ ::util::Status BfrtNode::CommitForwardingPipelineConfig() {
   if (!initialized_) {
     return MAKE_ERROR(ERR_NOT_INITIALIZED) << "Not initialized!";
   }
-  CHECK_RETURN_IF_FALSE(bfrt_config_.programs_size() > 0);
+  BfrtDeviceConfig bfrt_config(bfrt_config_);
+  CHECK_RETURN_IF_FALSE(bfrt_config.programs_size() > 0);
```
pudelkoM, line 136: This copy is not needed anymore, right?

stratum/hal/lib/barefoot/bfrt_p4runtime_translator.cc
```diff
@@ -0,0 +1,695 @@
...
+    absl::flat_hash_map<uint32, int32>* meta_to_bit_width;
+    if (ctrl_hdr_name == kIngressMetadataPreambleName) {
+      meta_to_type_uri = &packet_in_meta_to_type_uri;
+      meta_to_bit_width = &packet_in_meta_to_bit_width;
+    } else if (ctrl_hdr_name == kEgressMetadataPreambleName) {
+      meta_to_type_uri = &packet_out_meta_to_type_uri;
+      meta_to_bit_width = &packet_out_meta_to_bit_width;
+    } else {
+      return MAKE_ERROR(ERR_UNIMPLEMENTED)
+             << "Undupported controller header" << ctrl_hdr_name;
```
pudelkoM, line 156:
```suggestion
      return MAKE_ERROR(ERR_UNIMPLEMENTED)
             << "Unsupported controller header " << ctrl_hdr_name;
```

stratum/hal/lib/barefoot/bfrt_p4runtime_translator.cc
```diff
@@ -0,0 +1,695 @@
...
+      if (to_sdk) {
+        from_bit_width = gtl::FindWithDefault(
+            table_to_field_to_bit_width_[table_id], field_id, 0);
+        to_bit_width = gtl::FindWithDefault(kUriToBitWidth, *uri, 0);
+      } else {
+        from_bit_width = gtl::FindWithDefault(kUriToBitWidth, *uri, 0);
+        to_bit_width = gtl::FindWithDefault(
+            table_to_field_to_bit_width_[table_id], field_id, 0);
+      }
+      if (from_bit_width && to_bit_width) {
```
pudelkoM, line 255: Could be inverted to `if not continue`

stratum/hal/lib/barefoot/bfrt_p4runtime_translator.cc
```diff
@@ -0,0 +1,695 @@
...
+          default:
+            return MAKE_ERROR(ERR_UNIMPLEMENTED)
+                   << "Unsupported field match type: "
+                   << field_match.ShortDebugString();
+        }
+      }  // else, we don't modify the value if it doesn't need to be translated.
+    }
+  }
+
+  if (translated_entry.action().type_case() == ::p4::v1::TableAction::kAction) {
```
pudelkoM, line 318: switch?

stratum/hal/lib/barefoot/bfrt_packetio_manager.cc
```diff
@@ -290,13 +297,15 @@ ::util::Status BfrtPacketioManager::ParsePacketIn(const std::string& buffer,
 
 ::util::Status BfrtPacketioManager::TransmitPacket(
     const ::p4::v1::PacketOut& packet) {
+  ASSIGN_OR_RETURN(const auto& translated_packet_out,
+                   bfrt_p4runtime_translator_->TranslatePacketOut(packet));
```
pudelkoM, line 301: Move after init check

pudelkoM (approved):

PR #904, opened 2022-01-15
[SDFAB-919] Support P4Runtime translation for action profile member

Yi-Tseng:
Implement `TranslateActionProfileMember` in the P4runtimeBfrtTranslator.
Also, add new test cases to improve the coverage.
And add new test cases for TableEntry to translate table entry with `action_profile_action_set` type.


pudelkoM (approved):
Looks good. We agreed on merging all feature PRs first, then into main.

PR #905, opened 2022-01-19
[SDFAB-920] Upport P4Runtime translation for PRE

Yi-Tseng:
Add support P4Runtime translation for PRE entries.
Note that the translation logic doesn't depend on the P4Info since we we know we are always translating the port numbers.


pudelkoM (approved):
Looks good.

PR #907, opened 2022-01-21
[SDFAB-942] Support P4Runtime translation - PacketIO

Yi-Tseng:
Add P4Runtime translation support for packet-io

Also, include the following changes:

- Preprocess P4Info before pushing it to other BfRt managers, since managers need to use the original bit width value from the P4 source code. (also, add a new test for this)
- Move `P4RuntimeBfrtTranslationWritterWraper` to `P4RuntimeBfrtTranslator` class and rename to `ReadResponseWriterWrapper`
- Add `StreamMessageResponseWriterWrapper`

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_node.cc
+++ b/stratum/hal/lib/barefoot/bfrt_node.cc
@@ -287,7 +287,7 @@ std::unique_ptr<BfrtNode> BfrtNode::CreateInstance(
   }
   ASSIGN_OR_RETURN(const auto& request,
                    p4runtime_bfrt_translator_->TranslateReadRequest(req));
-  P4RuntimeBfrtTranslationWriterWrapper writer_wrapper(
+  P4RuntimeBfrtTranslator::ReadResponseWriterWrapper writer_wrapper(
       writer, p4runtime_bfrt_translator_);
   writer = &writer_wrapper;
   ::p4::v1::ReadResponse resp;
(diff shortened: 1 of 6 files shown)


stratum/hal/lib/barefoot/bfrt_node.cc
```diff
@@ -138,6 +138,14 @@ ::util::Status BfrtNode::CommitForwardingPipelineConfig() {
   RETURN_IF_ERROR(bf_sde_interface_->AddDevice(device_id_, bfrt_config_));
 
   // Push pipeline config to the managers.
+  RETURN_IF_ERROR(p4runtime_bfrt_translator_->PushForwardingPipelineConfig(
+      bfrt_config_.programs(0).p4info()));
+
+  // Augment the P4Info so managers will use the original bitwith from P4 code
+  // for every fields (e.g, 9-bit port number instead of 32-bits).
+  ASSIGN_OR_RETURN(const auto& low_level_p4info,
+                   p4runtime_bfrt_translator_->GetLowLevelP4Info());
```
Yi-Tseng, line 147: Not sure how to name this. Basically, it is a P4Info with original bit width for match field, packet metadata, and action params.
The reason we need to pass this p4info is mostly that packet-io manager needs to use the original bit width to parse packet-io messages.

pudelkoM (approved):

PR #912, opened 2022-01-27
Support P4 Studio 9.7.1 & 9.8.0

Yi-Tseng:
Add support for Intel P4 Studio(aka SDE) version 9.7.1 and 9.8.0
This PR also deprecate old SDE 9.3.1 support and cleanup the PI_bf dependency

- [x] Find a better way to compare semantic version in bash
- [x] Test on HW switch
~Find a way to avoid including PI when building stratum_bfrt~ (Will remove PI after we deprecate 9.5.x)


pudelkoM (approved):
Lots of good cleanup here! Very much appreciated.
Only comment I have is, this could maybe be split up into multiple PRs once testing is finished? Remove 9.3.1 and cleanup PI. Then add 9.7.1 and 9.8.0. Maybe the numeric version as well.

Yi-Tseng:
> Lots of good cleanup here! Very much appreciated. Only comment I have is, this could maybe be split up into multiple PRs once testing is finished? Remove 9.3.1 and cleanup PI. Then add 9.7.1 and 9.8.0. Maybe the numeric version as well.

will do, thanks

stratum/hal/bin/barefoot/build-bf-sde.sh
```diff
@@ -150,7 +158,8 @@ else
     echo "SDE version: ${SDE_VERSION}"
 fi
 
-if [[ $SDE_VERSION == "9.7.0" ]]; then
+if [[ ! $(numeric_version "$SDE_VERSION") < $(numeric_version "9.7.0") ]]; then
```
pudelkoM, line 161:
```suggestion
if [[ $(numeric_version "$SDE_VERSION") -ge $(numeric_version "9.7.0") ]]; then
```

https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html

pudelkoM (approved):

PR #915, opened 2022-02-01
Support P4Studio 9.5.2

Yi-Tseng:
Add P4Studio 9.5.2 support


stratum/hal/lib/barefoot/bf_sde_wrapper.cc
```diff
@@ -1648,7 +1648,7 @@ std::string BfSdeWrapper::GetBfChipType(int device) const {
 std::string BfSdeWrapper::GetSdeVersion() const {
 #if defined(SDE_9_3_1)
   return "9.3.1";
-#elif defined(SDE_9_5_0)
+#elif defined(SDE_9_5_0) || defined(SDE_9_5_2)
   return "9.5.0";
```
pudelkoM, line 1652: I don't think that's smart. We should return the correct string.

pudelkoM:
Update the `stratum/tools/release-stratum.sh` script as well?

stratum/hal/lib/barefoot/bf_sde_wrapper.cc
Yi-Tseng: oops.
I use "search and replace" tool to change this file.
Let me fix that...

bocon13:
Should we just replace 9.5.0 with 9.5.2? Or would you prefer to remove 9.5.0 in a future commit?

Yi-Tseng:
> Should we just replace 9.5.0 with 9.5.2? Or would you prefer to remove 9.5.0 in a future commit?

Maybe we can remove this in the next release? I can add a comment in the README

pudelkoM (approved):

PR #919, opened 2022-02-03
Make `BfrtP4RuntimeTranslator` canonical bytestring aware

pudelkoM:
The `BfrtP4RuntimeTranslator` should respect the feature flag (`incompatible_enable_bfrt_legacy_bytestring_responses`) and return canonical byte strings after translation, if requested.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_p4runtime_translator.cc
+++ b/stratum/hal/lib/barefoot/bfrt_p4runtime_translator.cc
@@ -14,6 +14,8 @@
 #include "stratum/lib/utils.h"
 #include "stratum/public/proto/error.pb.h"
 
+DECLARE_bool(incompatible_enable_bfrt_legacy_bytestring_responses);
+
 namespace stratum {
 namespace hal {
 namespace barefoot {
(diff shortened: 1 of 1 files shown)


stratum/hal/lib/barefoot/bfrt_p4runtime_translator.cc
```diff
@@ -683,6 +685,9 @@ ::util::StatusOr<std::string> BfrtP4RuntimeTranslator::TranslateTnaPortId(
     const uint32 sdk_port_id = singleton_port_to_sdk_port_[port_id];
     std::string sdk_port_id_bytes = P4RuntimeByteStringToPaddedByteString(
         Uint32ToByteStream(sdk_port_id), NumBitsToNumBytes(bit_width));
+    if (!FLAGS_incompatible_enable_bfrt_legacy_bytestring_responses) {
+      sdk_port_id_bytes = ByteStringToP4RuntimeByteString(sdk_port_id_bytes);
```
Yi-Tseng, line 689: just want to confirm, we also want to make the byte string canonical when writing to the SDK, right?

pudelkoM: Probably not. The Wrapper will just pad it again. Good catch.

PR #929, opened 2022-02-23
Handle errors without details from switch interface on P4 writes and reads

pudelkoM:
When no pipeline has been pushed, switch interface implementations will return `ERR_NOT_INITIALIZED` on P4 reads and writes. This trips up the request logging logic in `P4Service` which currently expects one status for every p4 entity.

Implementations are not required to, and many currently don't, actually populate the results vector:
https://github.com/stratum/stratum/blob/e13ca1a4914429f9729946f7a741121a5fb4a298/stratum/hal/lib/common/switch_interface.h#L129-L132

The only guarantee given is that `results` will either be empty or 1 for 1.

I also don't find it sensible to change this and fix all implementations. Most implementations perform pre-checks that would need to be significantly more complicated.

https://github.com/stratum/stratum/blob/e13ca1a4914429f9729946f7a741121a5fb4a298/stratum/hal/lib/barefoot/bfrt_node.cc#L189-L197

Instead, I'd add a check in `LogWriteRequest`/`LogReadRequest` to cover the empty case and skip logging in those cases. Alternatively we log the overall error, i.e. `ERR_NOT_INITIALIZED`.

Example log:
```
E20220221 11:30:53.749909    50 bfrt_node.cc:189] StratumErrorSpace::ERR_NOT_INITIALIZED: Not initialized!
E20220221 11:30:53.750008    50 p4_service.cc:316] Failed to write forwarding entries to node 1: Not initialized!
E20220221 11:30:53.750037    50 p4_service.cc:219] Size mismatch: 0 != 1. Did not log anything!
```

### DIFF
--- a/stratum/hal/lib/common/p4_service.cc
+++ b/stratum/hal/lib/common/p4_service.cc
@@ -210,12 +210,13 @@ namespace {
 
 // Helper to facilitate logging the write requests to the desired log file.
 void LogWriteRequest(uint64 node_id, const ::p4::v1::WriteRequest& req,
+                     const ::util::Status& status,
                      const std::vector<::util::Status>& results,
                      const absl::Time timestamp) {
   if (FLAGS_write_req_log_file.empty()) {
     return;
   }
-  if (results.size() != req.updates_size()) {
+  if (!results.empty() && results.size() != req.updates_size()) {
     LOG(ERROR) << "Size mismatch: " << results.size()
                << " != " << req.updates_size() << ". Did not log anything!";
     return;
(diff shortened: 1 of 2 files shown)


stratum/hal/lib/common/p4_service.cc
```diff
@@ -215,7 +215,9 @@ void LogWriteRequest(uint64 node_id, const ::p4::v1::WriteRequest& req,
   if (FLAGS_write_req_log_file.empty()) {
     return;
   }
-  if (results.size() != req.updates_size()) {
+  if (results.empty()) {
+
+  } else if (results.size() != req.updates_size()) {
```
Yi-Tseng, line 220: How about
```suggestion
if (!results.empty() && results.size() != req.updates_size()) {
```

pudelkoM: That was just a placeholder. I have not thought about the solution yet, I guess yours could work.

stratum/hal/lib/common/p4_service.cc
```diff
@@ -210,57 +210,61 @@ ::grpc::Status ToGrpcStatus(const ::util::Status& status,
...
       absl::FormatTime("%Y-%m-%d %H:%M:%E6S", timestamp, absl::LocalTimeZone());
-  for (size_t i = 0; i < results.size(); ++i) {
-    absl::StrAppend(&msg, ts, ";", node_id, ";",
-                    req.updates(i).ShortDebugString(), ";",
-                    results[i].error_message(), "\n");
+  for (size_t i = 0; i < req.updates_size(); ++i) {
+    absl::StrAppend(
+        &msg, ts, ";", node_id, ";", req.updates(i).ShortDebugString(), ";",
+        results.empty() ? status.error_message() : results[i].error_message(),
+        "\n");
```
pudelkoM, line 231: If this change is to complicated, I'd propose to drop logging of write errors without details completely. Even today certain requests with errors (auth, valid device ID, ...) are not written to the log.

Yi-Tseng:
I like the current/simple solution.
We don't have to log the error for each entry since the error is not really related to the entry itself.
And it is clear enough for me(and users I think) to figure out why the controller cannot write forwarding entries

stratum/hal/lib/common/p4_service_test.cc
```diff
@@ -745,6 +745,35 @@ TEST_P(P4ServiceTest, WriteFailureForAuthError) {
...
+          DoAll(SetArgPointee<1>(kExpectedResults),
+                Return(::util::Status(StratumErrorSpace(), ERR_NOT_INITIALIZED,
+                                      kAggrErrorMsg))));
+
+  // Invoke the RPC and validate the results.
+  ::grpc::Status status = stub_->Write(&context, req, &resp);
+  EXPECT_FALSE(status.ok());
+  EXPECT_EQ(::grpc::StatusCode::FAILED_PRECONDITION, status.error_code());
+  EXPECT_THAT(status.error_message(), HasSubstr(kAggrErrorMsg));
+  // TODO(max): P4Runtime spec says error_details should be empty in not flow
```
Yi-Tseng, line 772: `in` -> `if`?

Yi-Tseng (approved):

pudelkoM:
> I like the current/simple solution. We don't have to log the error for each entry since the error is not really related to the entry itself. And it is clear enough for me(and users I think) to figure out why the controller cannot write forwarding entries

Works for me as well. At minimum the write log MUST contain all successfully installed entries. File-logging rejected entries is nice, but optional. They get still get logged normal and the gRPC code is failed as well.

PR #930, opened 2022-02-24
Allow zero burst sizes in P4 `MeterConfig`

pudelkoM:
In Aether we have a use case for setting the meter burst sizes to 0. While the various meter _rates_ can already be 0, effectively implementing a packet drop, the initial bucket is still filled with `burst_size` bytes, potentially marking a few packets in an undesirable way.
This change deviates from the trTCM spec by allowing zero burst sizes in P4 `MeterConfig`s.


pudelkoM:
@pierventre

pierventre (approved):

PR #931, opened 2022-03-01
Only log P4RT arbitration events if mastership changes

pudelkoM:
In Aether ONOS re-asserts mastership every minute. We got a complaint that these mastership log messages pollute the log file. This change reduces the amount of messages by only logging if the mastership has actually changed. Thus keeping the log useful, while reducing the amount.


pudelkoM:
One (of many) problems of this change is that it prevents logging on ANY backup mastership updates, as `is_master != was_master` is never true for backups that don't transition. Seems like we need something more elaborate than this.

pudelkoM:
Also consider solving this with #781

PR #935, opened 2022-03-19
Update Bazel to 4.2.2

pudelkoM:
This change updates Bazel to the latest 4.x.x release, which is 4.2.2


pudelkoM:
I faintly remember encountering issues when first trying this some months ago...
Maybe those where related to Python rules/deps that we removed by now.

PR #938, opened 2022-03-21
Do not expect or emit padded byte strings in BFRT P4Runtime translator

Yi-Tseng:
- No need to add padding zeros when writing through the SDE wrapper since it will help add zeros before writing to the SDE
 - Values from SDE wrapper for port number can be 1 or 2 bytes when we set `incompatible_enable_bfrt_legacy_bytestring_responses` to false

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_p4runtime_translator.cc
+++ b/stratum/hal/lib/barefoot/bfrt_p4runtime_translator.cc
@@ -678,8 +678,7 @@ BfrtP4RuntimeTranslator::TranslateP4Info(
     // id(uint32) -> sdk port id(1 or 2 bytes)
     const uint32 port_id = ByteStreamToUint<uint32>(value);
     RET_CHECK(singleton_port_to_sdk_port_.count(port_id));
-    const uint32 sdk_port_id = singleton_port_to_sdk_port_[port_id];
-    return Uint32ToByteStream(sdk_port_id);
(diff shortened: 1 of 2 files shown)


pudelkoM:
low level driver being the SDE wrapper here? Or does the SDE accept truncated strings now?

Yi-Tseng:
> low level driver being the SDE wrapper here? Or does the SDE accept truncated strings now?

I mean SDE accepts truncated byte strings
I tested with SDE 9.7.0 and didn't see issues.

pudelkoM:
Did you test the SDE with custom code? Because all translated entries still go through the wrapper, which does the padding. Either way, the wrapper will take care of it, no need to pad here.

Yi-Tseng:
> 



> Did you test the SDE with custom code? Because all translated entries still go through the wrapper, which does the padding. Either way, the wrapper will take care of it, no need to pad here.

I was wrong, I didn't realized that the wrapper adds padding zeros before sending to the SDE

pudelkoM:
> > 
> 
> > Did you test the SDE with custom code? Because all translated entries still go through the wrapper, which does the padding. Either way, the wrapper will take care of it, no need to pad here.
> 
> I was wrong, I didn't realized that the wrapper adds padding zeros before sending to the SDE

No worries, I wasn't sure if the SDE changed either.

pudelkoM (approved):

PR #942, opened 2022-03-23
Accept action IDs on `DirectCounterEntry` in Stratum-bfrt

pudelkoM:
Those are still valid and the action field has to be ignored, according to the P4RT spec.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager.cc
@@ -629,15 +629,11 @@ std::unique_ptr<BfrtTableManager> BfrtTableManager::CreateInstance(
                        direct_counter_entry, /*to_sdk=*/true));
   // Read table entry first.
   const auto& table_entry = translated_direct_counter_entry.table_entry();
-  RET_CHECK(table_entry.action().action().action_id() == 0)
-      << "Found action on DirectCounterEntry "
-      << translated_direct_counter_entry.ShortDebugString();
   ASSIGN_OR_RETURN(uint32 table_id,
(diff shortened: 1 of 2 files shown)


pudelkoM:
Thanks @daniele-moro for reporting.

pudelkoM:
cc @pier
We might have code in ONOS that was affected by this, too.

daniele-moro:
I think ONOS code is not affected but wait for confirmation from @pierventre. Direct counters in ONOS are either read as part of a read of table entries, or doing reads by specifying only the table entry handle (i.e., match + priority). See: https://github.com/opennetworkinglab/onos/blob/dc08c95eed882936edf47527ca64d416009e8514/drivers/p4runtime/src/main/java/org/onosproject/drivers/p4runtime/P4RuntimeFlowRuleProgrammable.java#L536-L551

pierventre:
> I think ONOS code is not affected but wait for confirmation from @pierventre. Direct counters in ONOS are either read as part of a read of table entries, or doing reads by specifying only the table entry handle (i.e., match + priority). See: https://github.com/opennetworkinglab/onos/blob/dc08c95eed882936edf47527ca64d416009e8514/drivers/p4runtime/src/main/java/org/onosproject/drivers/p4runtime/P4RuntimeFlowRuleProgrammable.java#L536-L551

Yup - we talked about this in private me and @pudelkoM and I confirm what you are saying. On our end we can remove some code around `encodeKey` for `TableEntry` but IMHO is not really worth.

pudelkoM:
Sounds good to me. "affected" was poorly worded here. "potential chance for simplification" is more appropriate as old behavior will continue to work.

Yi-Tseng (approved):

PR #944, opened 2022-03-24
Replace raw text proto string terminal `PROTO` with `pb`

pudelkoM:
Starting with clang-13 this is the enforced terminal. This change should make the eventual update easier.


pudelkoM:
Lint check is expected to fail, as some `_test.cc` files are not formatted yet.

PR #948, opened 2022-04-07
Do not provide `nullptr` ostream overload on MacOS anymore

pudelkoM:
This was needed until Mojave, but seems to be fixed starting with Monterey.


pudelkoM:
For context, this fix is required on Monterey on a M1 cpu, but with a x86_64 toolchain setup. Please provide more data points if possible.

Yi-Tseng (approved):

PR #968, opened 2022-05-14
Add table entry test in bfrt_table_manager_test

Dscano:
I added a test for inserting a entry in a table


pudelkoM (approved):
This looks good to me.
I hope the gtest/gmock invocations were not too confusing.

PR #978, opened 2022-06-04
Add `BfrtTableManager` delete and modify entry test

Dscano:
1) Added a global variable for a table entry
2) Fixed some part of the code of WriteTableEntryTest
3) Add test related to modification and deletion of table entry

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager_test.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager_test.cc
@@ -51,84 +51,41 @@ class BfrtTableManagerTest : public ::testing::Test {
     const std::string kSamplePipelineText = R"pb(
       programs {
         name: "test pipeline config",
-        p4info {
-          pkg_info {
-            arch: "tna"
-          }
-          tables {
-            preamble {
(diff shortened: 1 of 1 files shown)


stratum/hal/lib/barefoot/bfrt_table_manager_test.cc
```diff
@@ -144,6 +144,22 @@ class BfrtTableManagerTest : public ::testing::Test {
...
+          mask: "\377\377"
+        }
+      }
+      action {
+        action {
+          action_id: 16783057
+        }
+      }
+      priority: 10
+)pb";
```
pudelkoM, line 162: We usually aim for the smallest possible scope when creating constants. In tests this would be the test class (`BfrtTableManagerTest`), as `static constexpr` member variables. Even though those variables are const and could be optimized away in most cases, they still need backing storage for uses involving a pointer. Hence we have code like `constexpr int BfrtTableManagerTest::kDevice1;` which creates the storage.

pudelkoM (commented):
LGTM, see minor comment

pudelkoM (approved):

PR #983, opened 2022-06-16
Add `BfrtTableManager` write table entry reject test 

Dscano:
1) Fixed  action id `kTableEntryText`
2) Added constant table in piepeline 
3) Various cases reject

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_table_manager_test.cc
+++ b/stratum/hal/lib/barefoot/bfrt_table_manager_test.cc
@@ -51,84 +51,41 @@ class BfrtTableManagerTest : public ::testing::Test {
     const std::string kSamplePipelineText = R"pb(
       programs {
         name: "test pipeline config",
-        p4info {
-          pkg_info {
-            arch: "tna"
-          }
-          tables {
-            preamble {
-              id: 33583783
-              name: "Ingress.control.table1"
-            }
-            match_fields {
-              id: 1
-              name: "field1"
-              bitwidth: 9
-              match_type: EXACT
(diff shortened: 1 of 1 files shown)


Dscano:
Format textproto is needed

stratum/hal/lib/barefoot/bfrt_table_manager_test.cc
```diff
@@ -51,84 +51,53 @@ class BfrtTableManagerTest : public ::testing::Test {
...
-          }
+        p4info {pkg_info {arch: "tna"} tables {
+          preamble {id: 33583783 name: "Ingress.control.table1"}
+          match_fields {id: 1 name: "field1" bitwidth: 9 match_type: EXACT}
+          match_fields {id: 2 name: "field2" bitwidth: 12 match_type: TERNARY}
+          match_fields {id: 3 name: "field3" bitwidth: 15 match_type: RANGE}
+          action_refs {id: 16794911}
+          const_default_action_id: 16836487
+          direct_resource_ids: 318814845
+          size: 1024
```
pudelkoM, line 62: Can you revert the textproto formatting?

clang-format should have not picked that up, as it's disabled in the config: https://github.com/stratum/stratum/blob/cb10d8501bd13b66b4c41f8a3d8139747d228765/.clang-format#L8-L10

Having to do it manually is unfortunate, but the autoformat produces suboptimal results.

Dscano: Sure, no problem.

PR #992, opened 2022-06-29
Add `BfrtTableManager` Tests

Dscano:
1) Fixed action id `kTableEntryText`
2) Added constant table and register in piepeline config
3) Various cases of reject register, meter, action profile member, table entry
4) Various cases of write register, meter, action profile member, table entry


Dscano:
fixed style

Issue #982, opened 2022-06-13
Connect ONOS with Tofino using pipeconf fabric-tna


tisasi001:
Hello,
I connected the ONOS controller using fabric-tna pipeconf to a Tofino switch and it seems to be well connected, as I get a green check in the ONOS GUI. Now that I have ONOS and the switch connected, I would like to try by sending packages from port to port in the switch. However, I don't really know how to do it, so I would appreciate if you answer some of my questions.
In my Tofino switch with stratum which p4 file do I have to change in order to change the data plane? Is it stratum/stratum/pipelines/main.p4?
In the ONOS controller what files do I have to change then?

Dscano:
Hi tisasi,
1) you have to connect a packet generator to the switch, e.i a device, and send traffic.
2)Once you "discover" your tofino switch with netcfg json file you load the pipeline written in the file in the switch, in other words the fabric-tna pipeline. You have also the app installed in ONOS capable to setting the flow rules in the pipeline.
3)https://opennetworking.org/2022-p4-workshop-gated/ here you can find 3  video tutorial for undestanding how you can set flow rules etc in the switch. The tutorial is made for the v1model but the approach is the same for the tofino.

Dscano:
Hi tisasi,

you should load the proper chassis configuration file https://github.com/stratum/stratum/tree/main/stratum/hal/config. Basically when you run stratum it enables all the port defined in the json file.

Best
Davide

tisasi001:
Hello,
In my chassis configuration (/etc/stratum/x86-64-bfrt2556x-1t-r0/chassis_config.pb.txt) I have all the single ports as Enabled:

![image](https://user-images.githubusercontent.com/64487317/173539358-1b48bb8d-866c-47c1-8c0d-df3dbc8d73dd.png)

When running stratum, it appears that it has been enabled:

![image](https://user-images.githubusercontent.com/64487317/173539522-151a60e9-abe7-4c0d-8c46-347ec0aac53f.png)

But then only these 4 ports are actually UP:
![image](https://user-images.githubusercontent.com/64487317/173539659-5791fca3-f6db-4bf8-b4d4-07f3c1f57ae6.png)

Dscano:
You can try the "bfrt commands" related to de ports direcly in the shell where you are running stratum and you can see the status of the ports there.

pudelkoM:
[...]

PR #994, opened 2022-06-30
Extend error messages in `BfrtP4RuntimeTranslator`

pudelkoM:
This PR adds more detailed error messages to `BfrtP4RuntimeTranslator` on translation failure.


pudelkoM:
As seen #982, translation errors are not very descriptive right now. This will not fix any of them, but might help in addressing configuration errors.

PR #995, opened 2022-06-30
Add all `//stratum/hal/lib/barefoot` test targets to CI

pudelkoM:


pudelkoM:
Some targets (non-test ones, weirdly) need the SDE.

PR #996, opened 2022-07-01
First stab at P4RT role config

pudelkoM:


stratum/lib/p4runtime/p4_role_config.proto
```diff
@@ -0,0 +1,44 @@
+// Copyright 2022-present Open Networking Foundation
```
pudelkoM, line 1: Should this file live under `stratum/stratum/public/proto/`?

pierventre:
cc @pierventre

stratum/public/proto/p4_role_config.proto
```diff
@@ -0,0 +1,44 @@
...
+
+import "p4/v1/p4runtime.proto";
+
+
+// The P4RoleConfig message conists of these fields:
+//  accessible_p4_ids - a list of the
+//  packet_in_filters -
+//  receives_packet_ins -
+message P4RoleConfig {
+  repeated uint32 accessible_p4_ids = 1;
```
pierventre, line 22: shall we split in readable/writable ?

stratum/public/proto/p4_role_config.proto
```diff
@@ -0,0 +1,44 @@
...
+  repeated p4.v1.PacketMetadata packet_in_filters = 2;
```
pierventre, line 23: do we need to specify somewhere the allowed values for `PacketMetadata` ?

pudelkoM: Are you thinking in terms of checking the `metadata_id` provided for validity against the p4info? Could be tricky sometimes as role config can happen before pipeline push.

pierventre: Initially, I was thinking having a realm of admissible values to make sure the values used here are "certified". BUT it is not really necessary. Moreover, there are many components that might want to use this - not only sd-fabric.

Probably, the most important thing to do in future is the exclusivity of the accesses to the resources (even though we might need to revisit in future but for now we should not have overlapping roles)

stratum/public/proto/p4_role_config.proto
pudelkoM: We agreed to split by "mutually exclusive" and "shared" attribute.

pudelkoM:
Superseded by other PRs.

PR #1000, opened 2022-07-11
Return `NOT_FOUND` for missing node IDs

yi-tseng-intel:
Also fix implementations for `X_switch.cc` to use `ERR_ENTRY_NOT_FOUND` code when a node id is missing.

### DIFF
--- a/stratum/hal/lib/barefoot/bfrt_switch.cc
+++ b/stratum/hal/lib/barefoot/bfrt_switch.cc
@@ -305,7 +305,7 @@ std::unique_ptr<BfrtSwitch> BfrtSwitch::CreateInstance(
       BfrtNode* bfrt_node =
           gtl::FindPtrOrNull(device_id_to_bfrt_node_, device_id);
       if (bfrt_node == nullptr) {
-        ::util::Status error = MAKE_ERROR(ERR_INVALID_PARAM)
+        ::util::Status error = MAKE_ERROR(ERR_NOT_FOUND)
                                << "Node ID " << node_id
                                << " mapped to unknown device " << device_id
                                << ".";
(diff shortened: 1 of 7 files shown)


pudelkoM (approved):

pudelkoM:
Not sure why CI is not getting triggered.

pudelkoM:
Looking at `error.proto` I found `ERR_ENTRY_NOT_FOUND`. This one is used in many places (e.g. channel) already and maps to `NOT_FOUND` in the end. Should we reuse that one instead of adding a new error?

yi-tseng-intel:
With bmv2:

```
python3 -m p4runtime_sh --grpc-addr 127.0.0.1:50001 --device-id 10 --election-id 1,2 --config ./p4src/v1model/build/fabric/bmv2/p4info.txt,./p4src/v1model/build/fabric/bmv2/bmv2.json
CRITICAL:root:StreamChannel error, closing stream
CRITICAL:root:P4Runtime RPC error (NOT_FOUND): Node with ID 10 is unknown or no config has been pushed to it yet.
CRITICAL:root:Failed to establish session with server
```

yi-tseng-intel:
> Looking at `error.proto` I found `ERR_ENTRY_NOT_FOUND`. This one is used in many places (e.g. channel) already and maps to `NOT_FOUND` in the end. Should we reuse that one instead of adding a new error?

The reason I create new instead of using exists one is that I thought the `ERR_ENTRY_NOT_FOUND` applies to p4runtime entry only. But maybe we can just reuse it

pudelkoM (approved):

pudelkoM:
> I thought the `ERR_ENTRY_NOT_FOUND` applies to p4runtime entry only.

I thought so too initially. Then I saw the existing usages of it and think it's fine.

PR #1009, opened 2022-07-16
Add `BfrtTableManager` test table action profile

Dscano:
Added tests related to log writing and rejection of action profile


Dscano:
waiting merge  #1008

PR #1012, opened 2022-07-20
Make `AuthPolicyChecker` buildable on non-Linux platforms

pudelkoM:
This change makes `AuthPolicyChecker` more portable by removing unused functions that were Linux specific.

### DIFF
--- a/stratum/lib/security/auth_policy_checker.cc
+++ b/stratum/lib/security/auth_policy_checker.cc
@@ -7,17 +7,6 @@
 #include <errno.h>
 #include <stdlib.h>
 
-// The current (stub) implementation of AuthPolicyChecker uses Linux specific
-// syscalls. To allow building on other platforms, we provide noop stubs for the
-// parts that depend on Linux.
-// TODO(max): The file watcher functions don't seem to be used anywhere.
-//            Investigate whether we can remove them.
-#ifdef __linux__
-#include <sys/epoll.h>
-#include <sys/inotify.h>
-#include <sys/types.h>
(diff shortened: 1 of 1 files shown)


stratum/lib/security/auth_policy_checker.cc
```diff
@@ -59,6 +67,7 @@ void ReadProtoIfValidFileExists(const std::string& path,
   }
 }
 
+#ifdef __linux__
```
pudelkoM, line 70: Platform `ifdefs` in implementation code are not great. On the other hand, this class is mostly a stub and those functions are unused.

stratum/lib/security/auth_policy_checker.cc
```diff
@@ -6,9 +6,17 @@
 
 #include <errno.h>
 #include <stdlib.h>
+
+// The current (stub) implementation of AuthPolicyChecker uses Linux specific
+// syscalls. To allow building on other platforms, we provide noop stubs for the
+// parts that depend on Linux.
+// TODO(max): The file watcher functions don't seem to be used anywhere.
```
bocon13, line 13: I'm guessing that these were originally added for the `Authorize()` and `RefreshPolicies()` methods below. 

I'm okay just removing them entirely, then when someone wants to implement the files below we have them introduce something more platform independent. For example, https://github.com/SpartanJ/efsw

pudelkoM: Seems reasonable to me as well.

PR #932, opened 2022-03-08
Use SONiC as Stratum base OS

Yi-Tseng:
TODO:
 - [x] Build stratum-bfrt image and test on HW
   - [x] native
   - [x] docker
 - [x] Build stratum-bcm image and test on HW
   - [x] OpenNSA
     - [x] native
     - [x] docker
   - [x] SDKLT
     - [x] native
     - [x] docker
 - [x]  Build mn-stratum and test
 - [x] Update docs
   - [x] stratum/hal/bin/bcm/standalone/README.md
   - [x] stratum/hal/bin/barefoot/README.run.md
   - [x] stratum/hal/bin/barefoot/README.build.md
   - [x] .github/ISSUE_TEMPLATE/bug_report.md

PR #1021, opened 2022-07-31
Add curly brackets `procmon_test`

Dscano:
After a pull from the main branch or download of the docker stratum dev environment  sometimes your test fails due to an issue in `procmon_test`.
Screenshot from 2022-07-31 22-19-11
Screenshot from 2022-07-31 22-19-23
Screenshot from 2022-07-31 22-30-14


pudelkoM:
Good catch. GCC got updated in #932. I guess new flags are on by default now.

What's the actual patch here? Diff still shows up empty.

Dscano:
Don't know why was empty, now should be ok

ffoulkes (approved):
LGTM

pudelkoM (approved):

PR #1033, opened 2022-08-09
Implement P4RT role config write request filtering

pudelkoM:
This PR implements P4RT role config write permissions filtering. Clients can use the list of IDs in the role config to limit their write access to certain P4 entities.


pudelkoM:
Still need to think about wild card reads and the ID filtering.

PR #1036, opened 2022-08-11
Implement role config pipeline push permissions

pudelkoM:
This PR implements the ability to prevent roles from pushing a pipeline.

### DIFF
--- a/stratum/lib/p4runtime/sdn_controller_manager.cc
+++ b/stratum/lib/p4runtime/sdn_controller_manager.cc
@@ -70,6 +70,23 @@ grpc::Status VerifyElectionIdIsActive(
                       "Election ID is not active for the role.");
 }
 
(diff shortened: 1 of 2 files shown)


stratum/lib/p4runtime/sdn_controller_manager.cc
```diff
@@ -366,6 +383,15 @@ grpc::Status SdnControllerManager::AllowRequest(
     role_name = request.role();
   }
 
+  {
+    absl::MutexLock l(&lock_);
+    grpc::Status status =
+        VerifyRoleCanPushPipeline(role_name, role_config_by_name_);
```
pudelkoM, line 389: Make this function a private member func? Would make the locking less ugly.

pudelkoM: After some testing, this did not improve things significantly.

PR #1042, opened 2022-08-17
Remove unneeded dep on role config rule and make publicly visible

pudelkoM:
With these fixes importing and using the role config in external projects becomes possible.

- [x] Add code snippets to readme?


pudelkoM:
Example: https://github.com/pudelkoM/stratum-role-config-example

PR #1045, opened 2022-09-08
Decrease `num_controller_connections_` even if there is no controller manager

pudelkoM:
This change fixes a bug where the `num_controller_connections_` counter in `P4Service` is not decremented correctly when there is no controller manager. This can happen after startup when the first P4RT stream connection setup attempt fails due to, e.g., an invalid election ID. The tests are amended to now check for the correctness of the counter.


pudelkoM:
cc @pierventre

pierventre (approved):

PR #1047, opened 2022-09-13
Include Credo firmware in Stratum-bfrt package for Wedge100bf-65x

yi-tseng-intel:
Wedge100BF-65x requires "credo_firmware.bin" to set up the retimer hardware when using the BSP. Startup will failed if this file does not exists.

### DIFF
--- a/bazel/external/bfsde.BUILD
+++ b/bazel/external/bfsde.BUILD
@@ -106,6 +106,14 @@ pkg_tar(
     strip_prefix = "barefoot-bin",
 )
 
+pkg_tar(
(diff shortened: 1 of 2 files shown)


pudelkoM (approved):

ramrattle:
Hi, Im using sde 9.10.0 and build is failing with this patch, when this change was removed it went fine, can you please help me with this @yi-tseng-intel ?
https://github.com/stratum/stratum/issues/1078

pudelkoM:
I think you're right. The rule should have a wildcard glob to still match on non-BSP builds.

PR #1055, opened 2022-10-08
Return more detailed errors to P4Runtime clients

pudelkoM:
With this change the P4Service passes errors from the lower layers upwards, instead of creating them itself. This allows for more rich error messages to be returned to clients.


pudelkoM:
CC @pierventre

PR #1061, opened 2022-10-18
Canonicalize P4RT role config PacketIn filter byte strings

pudelkoM:
This change enforces canonicalization on incoming role config PacketIn filter strings to prevent false negative matches when comparing to canonical PacketIns later. Additionally, the filter, if set, cannot be empty.

The PacketIn filtering feature is incompatible with the `--incompatible_enable_bfrt_legacy_bytestring_responses` flag, as canonicalization is always performed in the P4Service, but only selectively in the `BfrtPacketIoManager`, hence leading to false negatives.


pudelkoM:
CC @pierventre

pierventre:
Thanks @pudelkoM, I believe we should have a discussion with @ccascone around this. I am not fully convinced tbh, but maybe I am simply wrong.

pierventre (dismissed):
Please revert the changes for the custom image. Other than that good to go for me.
Do you plan to remove the incompatible flag in this pr ?

pudelkoM:
> Please revert the changes for the custom image. Other than that good to go for me. Do you plan to remove the incompatible flag in this pr ?

Thanks for confirming! No the flag would get removed on a follow-up.

PR #1069, opened 2022-11-28
Expose a virtual CPU interface with Stratum-bfrt on Tofino

pudelkoM:
This PR introduces an additional TAP interface on Tofino. Via software
processing and a special constant in PacketIn headers (`0xbf01`), it is connected to the regular
PCIe CPU interface and can be used to steer packets from the data plane
to regular apps on the host. This implements a workaround for the broken
Ethernet CPU ports on certain Edgecore Wedge switches.


yi-tseng-intel:
LGTM, I think I can give it a try with our internal project :)
We might want to add some document to tell people to include special bytes to packet-in message when enabling TAP interface.

PR #1090, opened 2023-02-02
Allow empty Digest configuration for DELETE operation

gcgirish:
This PR supports empty Digest config for DELETE operation. For DELETE operation all that is needed is the DigestId.


pierventre (dismissed):
do we need to update PI ?

gcgirish:
> do we need to update PI ?

This is my first commit to Stratum project. What is PI?

pierventre:
> > do we need to update PI ?
> 
> This is my first commit to Stratum project. What is PI?

hal/pi for bmv2 but I am not fully sure; perhaps a question for @pudelkoM

pudelkoM:
> do we need to update PI ?

depends on if PI already handles deletes correctly. Worth checking, but what we observed right now is BFRT only.

gcgirish:
> > do we need to update PI ?
> 
> depends on if PI already handles deletes correctly. Worth checking, but what we observed right now is BFRT only.

I cant find reference to Digest handling for BMV2 in stratum. Yes, we have seen this issue for BFRT only.

pierventre:
> > > do we need to update PI ?
> > 
> > 
> > depends on if PI already handles deletes correctly. Worth checking, but what we observed right now is BFRT only.
> 
> I cant find reference to Digest handling for BMV2 in stratum. Yes, we have seen this issue for BFRT only.

I am confused - dont we support digests on stratum-bmv2 ? if we supprt, are we not doing DELETE in the tests ? Or is it a lucky case because the digest config is gone every time we do sw tests ?

stratum/hal/lib/barefoot/bfrt_table_manager_test.cc
```diff
@@ -578,11 +597,6 @@ TEST_F(BfrtTableManagerTest, DeleteDigestEntrySuccess) {
 
   const std::string kDigestEntryText = R"pb(
     digest_id: 401732455
-    config {
```
pierventre, line 581: shall we maintain this test and verify that brft_table_manager does not care? It seems to me another test we get for free.

pierventre (dismissed):

pudelkoM:
> > > > do we need to update PI ?
> > > 
> > > 
> > > depends on if PI already handles deletes correctly. Worth checking, but what we observed right now is BFRT only.
> > 
> > 
> > I cant find reference to Digest handling for BMV2 in stratum. Yes, we have seen this issue for BFRT only.
> 
> I am confused - dont we support digests on stratum-bmv2 ? if we supprt, are we not doing DELETE in the tests ? Or is it a lucky case because the digest config is gone every time we do sw tests ?

We do, but Stratum-bmv2 to PI is nearly completely pass-through from the Stratum view. We don't even look into the P4 entities before handing them over. If you're referring to the unit tests modified here, those are for bfrt only.

pierventre:
> > > > > do we need to update PI ?
> > > > 
> > > > 
> > > > depends on if PI already handles deletes correctly. Worth checking, but what we observed right now is BFRT only.
> > > 
> > > 
> > > I cant find reference to Digest handling for BMV2 in stratum. Yes, we have seen this issue for BFRT only.
> > 
> > 
> > I am confused - dont we support digests on stratum-bmv2 ? if we supprt, are we not doing DELETE in the tests ? Or is it a lucky case because the digest config is gone every time we do sw tests ?
> 
> We do, but Stratum-bmv2 to PI is nearly completely pass-through from the Stratum view. We don't even look into the P4 entities before handing them over. If you're referring to the unit tests modified here, those are for bfrt only.

No, I am referring to the recovery-nat tests that are failing due to this and I am wondering why the same tests on software are not

gcgirish:
> > > > > > do we need to update PI ?
> > > > > 
> > > > > 
> > > > > depends on if PI already handles deletes correctly. Worth checking, but what we observed right now is BFRT only.
> > > > 
> > > > 
> > > > I cant find reference to Digest handling for BMV2 in stratum. Yes, we have seen this issue for BFRT only.
> > > 
> > > 
> > > I am confused - dont we support digests on stratum-bmv2 ? if we supprt, are we not doing DELETE in the tests ? Or is it a lucky case because the digest config is gone every time we do sw tests ?
> > 
> > 
> > We do, but Stratum-bmv2 to PI is nearly completely pass-through from the Stratum view. We don't even look into the P4 entities before handing them over. If you're referring to the unit tests modified here, those are for bfrt only.
> 
> No, I am referring to the recovery-nat tests that are failing due to this and I am wondering why the same tests on software are not

I think bmv2 does not validate this and stratum is transparent to any requests going to bmv2.

pierventre (approved):

PR #1092, opened 2023-02-08
Make stratum_dpdk portable and bump p4-dpdk-target to latest version

osinstom:
This PR makes `stratum_dpdk` more portable and enables running it on VMs by using `nehalem` CPU instruction set.

### DIFF
--- a/bazel/external/dpdk.BUILD
+++ b/bazel/external/dpdk.BUILD
@@ -25,8 +25,7 @@ cc_library(
         "lib/libtdi_pna.so*",
         "lib/libtdi.so*",
         # DPDK libs
-        "lib/x86_64-linux-gnu/dpdk/pmds-22.2/*.so*",
-        "lib/x86_64-linux-gnu/librte_*.so*",
+        "lib/x86_64-linux-gnu/**/*.so*",
     ]),
     hdrs = glob([
         "include/bf_pal/*.h",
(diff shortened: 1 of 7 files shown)


bazel/external/dpdk.BUILD
```diff
@@ -25,8 +25,7 @@ cc_library(
         "lib/libtdi_pna.so*",
         "lib/libtdi.so*",
         # DPDK libs
-        "lib/x86_64-linux-gnu/dpdk/pmds-22.2/*.so*",
-        "lib/x86_64-linux-gnu/librte_*.so*",
+        "lib/x86_64-linux-gnu/**/*.so*",
```
osinstom, line 28: The latest DPDK uses `pmds-23.0` - this way we can be independent from the DPDK version.

stratum/hal/bin/tdi/dpdk/BUILD
```diff
@@ -125,6 +125,7 @@ pkg_deb(
         "libatomic1",
         "libpcap0.8",
         "libelf-dev",
+        "libjansson-dev",
```
osinstom, line 128: I found it needed for `stratum_dpdk --version`

pierventre (approved):

bazel/external/dpdk.BUILD
pudelkoM: We generally careful what we package, as the files here get installed under `/usr/lib` and `apt` does not allow conflicts. Could this below also fix it?

```suggestion
        "lib/x86_64-linux-gnu/pmds-/*.so*",
        "lib/x86_64-linux-gnu/librte_*.so*",
```

bazel/external/dpdk.BUILD
```diff
@@ -70,8 +69,7 @@ pkg_tar_with_symlinks(
...
```
pudelkoM, line 72: same here, see above

pudelkoM:
The failing checks are expected to fail either because of missing creds or outstanding code lints.

pudelkoM (approved):
lgtm

pierventre (approved):

PR #1095, opened 2023-02-13
Enable setting packet direction per port

osinstom:
This PR adds support to set per-port packet direction for the DPDK target. Under the hood, it sets `net_port_mask` that is used to populate direction registers in P4-DPDK: https://github.com/p4lang/p4-dpdk-target/blob/main/src/pipe_mgr/shared/dal/dpdk/dal_init.c#L159. 

Note that it puts restrictions on a chassis config - DPDK port indexes must always start from 0. Otherwise, mapping done by net_port_mask won't work.

### DIFF
--- a/stratum/hal/lib/common/common.proto
+++ b/stratum/hal/lib/common/common.proto
@@ -278,9 +278,9 @@ message PortConfigParams {
   }
   // DPDK packet direction types.
   enum DpdkPacketDirection {
-    DIRECTION_HOST = 0;
-    DIRECTION_NETWORK = 1;
-    DIRECTION_UNKNOWN = 2;
+    DIRECTION_UNKNOWN = 0;
+    DIRECTION_HOST = 1;
+    DIRECTION_NETWORK = 2;
   };
   // The configured admin state for this port.
   AdminState admin_state = 1;
--- a/stratum/hal/lib/common/common.proto
+++ b/stratum/hal/lib/common/common.proto
@@ -280,7 +280,7 @@ message PortConfigParams {
   enum PacketDirection {
     DIRECTION_HOST = 0;
     DIRECTION_NETWORK = 1;
-    DIRECTION_NONE = 2;
+    DIRECTION_UNKNOWN = 2;
   };
   // The configured admin state for this port.
   AdminState admin_state = 1;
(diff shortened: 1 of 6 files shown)


stratum/hal/lib/common/common.proto
```diff
@@ -276,6 +276,12 @@ message PortConfigParams {
     DEVICE_TYPE_VIRTIO_NET = 1;
     DEVICE_TYPE_VIRTIO_BLK = 2;
   }
+  // Packet Direction types
+  enum PacketDirection {
+    DIRECTION_HOST = 0;
+    DIRECTION_NETWORK = 1;
+    DIRECTION_NONE = 2;
```
daniele-moro, line 283: Do we care about having a `DIRECTION_NONE`, if we need it for default value, then we should put it a 1st element in the enum.

osinstom: `DIRECTION_NONE` can be useful for PSA. The order of fields corresponds to PNA values (RX=0, TX=1), but I'm fine with changing it.

ccascone, line 283: Nit: more self-documenting names, to avoid ambiguity on the actual direction
```suggestion
    DIRECTION_FROM_HOST = 0;
    DIRECTION_FROM_NETWORK = 1;
    DIRECTION_UNKNOWN = 2;
```

daniele-moro: OK, then let's keep it as is.

daniele-moro (approved):

stratum/hal/lib/tdi/dpdk/dpdk_sde_target.cc
```diff
@@ -132,7 +132,8 @@ ::util::Status TdiSdeWrapper::AddPort(int device, int port, uint64 speed_bps,
   port_attrs->port_dir = PM_PORT_DIR_DEFAULT;
   port_attrs->port_in_id = port;
   port_attrs->port_out_id = port;
-  port_attrs->net_port = config.packet_dir;
+  port_attrs->net_port =
+      config.packet_dir == hal::PortConfigParams::DIRECTION_NETWORK ? 1 : 0;
```
pudelkoM, line 136: If we do a conditional instead of a direct assignment here, we don't really need to ensure that `DIRECTION_NETWORK` has a specific value.

stratum/hal/lib/common/common.proto
pudelkoM: With the conditional assignment in place already, changing the order was really easy.

PR #1101, opened 2023-09-25
fix gnmi subscribe request sample_interval and heartbeat_interval unit bug

chujieyang:
According to [gnmi-specification](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md), STREAM subscription mode **sample_interval** param is unsigned 64-bit integer representing nanoseconds and **heartbeat_interval** is specified as an unsigned 64-bit integer in nanoseconds.

But in stratum, the logic directly apply the sample_interval and heartbeat_interval in nanoseconds unit to function Periodic which param is ms unit.

So, here need to be converted.

### DIFF
--- a/bazel/external/bfsde.BUILD
+++ b/bazel/external/bfsde.BUILD
@@ -2,9 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 load("@//bazel/rules:package_rule.bzl", "pkg_tar_with_symlinks")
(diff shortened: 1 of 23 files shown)


chujieyang:
@bocon13   why not the build_and_test license-check is failed，i did not change those files.

chujieyang:
@bocon13 please review this pr.

pudelkoM (approved):

Issue #1103, opened 2023-11-09
Stratum Source Build Fails Due to Go Related Error


masesk:
**Describe the bug**
Stratum fails to build when cloning and building on a new system with a Go related issue.

**Environment**
Please Note: I have previously built Stratum on the exact same environment with no errors or issues.

- WSL 2 running Ubuntu 22.04
- Ubuntu clang version 14.0.0-1ubuntu1.1
- bazel 4.2.3

If building from source:
`bazel version`
bazel 4.2.3

**To Reproduce**
1. Clone Stratum on main, expunge any previous dependencies with `bazel clean --expunge`
2. Run this command `bazel build //stratum/hal/bin/dummy:stratum_dummy --sandbox_debug --verbose_failures`
3. See the error below:

```
ERROR: /home/user/.cache/bazel/_bazel_user/222cade390d2e9be9ec3a02be0f4ca7e/external/org_golang_google_grpc/internal/syscall/BUILD.bazel:3:11: GoCompilePkg external/org_golang_google_grpc/internal/syscall/go_default_library.a [for host] failed: (Exit 1): linux-sandbox failed: error executing command
  (cd /home/user/.cache/bazel/_bazel_user/222cade390d2e9be9ec3a02be0f4ca7e/sandbox/linux-sandbox/537/execroot/com_github_stratum_stratum && \
  exec env - \
    CGO_ENABLED=1 \
    GOARCH=amd64 \
    GOOS=linux \
    GOPATH='' \
    GOROOT=external/go_sdk \
    GOROOT_FINAL=GOROOT \
    PATH=/usr/bin:/bin \
    TMPDIR=/tmp \
[...]

PR #1104, opened 2023-11-18
override go net, text, and sys modules after grpc dependencies due to new go version

masesk:
override go net, text, and sys modules after grpc dependencies due to new go version

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -93,6 +93,33 @@ load("@io_bazel_rules_go//tests:grpc_repos.bzl", "grpc_dependencies")
 grpc_dependencies()
 
 # ---------------------------------------------------------------------------
+#       Overiride net, text, and sys Go modules to patch new Go update
+# ---------------------------------------------------------------------------
+
+go_repository(
+    name = "org_golang_x_net",
+    importpath = "golang.org/x/net",
+    sum = "h1:hZ/3BUoy5aId7sCpA/Tc5lt8DkFgdVS2onTpJsZ/fl0=",
+    version = "v0.1.0",
+)
+
+go_repository(
+    name = "org_golang_x_text",
+    importpath = "golang.org/x/text",
+    sum = "h1:g61tztE5qeGQ89tm6NTjjM9VPIm088od1l6aSorWRWg=",
+    version = "v0.3.0",
+)
+
+go_repository(
+    name = "org_golang_x_sys",
+    importpath = "golang.org/x/sys",
+    sum = "h1:eG7RXZHdqOJ1i+0lgLgCpSXAp6M3LYlAo6osgSi0xOM=",
+    version = "v0.5.0",
+)
+
+
+
+# ---------------------------------------------------------------------------
 #       Load CDLang dependencies.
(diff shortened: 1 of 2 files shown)


masesk:
#1103

WORKSPACE
```diff
@@ -92,6 +92,33 @@ gazelle_dependencies(go_sdk = "go_sdk")
...
 
+# ---------------------------------------------------------------------------
+#       Overiride net, text, and sys Go modules to patch new Go update
+# ---------------------------------------------------------------------------
+
+go_repository(
+    name = "org_golang_x_net",
+    importpath = "golang.org/x/net",
+    sum = "h1:hZ/3BUoy5aId7sCpA/Tc5lt8DkFgdVS2onTpJsZ/fl0=",
+    version = "v0.1.0",
```
bocon13, line 103: Is there a reason that you picked this version?

Otherwise, I would suggest v0.13.0

WORKSPACE
```diff
@@ -92,6 +92,33 @@ gazelle_dependencies(go_sdk = "go_sdk")
...
+)
+
+go_repository(
+    name = "org_golang_x_text",
+    importpath = "golang.org/x/text",
+    sum = "h1:g61tztE5qeGQ89tm6NTjjM9VPIm088od1l6aSorWRWg=",
+    version = "v0.3.0",
```
bocon13, line 110: Is there a reason for this version?

Suggest v0.13.0

WORKSPACE
```diff
@@ -92,6 +92,33 @@ gazelle_dependencies(go_sdk = "go_sdk")
...
+)
+
+go_repository(
+    name = "org_golang_x_sys",
+    importpath = "golang.org/x/sys",
+    sum = "h1:eG7RXZHdqOJ1i+0lgLgCpSXAp6M3LYlAo6osgSi0xOM=",
+    version = "v0.5.0",
```
bocon13, line 117: Is there a reason for this version?

Suggest v0.13.0

bocon13 (commented):
The build is still failing due to another unpin tool -- reuse. I've fixed that in a new PR that could replace this one #1105 

Ideally, we start to push versions forward, but I think this solution is totally acceptable and also fits with Bazel's mantra of pinning dependencies.

Thanks for putting this together!

WORKSPACE
masesk: My apologies for the late reply. I got these versions by inspecting some gazelle file for the version we are using for sys (which was the issue from the start) then found the latest compatible version that matches that version of sys. I can test v0.13.0 and report if it works. If it does, I will change all of them to v0.13.0

masesk: This still seems to fail for me.

bocon13: That's a good reason to go with the versions you picked. The versions I suggested were the latest before the most obvious breakage, but I agree that there are still problems with them.

bocon13:
Do you want to add the reuse from #1105 change to this PR?

File is `.circleci/config_template.yml`

That should make CI happy, then we can merge this one.

